PR #8, opened 2016-07-01
[CARBONDATA-30] Record load performance statistics

Zhangshunyu:
1)We can use a parameter "enable.data.loading.statistics" which can be configured by user to determine whether the statistics will be recorded and calculated during data loading, the default value is false.
2)We defined a dummy util, if we don't need to record the statistics, it will do nothing.

For example, we set "enable.data.loading.statistics" to "true" and run CarbonExample, the message as following:

Data load request has been received for table default.t3
```
========== TIME_STATISTICS PartitionID: 0==========
STAGE 1 ->Load csv to DataFrame and generate block distinct values: 0.33(s)
STAGE 2 ->Global dict shuffle and write dict file: 0.106(s)
STAGE 2.1 ->  |_maximum distinct column shuffle time: 0.031(s)
STAGE 2.2 ->  |_maximum distinct column write dict file time: 0.053(s)
STAGE 3 ->LRU cache load: 0.001(s)
STAGE 4 ->Total cost of gen surrogate key, sort and write to temp files: 0.625(s)
STAGE 4.1 ->  |_read csv file: 0.47(s)
STAGE 4.2 ->  |_transform to surrogate key: 0.377(s)
STAGE 4.3 ->  |_sort rows and write to temp file: 0.366(s)
STAGE 5 ->Tansform to MDK, compress and write fact files: 0.165(s)
========== BLOCK_INFO ==========
BLOCK_INFO ->Node host: localhost
BLOCK_INFO ->The block count in this node: 1
===============Load_Speed_Info===============
Total Num of Records Processed: 1000
Total Time Cost: 1.227(s)
Total Load Speed: 814records/s
Generate Dictionaries Speed: 3030records/s
Read CSV Speed: 2127 records/s
Generate Surrogate Key Speed: 2652 records/s
Sort Key/Write Temp Files Speed: 2732 records/s
MDK Step Speed: 6060 records/s
=============================================
```
## How was this patch tested?
Pass the exist tests.

### DIFF
--- a/core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
@@ -415,6 +415,14 @@ public final class CarbonCommonConstants {
    */
   public static final String AGGREAGATE_COLUMNAR_KEY_BLOCK_DEFAULTVALUE = "true";
   /**
+   * TIME_STAT_UTIL_TYPE
+   */
+  public static final String ENABLE_DATA_LOADING_STATISTICS = "enable.data.loading.statistics";
+  /**
+   * TIME_STAT_UTIL_TYPE_DEFAULT
+   */
+  public static final String ENABLE_DATA_LOADING_STATISTICS_DEFAULT = "false";
+  /**
    * IS_INT_BASED_INDEXER
    */
   public static final String HIGH_CARDINALITY_VALUE = "high.cardinality.value";
--- a/core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsFactory.java
+++ b/core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsFactory.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.carbondata.core.util;
+
+import org.carbondata.core.constants.CarbonCommonConstants;
+
+public class CarbonTimeStatisticsFactory {
+  private static String LoadStatisticsInstanceType;
+  private static LoadStatistics LoadStatisticsInstance;
+
+  static {
+    CarbonTimeStatisticsFactory.updateTimeStatisticsUtilStatus();
+    LoadStatisticsInstance = genLoadStatisticsInstance();
+  }
+
+  private static void updateTimeStatisticsUtilStatus() {
+    LoadStatisticsInstanceType = CarbonProperties.getInstance()
+        .getProperty(CarbonCommonConstants.ENABLE_DATA_LOADING_STATISTICS,
+            CarbonCommonConstants.ENABLE_DATA_LOADING_STATISTICS_DEFAULT);
+  }
+
+  private static LoadStatistics genLoadStatisticsInstance() {
+    switch (LoadStatisticsInstanceType.toLowerCase()) {
+      case "false":
+        return CarbonLoadStatisticsDummy.getInstance();
+      case "true":
+        return CarbonLoadStatisticsImpl.getInstance();
+      default:
+        return CarbonLoadStatisticsDummy.getInstance();
+    }
+  }
+
+  public static LoadStatistics getLoadStatisticsInstance() {
+    return LoadStatisticsInstance;
+  }
+
+}
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataLoadRDD.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataLoadRDD.scala
@@ -18,13 +18,18 @@
 
 package org.carbondata.spark.rdd
 
+import java.lang.Long
 import java.util.UUID
 
 import scala.collection.JavaConverters._
 import scala.util.Random
 
-import org.apache.spark.{Logging, Partition, SerializableWritable, SparkContext, TaskContext}
+import org.apache.spark.Logging
+import org.apache.spark.Partition
+import org.apache.spark.SerializableWritable
+import org.apache.spark.SparkContext
 import org.apache.spark.SparkEnv
+import org.apache.spark.TaskContext
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.execution.command.Partitioner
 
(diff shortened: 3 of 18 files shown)


core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -415,6 +415,14 @@
    */
   public static final String AGGREAGATE_COLUMNAR_KEY_BLOCK_DEFAULTVALUE = "true";
   /**
+   * TIME_STAT_UTIL_TYPE
+   */
+  public static final String TIME_STAT_OPEN = "time.statistics.open";
```
ravipesala: Name looks not understandable. Better change as `enable.data.loading.statistics`

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsFactory.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public static AbstractTimeStatisticsUtil getTimeStatisticsUtil() {
+    switch (timeStatisticsUtilType.toLowerCase()) {
+      case "false":
+        return CarbonTimeStatisticsDummyUtil.getInstance();
+      case "true":
+        return CarbonTimeStatisticsUtil.getInstance();
+      default:
+        throw new UnsupportedOperationException("Not supported statistics type");
```
ravipesala: Don't throw exception. Just log it and use default value.

core/src/main/java/org/carbondata/core/util/AbstractTimeStatisticsUtil.java
```diff
@@ -0,0 +1,59 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.carbondata.core.util;
+
+public interface AbstractTimeStatisticsUtil {
```
ravipesala: It is interface so why you have used `Abstract` in name?

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsFactory.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  private static void updateTimeStatisticsUtilStatus() {
+    timeStatisticsUtilType = CarbonProperties.getInstance()
+        .getProperty(CarbonCommonConstants.TIME_STAT_OPEN,
+            CarbonCommonConstants.TIME_STAT_OPEN_DEFAULT);
+  }
+
+  public static AbstractTimeStatisticsUtil getTimeStatisticsUtil() {
+    switch (timeStatisticsUtilType.toLowerCase()) {
```
ravipesala: No need to check every time. You can create instance  in static block

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsUtil.java
```diff
@@ -0,0 +1,362 @@
...
+
+  private static final LogService LOGGER =
+          LogServiceFactory.getLogService(CarbonTimeStatisticsUtil.class.getName());
+
+  /*
+   *We only care about the earliest start time(EST) and the latest end time(LET) of different
+   *threads, who does the same thing, LET - EST is the cost time of doing one thing using
+   *multiple thread.
+ */
+  private static long loadCsvfilesToDfStartTime = 0;
```
ravipesala: It is singleton class so why used `static` modifier for all variables.

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsUtil.java
```diff
@@ -0,0 +1,362 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.carbondata.common.logging.LogService;
+import org.carbondata.common.logging.LogServiceFactory;
+
+/**
+ * A util which provide methods used to record time information druing data loading.
+ */
+public class CarbonTimeStatisticsUtil implements AbstractTimeStatisticsUtil {
```
ravipesala: class name which has `util` is confusing, since we use only static and utility methods used in util classes. Better remove `util` from class name. And also these statistics are only for loading data so class name should reflect that.

ravipesala:
@Zhangshunyu How do you handle concurrent query scenario in recording statistics?

Zhangshunyu:
@ravipesala 
Done, please check agian.
For concurrent query scenario, of you mean multiple threads
if  thread1 and thread2 starts at the same time and ends at different time, and not streaming like process, we record the max of them.
If threre are many threads doing streaming like process, we only record the total time of this step.
but if you mean different users they load data concurrent, it is not support, because this AR is only used by the developer to estimate the load time.

Thank you!

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsFactory.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+package org.carbondata.core.util;
+
+import org.carbondata.core.constants.CarbonCommonConstants;
+
+public class CarbonTimeStatisticsFactory {
+  private static String timeStatisticsUtilType;
+  private static LoadStatistics LoadStatisticsInstance;
```
ravipesala: use proper name for variable like `loadStatisticsInstance`

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -159,6 +166,31 @@ class CarbonDataLoadRDD[K, V](
         }
         carbonLoadModel.setSegmentId(String.valueOf(loadCount))
         setModelAndBlocksInfo()
+        val surrogatekeysTotalTimeArray = new Array[Long](2)
```
ravipesala: I did not get what you are trying to do here. It supposed everything abstract. Handle every thing inside your statistics class.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -195,6 +198,9 @@ class CarbonBlockDistinctValuesCombineRDD(
           }
         }
       }
+      val dfGenEndTime = System.currentTimeMillis()
```
ravipesala: you can take time inside your statistics class itself.

ravipesala:
@Zhangshunyu 
It may be devloper to estimate but we should have mechanism either to throw exception or handle concurrent scenario.
And also the API's are not abstract enough, all the handling code should inside the APIs and also the stages are not understandable. Please have a look at PR https://github.com/HuaweiBigData/carbondata/pull/356  , Please check how the stages and calculation is handled.

Zhangshunyu:
@ravipesala pls review.
For stages, I think load is streaming like process, and it's hard to divide the stages into independent stage1， stage2, ……，stage n, so I hanlded the statistics like the log info shows.

QiangCai:
@Zhangshunyu  
Firstly, you can split dataloading flow to some stages by reference to PR HuaweiBigData/carbondata#356. I think these stages also should contain e2e, driver, executor and so on.

Secondly, you can simplify the API in CarbonLoadStatisticsImpl, CarbonLoadStatisticsDummy and LoadStatistics. Now you defined  methods for each stage, but I think you can use one method with a stage parameter to abstract the API.

PR #10, opened 2016-07-01
[CARBONDATA-29] Make inverted index can be configurable

Zhangshunyu:
## # What changes were proposed in this pull request?

Using DDL to determine which column not do inverted index, if the user doesn't specify for one column, the default value is true which means do inverted index.
For example:

```
CREATE TABLE IF NOT EXISTS index
(id Int, name String, city String)
STORED BY 'org.apache.carbondata.format'
TBLPROPERTIES('NO_INVERTED_INDEX'='name,city')
```
## # How was this patch tested?

Pass the exist tests including new testcases.


chenliang613:
@ravipesala  please review this PR

core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
```diff
@@ -0,0 +1,134 @@
...
+
+package org.carbondata.core.datastorage.store.columnar;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.carbondata.core.constants.CarbonCommonConstants;
+import org.carbondata.core.util.ByteUtil;
+
+public class BlockIndexerStorageForNoInvertedIndex implements IndexStorage<int[]> {
```
ravipesala: Lot of code is duplicated , Better make Abstract class and move the common code.

core/src/main/java/org/carbondata/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -166,34 +166,20 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
-      last = startKey;
-      for (int j = startKey + 1; j < numerOfRows; j++) {
-        if (ByteUtil.UnsafeComparer.INSTANCE
-            .compareTo(dimColumnDataChunk.getCompleteDataChunk(), j * filterValues[k].length,
-                filterValues[k].length, filterValues[k], 0, filterValues[k].length) == 0) {
-          bitSet.flip(j);
-          last++;
-        } else {
-          break;
+    if (dimColumnDataChunk instanceof FixedLengthDimensionDataChunk) {
```
ravipesala: No need of `if` condition

core/src/main/java/org/carbondata/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -166,34 +166,20 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+      FixedLengthDimensionDataChunk fixedChunk = (FixedLengthDimensionDataChunk) dimColumnDataChunk;
```
ravipesala: no need of type casting

core/src/main/java/org/carbondata/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -166,34 +166,20 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+      int start = 0;
```
ravipesala: what is the use of `start`

core/src/main/java/org/carbondata/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -166,34 +166,20 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+      bitSet.flip(0, numerOfRows);
+      byte[][] filterValues = dimColumnExecuterInfo.getFilterKeys();
+      for (int k = 0; k < filterValues.length; k++) {
+        for (int j = start + 1; j < numerOfRows; j++) {
```
ravipesala: I guess the logic `j = start + 1` seems wrong. It supposed to start from 0 like `j = 0`

processing/src/main/java/org/carbondata/processing/store/writer/CarbonFactDataWriterImplForIntIndexAndAggBlock.java
```diff
@@ -106,12 +111,36 @@ public NodeHolder buildDataNodeHolder(IndexStorage<int[]>[] keyStorageArray, byt
...
+          allMinValue[i] = keyStorageArray[i].getMin();
+          allMaxValue[i] = keyStorageArray[i].getMax();
+        } else {
+          allMinValue[i] = updateMinMaxForNoDictionary(keyStorageArray[i].getMin());
+          allMaxValue[i] = updateMinMaxForNoDictionary(keyStorageArray[i].getMax());
+        }
       } else {
-        allMinValue[i] = updateMinMaxForNoDictionary(keyStorageArray[i].getMin());
-        allMaxValue[i] = updateMinMaxForNoDictionary(keyStorageArray[i].getMax());
+        byte[][] minmaxArray = keyStorageArray[i].getKeyBlock();
```
ravipesala: Move the code of calculating min max to the `BlockIndexerStorageForNoInvertedIndex`. It already has method to get min and max so better calculate there and make it available to `getMin` and `getMax` methods.

Zhangshunyu:
@ravipesala pls check, and the CI on older huaweibigdata/carbondata has passed which only has conflict with older repository but has no conflicts with this new base branch.

core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
Zhangshunyu: Here they have different input parameter that one is keyBlockInput and another is ColumnWithIntIndex and I think keep them in 2 impl is also ok.

core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * @return the keyBlock
+   */
+  public byte[][] getKeyBlock() {
+    return keyBlock;
+  }
+
+  @Override public byte[] getMin() {
```
ravipesala: Move this calculation code to constructor, you are unnecessarily sorting twice.

core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
ravipesala: Please try to extract code to Abstract class

Zhangshunyu:
@ravipesala Done,  All the test cases and static check have passed. please check.

ravipesala:
LGTM

chenliang613:
OK, LGTM

PR #13, opened 2016-07-02
[CARBONDATA-33]Fixed folder path issue when storePath contains '-'

ravipesala:
If storePath contains '-' like "incubator-carbon/store" , testcases fails and also potential bug to fail the queries.


chenliang613:
tested ok, LGTM!

PR #14, opened 2016-07-02
[CARBONDATA-31] To unify definition for SparkContext,CarbonContext,HiveContext,SQLContext

chenliang613:
For SparkContext,CarbonContext,HiveContext,SQLContext definition, different places having different definition, for example:
1)val sc = new CaronContext, val carboncontext = new CaronContext
2)var oc: HiveContext , var context: HiveContext
....
To unify definition for SparkContext,CarbonContext,HiveContext,SQLContext as below:
CarbonContext : cc
SQLContext : sqlContext
SparkContext : sc
HiveContext : hiveContext

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
@@ -101,7 +101,8 @@ case class DictionaryMap(dictionaryMap: Map[String, Boolean]) {
   }
 }
 
-class CarbonMetastoreCatalog(hiveContext: HiveContext, val storePath: String, client: ClientInterface)
+class CarbonMetastoreCatalog(hiveContext: HiveContext, val storePath: String,
+    client: ClientInterface)
   extends HiveMetastoreCatalog(client, hiveContext)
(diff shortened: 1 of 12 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
```diff
@@ -101,8 +101,8 @@ case class DictionaryMap(dictionaryMap: Map[String, Boolean]) {
   }
 }
 
-class CarbonMetastoreCatalog(hive: HiveContext, val storePath: String, client: ClientInterface)
-  extends HiveMetastoreCatalog(client, hive)
+class CarbonMetastoreCatalog(hiveContext: HiveContext, val storePath: String, client: ClientInterface)
+  extends HiveMetastoreCatalog(client, hiveContext)
```
ravipesala: check style is failing here. Please do `mvn scalastyle:check` and `mvn checkstyle:check`

chenliang613: yes, the length is exceeding 100, will modify it.

ravipesala:
LGTM

PR #15, opened 2016-07-03
[CARBONDATA-34] Fixed failing of drop table with db name

ravipesala:
Table drop fails when drop table with dbname like
`DROP TABLE default.tablename` 
This PR fixes it.
And also it fixes the CARBONDATA-33 issue when store path is hdfs.


chenliang613:
test is ok. LGTM

gvramana:
LGTM

PR #20, opened 2016-07-04
[WIP][CARBONDATA-16] BLOCK distribution in query is not correct in qu…

mohammadshahidkhan:
…ery when number of executors are less than the cluster size.


mohammadshahidkhan:
[WIP] Not tested the scenario when when no executor is in active mode.

ravipesala:
@mohammadshahidkhan Please close this PR as it is already raised against HuaweiBigData

mohammadshahidkhan:
Closing PR as it is already raised in HuaweiBigData
https://github.com/HuaweiBigData/carbondata/pull/787

PR #21, opened 2016-07-05
[CARBONDATA-36]fix dictionary exception when data column count less then schema

eason-lyx:
when csv head is id,name,age,city
if the data record like as the following:
1,a,10,china
2,b
3,c,11,usa
then,when read this row:2,b,generating dicionary will failure,and the remain record (3,c,11,usa)will no generate dictionary.

this pr to prevent put null value to Set,and let it continue.
@ravipesala can u please help to check this pr?
thank u.


ravipesala:
LGTM

chenliang613:
LGTM +1

eason-lyx:
This closes #21

PR #23, opened 2016-07-05
[CARBONDATA-40]	Make the derby db metastore configurable

ravipesala:
Details are given in https://issues.apache.org/jira/browse/CARBONDATA-40

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonContext.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonContext.scala
@@ -17,9 +17,11 @@
 
 package org.apache.spark.sql
 
+import java.io.File
+
 import scala.language.implicitConversions
 
-import org.apache.spark.SparkContext
+import org.apache.spark.{Logging, SparkContext}
 import org.apache.spark.sql.catalyst.ParserDialect
 import org.apache.spark.sql.catalyst.analysis.{Analyzer, OverrideCatalog}
 import org.apache.spark.sql.catalyst.optimizer.{DefaultOptimizer, Optimizer}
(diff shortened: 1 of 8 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonContext.scala
```diff
@@ -32,8 +34,22 @@ import org.carbondata.common.logging.LogServiceFactory
...
-class CarbonContext(val sc: SparkContext, val storePath: String) extends HiveContext(sc) {
+class CarbonContext(
+    val sc: SparkContext,
+    val storePath: String,
+    metaStorePath: String) extends HiveContext(sc) with Logging {
   self =>
+
+  def this (sc: SparkContext) = {
+    this (sc,
+      new File("./carbonstore").getCanonicalPath,
```
gvramana: there is a configuration CarbonCommonConstants.STORE_LOCATION, that should be taken as default value when storelocation is not provided.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonContext.scala
```diff
@@ -32,8 +34,22 @@ import org.carbondata.common.logging.LogServiceFactory
...
+      new File("./carbonmetastore").getCanonicalPath)
```
gvramana: we also can add a configuration  CarbonCommonConstants.META_STORE_LOCATION, defaulting to ./carbonmetastore.

gvramana:
Other than configuration comments, remaining LGTM.

ravipesala:
Fixed review comments

QiangCai:
Verified in local spark shell.

QiangCai:
Good. I also find this compile error in test case. I will cancel my modification.

QiangCai:
I think this pr is important to simplify QuickStart.
@ravipesala can you refresh [QuickStart](https://github.com/apache/incubator-carbondata/blob/master/docs/Quick-Start.md)?

ravipesala:
@chenliang613 Please review and merge this PR.

chenliang613:
verified , LGTM

PR #29, opened 2016-07-07
[CARBONDATA-44]Segment folders are deleted for multiple loads

ravipesala:
When loading the multiple times and the store path is relative then the old segment folders are deleted.
The defect is able to reproduce when the store path is relative path, but I am not able to reproduce the issue when the store path is hdfs as mentioned in the jira issue.
https://issues.apache.org/jira/browse/CARBONDATA-44


chenliang613:
tested ok, LGTM

PR #31, opened 2016-07-08
[CARBONDATA-46]Fixed dataframe api in carbon

ravipesala:
Currently `examples/src/main/scala/org/carbondata/examples/DataFrameAPIExample.scala` is not working. This PR fixes it.
https://issues.apache.org/jira/browse/CARBONDATA-46


chenliang613:
verified, LGTM.

PR #36, opened 2016-07-13
CARBONDATA-5 data miss match with hive

mohammadshahidkhan:
data mismatch between the carbon Table and Hive Table for columns having \N for non numeric data type
Modification: 
A property "**serialization.null.format**" is exposed to configure the value to be treated as null.
Default value of key serialization.null.format to be treated as null is \N.


mohammadshahidkhan:
image

image

mohammadshahidkhan:
Closing this PR as similar issue raised in https://github.com/HuaweiBigData/carbondata/pull/802

PR #39, opened 2016-07-14
[CARBONDATA-1]CarbonData on docker

QiangCai:
This pr provide a guide to deploy carbondata on docker.

Issue Link:
https://issues.apache.org/jira/browse/CARBONDATA-1

### DIFF
--- a/dev/dockerfile
+++ b/dev/dockerfile
@@ -1,0 +1,9 @@
+FROM sequenceiq/spark:1.6.0
+MAINTAINER QiangCai
+
+RUN mkdir ${SPARK_HOME}/carbondata
+ADD carbondata.jar ${SPARK_HOME}/lib
+COPY carbonplugins ${SPARK_HOME}/carbondata/carbonplugins
+ADD data.csv ${SPARK_HOME}/carbondata
+RUN chmod 755 ${SPARK_HOME}/carbondata/data.csv
+Run chmod 755 -R ${SPARK_HOME}/carbondata/carbonplugins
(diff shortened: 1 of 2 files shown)


dev/dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM sequenceiq/spark:1.6.0
```
jackylk: It is better to keep the same as pom.xml, 1.5.2

docs/CarbonData-on-docker.md
```diff
@@ -0,0 +1,72 @@
+#  CarbonData on docker
+This tutorial provides an easy way to try CarbonData. To follow along with this guide, please install [docker](https://www.docker.com/products/docker) at first.
+
+## 1 Building carbondata
+To get the jar files, we need build carbondata project. Here Spark version is 1.6.1 and Hadoop version is 2.7.2.
+```shell
+git clone https://github.com/apache/incubator-carbondata.git carbondata
```
jackylk: Why not do all these steps in the docker image? 
To let user try carbon quickly, I suggest just let user run carbonate-docker image and can start query immediately. Is there any difficulties in do that?

jackylk:
Thanks for working on this. Please see the comments.

dev/dockerfile
QiangCai: sequenceiq/spark didn't release version 1.5.2. 
Version 1.5.1 is unavailable now.
So I can only use version 1.6.0.

docs/CarbonData-on-docker.md
QiangCai: Now we don't release a packaged carbon version.  Because we can reuse exists maven repository in local machine, so I use maven to build carbon in local machine. In docker image, we need to download many dependence jars from Apache maven repository again. 

Better to download a packaged carbon version directly.   After release version, we can skip maven building, and just need to use wget to download the packaged carbon version in docker image.

QiangCai:
close this pr

PR #41, opened 2016-07-16
[CARBONDATA-60] Fixed wrong result when using union all

ravipesala:
https://issues.apache.org/jira/browse/CARBONDATA-60


chenliang613:
verified,LGTM

PR #42, opened 2016-07-17
[CARBONDATA-61] Change cube to table

jackylk:
Due to historical reason, CarbonData uses cube concept instead of table, but it is actually not doing aggregation table, so in this PR all cube and Cube term are changed to table and Table.

JIRA issue is pending to create


chenliang613:
@ravipesala  please review

ravipesala:
LGTM. Verified.

chenliang613:
LGTM +1

PR #44, opened 2016-07-19
CARBONDATA-80 Dictionary values should be equally distributed in buckets while loading in memory

manishgupta88:
jira link: 
https://issues.apache.org/jira/browse/CARBONDATA-80

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/cache/dictionary/AbstractColumnDictionaryInfo.java
+++ b/core/src/main/java/org/apache/carbondata/core/cache/dictionary/AbstractColumnDictionaryInfo.java
@@ -25,6 +25,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
+import org.carbondata.core.util.CarbonUtil;
 
 /**
  * class that implements cacheable interface and methods specific to column dictionary
@@ -62,6 +63,11 @@ public abstract class AbstractColumnDictionaryInfo implements DictionaryInfo {
   private long dictionaryMetaFileLength;
 
   /**
+   * size of one dictionary bucket
+   */
+  private final int dictionaryOneChunkSize = CarbonUtil.getDictionaryChunkSize();
+
+  /**
    * This method will return the timestamp of file based on which decision
    * the decision will be taken whether to read that file or not
    *
@@ -96,6 +102,16 @@ public abstract class AbstractColumnDictionaryInfo implements DictionaryInfo {
    */
   @Override public void incrementAccessCount() {
     accessCount.incrementAndGet();
+  }
+
+  /**
+   * This method will return the size of of last dictionary chunk so that only that many
+   * values are read from the dictionary reader
+   *
+   * @return size of last dictionary chunk
+   */
+  @Override public int getSizeOfLastDictionaryChunk() {
+    return 0;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
```diff
@@ -112,10 +113,35 @@ public ColumnDictionaryInfo(DataType dataType) {
...
+      if (differenceInLastDictionaryAndOneChunkSize > 0) {
+        // if difference is greater than new dictionary size then copy a part of list
+        // else copy the complete new dictionary chunk list in the last dictionary chunk list
+        if (differenceInLastDictionaryAndOneChunkSize >= newDictionaryChunk.size()) {
+          lastDictionaryChunk.addAll(newDictionaryChunk);
+        } else {
+          List<byte[]> subListOfNewDictionaryChunk =
+              newDictionaryChunk.subList(0, differenceInLastDictionaryAndOneChunkSize);
+          lastDictionaryChunk.addAll(subListOfNewDictionaryChunk);
+          subListOfNewDictionaryChunk.clear();
```
gvramana: Use one more sub list and add remaining

manishgupta88:
@gvramana ...please review it.

core/src/main/java/org/carbondata/core/cache/dictionary/AbstractColumnDictionaryInfo.java
```diff
@@ -241,22 +257,21 @@ private void decrementAccessCount() {
...
     byte[] dictionaryValueInBytes = null;
-    int totalSizeOfDictionaryChunksTraversed = 0;
-    for (List<byte[]> oneDictionaryChunk : dictionaryChunks) {
-      totalSizeOfDictionaryChunksTraversed =
-          totalSizeOfDictionaryChunksTraversed + oneDictionaryChunk.size();
-      // skip the dictionary chunk till surrogate key is lesser than size of
-      // dictionary chunks traversed
-      if (totalSizeOfDictionaryChunksTraversed < surrogateKey) {
-        continue;
+    if (surrogateKey > 0) {
```
kumarvishal09: why we need this check ? As data loading surrogate key starts from 1 , so in any case if surrogate key is less than 1 then  there is some bug in unpacking the mdkey

core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryChunkWrapper.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+
+import org.carbondata.format.ColumnDictionaryChunk;
+
+/**
+ * This class is a wrapper over column dictionary chunk thrift object.
+ * The wrapper class wraps the list<ColumnDictionaryChunk> and provides an API
+ * to fill the byte array into list
+ */
+public class ColumnDictionaryChunkWrapper implements Iterator<byte[]> {
```
kumarvishal09: Please extend carbonIterator class as here we are not supporting remove operation

kumarvishal09: Please correct me if I am wrong it is an iterator over dictionary values?? If so then change the class name to ColumnDictionaryChunkIterator

core/src/main/java/org/carbondata/core/cache/dictionary/AbstractColumnDictionaryInfo.java
```diff
@@ -241,22 +257,21 @@ private void decrementAccessCount() {
...
+      // surrogate key starts from 1 and list index will start from 0, so lets say if surrogate
+      // key is 10 then value will present at index 9 of the dictionary chunk list
+      int actualSurrogateIndex = surrogateKey - 1;
+      // lets say dictionaryOneChunkSize = 10, surrogateKey = 10, so bucket index will
+      // be 0 and dictionary chunk index will be 9 to get the value
+      int dictionaryBucketIndex = actualSurrogateIndex / dictionaryOneChunkSize;
+      if (dictionaryChunks.size() > dictionaryBucketIndex) {
```
kumarvishal09: This if check is required?? As i cannot find any scenario when dictionary bucket index will more than dicitonaryChunkSize.

core/src/main/java/org/carbondata/core/cache/dictionary/AbstractColumnDictionaryInfo.java
```diff
@@ -241,22 +257,21 @@ private void decrementAccessCount() {
...
+        int indexInsideBucket = actualSurrogateIndex % dictionaryOneChunkSize;
+        List<byte[]> dictionaryBucketContainingSurrogateValue =
+            dictionaryChunks.get(dictionaryBucketIndex);
+        if (dictionaryBucketContainingSurrogateValue.size() > indexInsideBucket) {
```
kumarvishal09: Same as above.Please check

core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryChunkWrapper.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  @Override public byte[] next() {
+    if (iteratorIndex >= columnDictionaryChunks.get(outerIndex).getValues().size()) {
+      iteratorIndex = 0;
+      outerIndex++;
+    }
+    ByteBuffer buffer = columnDictionaryChunks.get(outerIndex).getValues().get(iteratorIndex);
+    int length = buffer.limit();
+    byte[] value = new byte[length];
+    buffer.get(value, 0, value.length);
```
kumarvishal09: Can we use buffer.array instead of calling get as get will copy the data so we can avoid System.arrayCopy

core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
```diff
@@ -112,10 +113,50 @@ public ColumnDictionaryInfo(DataType dataType) {
...
+          dictionaryOneChunkSize - lastDictionaryChunk.size();
+      if (differenceInLastDictionaryAndOneChunkSize > 0) {
+        // if difference is greater than new dictionary size then copy a part of list
+        // else copy the complete new dictionary chunk list in the last dictionary chunk list
+        if (differenceInLastDictionaryAndOneChunkSize >= newDictionaryChunk.size()) {
+          lastDictionaryChunk.addAll(newDictionaryChunk);
+        } else {
+          List<byte[]> subListOfNewDictionaryChunk =
+              newDictionaryChunk.subList(0, differenceInLastDictionaryAndOneChunkSize);
+          lastDictionaryChunk.addAll(subListOfNewDictionaryChunk);
```
kumarvishal09: Why we are calling sublist instead we can iterate and fill the values to lastDictionaryChunk

manishgupta88: @kumarvishal09 ...sublist will create a new object but it will return a view of the parent list itself and all operations will be performed on the parent list. It will not create and fill a new array. So I think it should be ok to use sublist

manishgupta88:
@gvramana @kumarvishal09 ...review comments have been handled. Kindly check and merge

core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
kumarvishal09: Yes you write invalid comment

gvramana:
LGTM

PR #49, opened 2016-07-21
[CARBONDATA-88] Fixed relative path issue from carbon-spark-shell

ravipesala:
This issue happens when `carbon-spark-shell` or `carbon-spark-sql` runs from any other folder other than   carbondata/bin folder.


chenliang613:
verified, LGTM

PR #50, opened 2016-07-21
[CARBONDATA-92]Refactored code to improve performance by using removing unnecessary result conversion

ravipesala:
Remove the unnecessary intermediate conversion of key while scanning.
Basically it removes one step in result conversion.
It avoids System.arraycopy while converting to result. 
It avoids the result preparation step.


chenliang613:
@jackylk @gvramana  please review it.

core/src/main/java/org/carbondata/core/carbon/datastore/chunk/impl/ColumnGroupDimensionDataChunk.java
```diff
@@ -67,10 +67,29 @@ public ColumnGroupDimensionDataChunk(byte[] dataChunk, DimensionChunkAttributes
...
+   * @param row
+   * @param info  @return
+   */
+  @Override public int fillConvertedChunkData(int rowId, int columnIndex, int[] row,
+      KeyStructureInfo info) {
+    int start = rowId * chunkAttributes.getColumnValueSize();
+    int sizeInBytes = info.getKeyGenerator().getKeySizeInBytes();
+    byte[] key = new byte[sizeInBytes];
+    System.arraycopy(dataChunk, start, key, 0, sizeInBytes);
+    long[] keyArray = info.getKeyGenerator().getKeyArray(key);
```
gvramana: we can avoid this copy also by changing getKeyArray interface to take start and end position, or wrapping it with ByteBuffer

core/src/main/java/org/carbondata/core/carbon/datastore/chunk/impl/FixedLengthDimensionDataChunk.java
```diff
@@ -69,9 +69,31 @@ public FixedLengthDimensionDataChunk(byte[] dataChunk, DimensionChunkAttributes
   }
 
   /**
+   * Converts to column dictionary integer value
+   * @param rowId
+   * @param columnIndex
+   * @param row
+   * @param restructuringInfo  @return
```
gvramana: What is the return value

ravipesala:
@gvramana Fixed comments, Please review

gvramana:
ok, I am taking the code and testing performance.

core/src/main/java/org/carbondata/core/carbon/datastore/chunk/impl/FixedLengthDimensionDataChunk.java
```diff
@@ -69,9 +69,27 @@ public FixedLengthDimensionDataChunk(byte[] dataChunk, DimensionChunkAttributes
...
+      rowId = chunkAttributes.getInvertedIndexesReverse()[rowId];
+    }
+    int start = rowId * chunkAttributes.getColumnValueSize();
+    int dict = 0;
+    for (int i = start; i < start + chunkAttributes.getColumnValueSize(); i++) {
+      dict <<= 8;
+      dict ^= dataChunk[i] & 0xFF;
+    }
+    row[columnIndex] = dict;
+    return columnIndex + 1;
```
kumarvishal09: @ravipesala do we need to return anything from this method .Please correct me if I am wrong:)

core/src/main/java/org/carbondata/scan/collector/impl/AbstractScannedResultCollector.java
```diff
@@ -170,33 +120,25 @@ private Object getMeasureData(MeasureColumnDataChunk dataChunk, int index, DataT
...
-
-
   /**
    * Below method will be used to update the fixed length key with the
    * latest block key generator
    *
    * @return updated block
    */
-  private void updateKeyWithLatestBlockKeygenerator() {
+  private void updateKeyWithLatestBlockKeygenerator(List<Object[]> listBasedResult) {
```
kumarvishal09: Please correct me if i m wrong but now we are from executor layer for dictionary columns we are sending surrogate keys ..key updation is required only when aggregation is done is executor ..this is because of dynamic cardinality. As now from DimensionColumnDataChunk we are filling the surrogate key ,so now do we need this method ??

core/src/main/java/org/carbondata/core/carbon/datastore/chunk/impl/FixedLengthDimensionDataChunk.java
ravipesala: @kumarvishal09  I am returning the column index for next column to fill. Please check the `ColumnGroupDimensionDataChunk.fillConvertedChunkData` to know why I am returning the column index,

core/src/main/java/org/carbondata/scan/collector/impl/AbstractScannedResultCollector.java
ravipesala: @kumarvishal09 Yes you are right, as we are filling with surrogates we don't require to update/restructure the key. Key update/restructure is not required only for `DictionaryBasedResultCollector` but it is required for `RawBasedResultCollector`. That is why it is added here.

PR #51, opened 2016-07-22
[CARBONDATA-49]Fixed the query fails if big data is loaded to local store.

ravipesala:
For carbondata the segment file is not splittable but it is splitting when the store is local. So we should avoid the file splitting for local store files.


chenliang613:
verified, LGTM

PR #52, opened 2016-07-22
[WIP] Support varchar datatype as SPARK

Zhangshunyu:
Support varchar datatype as spark, and spark treat varchar as string to process it.


Zhangshunyu:
Could any one please review this code?

PR #54, opened 2016-07-25
[CARBONDATA-107] Remove unnecessary ConvertToSafe

jackylk:
CarbonDictionaryDecoder is using InternalRow only, so it should be able to process UnsafeRow.
By changing `canProcessUnsafeRows` and `canProcessSafeRows` to `true`, the planner will remove unnecessary ConvertToSafe operator


ravipesala:
LGTM

PR #55, opened 2016-07-26
[CARBONDATA-108] Remove project in strategy

jackylk:
Modify CarbonStrategy, when project equals scan column, do scan without project

Modify some test cases to drop table


chenliang613:
LGTM

PR #56, opened 2016-07-26
[CARBONDATA-117] blocklets distribution for query performance improvement 

mohammadshahidkhan:


kumarvishal09:
@mohammadshahidkhan Please handle all the commnets raised as a part of https://github.com/HuaweiBigData/carbondata/pull/882  in this pr if applicable

core/src/main/java/org/carbondata/core/carbon/datastore/block/BlockletInfos.java
```diff
@@ -0,0 +1,112 @@
...
+  public int getNoOfBlockLets() {
+    return noOfBlockLets;
+  }
+
+  /**
+   * sets the number of blockLets
+   *
+   * @param noOfBlockLets
+   */
+  public void setNoOfBlockLets(int noOfBlockLets) {
```
PraveenAdlakha: we can use lombok plugin in the project will save us writing lot of code.Its being used in many apache projects.Let me know in case you need any help.

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
```diff
@@ -185,6 +207,16 @@ public long getBlockLength() {
         > ((TableBlockInfo) other).blockOffset + ((TableBlockInfo) other).blockLength) {
       return 1;
     }
+    //compare the startBlockLetNumber
+    int diffStartBlockLetNumber =
+        blockletInfos.getStartBlockletNumber() - ((TableBlockInfo) other).blockletInfos
```
PraveenAdlakha: Hi ,

Would like to suggest a couple of things here remove Comparable<Distributable>  from the class definition as Distributable is already implementing it.

Let use generics so that we donot have to typeCast everywhere in the compare method for that need to do two things:
1) Change the class definition of Distributable  to :  
public abstract class Distributable<T extends Distributable<T>> implements Comparable<T>

2)  Change the compareTo method definition to:  
public int compareTo(TableBlockInfo other)

Let me know incase you are facing any isssue in doing this I will provide the patch if needed.

core/src/main/java/org/carbondata/core/carbon/datastore/block/BlockletInfos.java
ravipesala: @PraveenAdlakha Thank you for your suggestion, lombok really simplifies the coding. Definitely we will discuss and add the plugin to carbondata.

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
ravipesala: @PraveenAdlakha  your point is valid. We can refactor this code. Can you work on it and provide the patch.

mohammadshahidkhan: @ravipesala @PraveenAdlakha i have already handled the review comments together with this PR.

core/src/main/java/org/carbondata/core/carbon/datastore/block/BlockletInfos.java
mohammadshahidkhan: @ravipesala But this PR is for improving the performance and it is not  decided that to be taken or not, and the comment is valid for the base code itsellf, so i think its better to handle this comment in separate PR. Please suggest.

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
mohammadshahidkhan: comment will be taken care in jirra [CARBONDATA-155](https://issues.apache.org/jira/browse/CARBONDATA-155)

gvramana:
LGTM

PR #61, opened 2016-08-04
Rename packages, add legal file, and fix headers

jbonofre:
Large pull request fixing all naming, legal and headers.

I still have an issue in the core tests (it still uses the "old" package name `org.carbondata` instead of `org.apache.carbondata`).


ravipesala:
Please comment the class `org.apache.carbondata.core.carbon.datastore.BlockIndexStoreTest` as the store need to be updated for this test case. So we can uncomment after update the store
And there are lot of compilation failures from integration/spark because of package change and imports missing.

jbonofre:
Hi guys,
Actually, to simplify the review, I will close this PR to create smaller (like #62).

Please ignore this PR for now. Can you review #62 and the coming ones ?

Thanks !

ravipesala:

PR #65, opened 2016-08-07
Fixed aggregate queries with using subqueries in Spark 1.6.2

ravipesala:
Queries using aggregates and sub queries like below are not working with Spark 1.6.2
`select Min(imei) from (select imei from table1 order by imei) t`.

This PR fixes this issue


chenliang613:
LGTM

PR #66, opened 2016-08-07
Fixed distinct count query issue in 1.6.2

ravipesala:
Distinct count queries like below are fixed in spark version 1.6.2
`select count(imei), count(distinct(imei)) from table1`

This PR fixes it.


chenliang613:
LGTM

PR #67, opened 2016-08-08
Fixed IN filter expression deserializing issue from thrift server

ravipesala:
Queries which have `IN` filter is not working from thriftserver, like 
`SELECT IMEI FROM TABLE1 WHERE IMEI IN ('1AA100074','1AA100075','1AA100077')`

This PR fixes it.


PraveenAdlakha:
+1 LGTM

PR #68, opened 2016-08-08
Fixed casesensitive column issue in self join queries

ravipesala:
Queries are failing when the case sensitive columns are used in self join queries like below
`select b.IMEI from table1 a join table1 b on a.imei=b.imei`


chenliang613:
LGTM

PR #75, opened 2016-08-15
[CARBONDATA-157]query exception when decimal(n,n) column has int filter

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-157

Because decimal(n,n) means only fractional part,no integer. so in spark optimizer, the filter expression decimalColumn=intValue is optimized to literal Expression(Null), and in carbon side,we do not judge on this literal Expression, so there is this exception. 
To resolve this, we should add judge on this.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/scan/filter/FilterExpressionProcessor.java
+++ b/core/src/main/java/org/apache/carbondata/scan/filter/FilterExpressionProcessor.java
@@ -215,6 +215,20 @@ public class FilterExpressionProcessor implements FilterProcessor {
       AbsoluteTableIdentifier tableIdentifier, Expression intermediateExpression) {
     ExpressionType filterExpressionType = expressionTree.getFilterExpressionType();
     BinaryExpression currentExpression = null;
+    if (filterExpressionType == ExpressionType.OR || filterExpressionType == ExpressionType.AND) {
+      Expression leftExpression = ((BinaryExpression) expressionTree).getLeft();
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/carbondata/scan/filter/FilterExpressionProcessor.java
```diff
@@ -215,6 +215,20 @@ private FilterResolverIntf createFilterResolverTree(Expression expressionTree,
...
     BinaryExpression currentExpression = null;
+    if (filterExpressionType == ExpressionType.OR || filterExpressionType == ExpressionType.AND) {
+      Expression leftExpression = ((BinaryExpression) expressionTree).getLeft();
+      Expression rightExpression = ((BinaryExpression) expressionTree).getRight();
+      if (!(leftExpression instanceof BinaryExpression ||
+          leftExpression instanceof ConditionalExpression)) {
+        expressionTree = rightExpression;
+        filterExpressionType = rightExpression.getFilterExpressionType();
+      }
+      else if (!(rightExpression instanceof BinaryExpression ||
```
Zhangshunyu: Style

Jay357089: handled

Jay357089:
CI passed. 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/30/

Jay357089:
CARBONDATA-157 has been resolved in pr#114. close this pr.

PR #76, opened 2016-08-16
[CARBONDATA-158] fix load data with first line is null

foryou2030:
jira url: https://issues.apache.org/jira/browse/CARBONDATA-158

in this pr, hanled load data failure
root cause:
while generating dictionary, need to read the firstline of the first non-empty line, but it only judge whether the line start with comment char 
solution:
judge whether the line is empty line


kumarvishal09:
LGTM

chenliang613:
LGTM

PR #77, opened 2016-08-19
CARBONDATA-153 Record count is not matching while loading the data when one data node went down in HA setup

mohammadshahidkhan:
Record count is not matching while loading the data when one data node went down in HA setup

As per previous implementation : 
Scenario: 
No. of Running executors = 3
No. Data Nodes = 2
Total Unique blocks = 96
Then as per the previous implementation no. of blocks per node = 32.
While assigning blocks only 66 blocks getting allocated amongs two execuotors. 
Third executor is not getting any blocks since we are considering only node local allocation.
Solution:
// so now we have a map of node vs blocks. allocate the block as per the order
createOutputMap(nodeBlocksMap, blocksPerNode, uniqueBlocks, nodeAndBlockMapping, activeNodes);

After doing node block mapping we will map the remaining activeNodes in the nodeBlocksMap in the  assignLeftOverBlocks so that assignLeftOverBlocks  can take care the assignment of remaining blocks.
// if any blocks remain then assign them to nodes in round robin.
assignLeftOverBlocks(nodeBlocksMap, uniqueBlocks, blocksPerNode);


gvramana:
LGTM, merged the same.

PR #81, opened 2016-08-22
[CARBONDATA-132] Fix the bug that the CSV file header exception can not be shown to user using beeline. 

Zhangshunyu:
## Why raise this pr:
**For bug fix: The exception that 'CSV File provided is not proper. Column names in schema and csv header are not same' can not be shown to beeline.**

For example, when data load is failed because of wrong csv file header in load DDL, the exception message only shows in executor side like "CSV header provided in DDL is not proper. Column names in schema and CSV header are not the same" but the **user using beeline can not get it from driver side because dirver only shows "Dataload Failure"** , it is very inconvenient for user to get the reason unless he check the executor log info.

## How to solve:
Get the Exception on driver side and parse the cause, get the casue message to driver. Show DataLoadingException is because that it is mainly about CSV file and wrapped in understandable message which can be shown to the user.

## How to test
Add new test cases:
1. If both ddl and file not have fileheader:
Beeline will show like : "DataLoad failure: CSV File provided is not proper. Column names in schema and csv header are not same. CSVFile Name : windows.csv"
2. If ddl did not provide the proper file header:
Beeline will show like :"DataLoad failure: CSV header provided in DDL is not proper. Column names in schema and CSV header are not the same."

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -29,7 +29,7 @@ import scala.util.control.Breaks._
 import org.apache.hadoop.conf.{Configurable, Configuration}
 import org.apache.hadoop.mapreduce.Job
 import org.apache.hadoop.mapreduce.lib.input.FileSplit
-import org.apache.spark.{Logging, Partition, SparkContext, SparkEnv}
+import org.apache.spark.{Logging, Partition, SparkContext, SparkEnv, SparkException}
 import org.apache.spark.sql.{CarbonEnv, SQLContext}
 import org.apache.spark.sql.execution.command.{AlterTableModel, CompactionCallableModel, CompactionModel, Partitioner}
 import org.apache.spark.sql.hive.DistributionUtil
@@ -46,6 +46,7 @@ import org.apache.carbondata.core.util.CarbonUtil
 import org.apache.carbondata.integration.spark.merger.{CompactionCallable, CompactionType}
 import org.apache.carbondata.lcm.locks.{CarbonLockFactory, ICarbonLock, LockUsage}
 import org.apache.carbondata.lcm.status.SegmentStatusManager
+import org.apache.carbondata.processing.etl.DataLoadingException
 import org.apache.carbondata.processing.util.CarbonDataProcessorUtil
 import org.apache.carbondata.spark._
 import org.apache.carbondata.spark.load._
@@ -724,6 +725,8 @@ object CarbonDataRDDFactory extends Logging {
         partitioner.partitionCount, currentLoadCount.toString)
       var loadStatus = CarbonCommonConstants.STORE_LOADSTATUS_SUCCESS
       var status: Array[(String, LoadMetadataDetails)] = null
+      var errorMessage: String = "DataLoad failure"
+      var executorMessage: String = ""
       try {
         status = new
             CarbonDataLoadRDD(sqlContext.sparkContext,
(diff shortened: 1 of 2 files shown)


integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -775,6 +777,13 @@ object CarbonDataRDDFactory extends Logging {
           loadStatus = CarbonCommonConstants.STORE_LOADSTATUS_FAILURE
           logInfo("DataLoad failure")
           logger.error(ex)
+          ex match {
+            case sparkException: SparkException =>
+              if (sparkException.getCause.isInstanceOf[DataLoadingException]) {
+                executorMessage = sparkException.getCause.getMessage
+              }
+            case _ =>
```
manishgupta88: @Zhangshunyu ...for underscore case also get the message and assign it to variable executorMessage because lets say the control comes to case _ then variable executorMessage will have empty value and message will remain as "DataLoad failure: " which will be incorrect

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -775,6 +777,13 @@ object CarbonDataRDDFactory extends Logging {
           loadStatus = CarbonCommonConstants.STORE_LOADSTATUS_FAILURE
           logInfo("DataLoad failure")
           logger.error(ex)
```
manishgupta88: @Zhangshunyu ....Move logInfo and logger.error below the case match block....also in logInfo log the executorMessage. you can move the variable message also above with default value as "DataLoad failure: " so that the same can be used for logInfo.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
Zhangshunyu: Here we only get DataLoadingException from executor and show it directly to user so that he can know his incorrect operation, but the other exception we still use "DataLoad Failure", because we do not show internal error to user.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
Zhangshunyu: OK

Zhangshunyu:
@manishgupta88 Fixed review comments, pls check.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -773,20 +776,27 @@ object CarbonDataRDDFactory extends Logging {
...
         case ex: Throwable =>
           loadStatus = CarbonCommonConstants.STORE_LOADSTATUS_FAILURE
-          logInfo("DataLoad failure")
+          ex match {
+            case sparkException: SparkException =>
+              if (sparkException.getCause.isInstanceOf[DataLoadingException]) {
+                executorMessage = sparkException.getCause.getMessage
+                errorMessage = errorMessage + ": " + executorMessage
+              }
+            case _ =>
```
manishgupta88: @Zhangshunyu ...for case _ also get the message from throwable and add it to error message same as you have done for sparkException case

Zhangshunyu:
@manishgupta88 Fixed review comments, pls check.

Zhangshunyu:
@manishgupta88 Fixed review comments and rebased, pls check.

CI has passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/25/

manishgupta88:
LGTM

chenliang613:
@Zhangshunyu  nice PR, description is very clear and provide clear verification information.

LGTM

PR #82, opened 2016-08-22
[CARBONDATA-165] Support loading fact file with header for all dictionary

foryou2030:
# Why raise this pr?
When fact csv already have header and  giving FILEHEADER along with  ALL_DICTIONARY_PATH option ,  header will be considered as data row , which is not correct.
FILEHEADER option must be given only when CSV do not have header .  We can read the header from fact file when FILEHEADER  is not  given with ALL_DICTIONARY_PATH 

# How to solve?
Add adapter for loading data with header. While loading fact CSV file with header, get file header from fact file instead of option("FILEHEADER") (the interface "getHeaderFormFactFile" did this)


foryou2030:
Passed CI test， pls check： http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/23/

core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
```diff
@@ -1443,5 +1446,32 @@ public static int getDictionaryChunkSize() {
...
+
+    DataInputStream fileReader = null;
+    BufferedReader bufferedReader = null;
+    String readLine = null;
+
+    try {
+      fileReader =
+          FileFactory.getDataInputStream(csvFilePath, FileFactory.getFileType(csvFilePath));
+      bufferedReader =
+          new BufferedReader(new InputStreamReader(fileReader, Charset.defaultCharset()));
```
manishgupta88: @foryou2030 ....instead of using Charset.defaultCharset() use the below line of code.
Charset.forName( CarbonCommonConstants.DEFAULT_CHARSET)

foryou2030: ok, handled

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -650,6 +650,51 @@ object GlobalDictionaryUtil extends Logging {
...
+   * @param carbonLoadModel
+   * @return headers
+   */
+  private def getHeaderFormFactFile(carbonLoadModel: CarbonLoadModel): Array[String] = {
+    var headers: Array[String] = null
+    var factFile: String = null
+    val fileType = FileFactory.getFileType(carbonLoadModel.getFactFilePath)
+    val filePath = FileFactory.getCarbonFile(carbonLoadModel.getFactFilePath, fileType)
+    if (filePath.isDirectory) {
+      val listFiles = filePath.getParentFile.listFiles()
```
manishgupta88: @foryou2030 ....Please clarify the below points.
1. Why you are using filePath.getParentFile?
2. There can be a case where one folder can contain only subFolders. In that case you will not be find a csv file. Ideally to find a file it should be a recursive call.

@QiangCai ...please review this method and correct me if I am wrong

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -650,6 +650,51 @@ object GlobalDictionaryUtil extends Logging {
...
 
   /**
+   * get file headers from fact file
+   *
+   * @param carbonLoadModel
+   * @return headers
+   */
+  private def getHeaderFormFactFile(carbonLoadModel: CarbonLoadModel): Array[String] = {
+    var headers: Array[String] = null
+    var factFile: String = null
```
QiangCai: factFile = arbonLoadModel.getFactFilePath.split(",")(0)

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: Because we have inovked  FileUtils.getPaths to convert user's input to file list in LoadTable, so these codes are unnecessary.

foryou2030: ok, thanks @QiangCai @manishgupta88 
i have handled

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -736,27 +761,20 @@ object GlobalDictionaryUtil extends Logging {
...
-            }
-            fileHeaders = fileHeadersArr.toArray
+          var headers = if (StringUtils.isEmpty(carbonLoadModel.getCsvHeader)) {
+            getHeaderFormFactFile(carbonLoadModel)
           } else {
-            logError("Not found file header! Please set fileheader")
-            throw new IOException("Failed to get file header")
+            carbonLoadModel.getCsvHeader.toLowerCase.split("" + CSVWriter.DEFAULT_SEPARATOR)
           }
+          headers = headers.map(headerName => headerName.trim)
```
QiangCai: headers = headers.map(_)

foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/43/

manishgupta88:
LGTM

PR #83, opened 2016-08-22
[CARBONDATA-169]ColDict and All Dictionary should not be used together

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-169

when user use columnDict and All Dictionary together, exception should be thrown.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -1016,6 +1016,14 @@ class CarbonSqlParser()
       throw new MalformedCarbonCommandException(errorMessage)
     }
 
+    //  COLUMNDICT and ALL_DICTIONARY_PATH can not be used together.
+    if (options.exists(_._1.equalsIgnoreCase("COLUMNDICT")) &&
+      options.exists(_._1.equalsIgnoreCase("ALL_DICTIONARY_PATH"))) {
+      val errorMessage = "Error: COLUMNDICT and ALL_DICTIONARY_PATH can not be used together" +
+        " in options"
+      throw new MalformedCarbonCommandException(errorMessage)
+    }
+
     // check for duplicate options
     val duplicateOptions = options filter {
       case (_, optionlist) => optionlist.size > 1
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/util/ExternalColumnDictionaryTestCase.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/util/ExternalColumnDictionaryTestCase.scala
@@ -20,13 +20,13 @@ package org.apache.carbondata.spark.util
 
 import java.io.File
 
+import org.apache.carbondata.core.carbon.CarbonDataLoadSchema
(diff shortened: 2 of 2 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -1001,6 +1001,12 @@ class CarbonSqlParser()
       "SERIALIZATION_NULL_FORMAT",
       "ALL_DICTIONARY_PATH"
     )
+    if (options.exists(_._1.equalsIgnoreCase("COLUMNDICT")) &&
+      options.exists(_._1.equalsIgnoreCase("ALL_DICTIONARY_PATH"))) {
+      val errorMessage = "Error: COLUMNDICT and ALL_DICTIONARY_PATH can not be used together" +
+        " in options"
+      throw new MalformedCarbonCommandException(errorMessage)
+    }
```
manishgupta88: @Jay357089 .....Move this newly added code below if loop for not isSupported check. I think ideally first we should validate the options whether they are among the ones allowed for load command and then perform other checks.

integration/spark/src/test/scala/org/apache/carbondata/spark/util/ExternalColumnDictionaryTestCase.scala
```diff
@@ -205,6 +205,20 @@ class ExternalColumnDictionaryTestCase extends QueryTest with BeforeAndAfterAll
       loadSqlRelation, "deviceInformationId", "10086")
   }
 
+  test("COLUMNDICT and ALL_DICTIONARY_PATH can not be used together") {
+    try {
+      sql(s"""
+        LOAD DATA LOCAL INPATH "$complexFilePath1" INTO TABLE loadSqlTest
+        OPTIONS('COLUMNDICT'='$extColDictFilePath1',"ALL_DICTIONARY_PATH"='$extColDictFilePath1')
+        """)
+      assert(false)
```
manishgupta88: assert(false) is not required as Load command will throw an exception and control will go to exception block

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
Jay357089: OK. i will move

integration/spark/src/test/scala/org/apache/carbondata/spark/util/ExternalColumnDictionaryTestCase.scala
Jay357089: write assert(false) here is just to make sure that flow goes to catch part, you know, if not go to catch part, then test case should fail. ...i think it is necessary...

manishgupta88: ok

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
Jay357089: done

manishgupta88:
LGTM

PR #84, opened 2016-08-22
[CARBONDATA-167]Fix that 'UndeclaredThrowableException' thrown instead of 'data loading failed' when load failed.

Zhangshunyu:
## Why raise this pr:
For bug fix:
The user using` beeline` when create a table named test and execute
` LOAD DATA INPATH  'hdfs://hacluster/test/chetan/vardhan.csv' INTO table test OPTIONS('DELIMITER'= ',' ,'QUOTECHAR'= '"', 'FILEHEADER '='~');` the fileheader is not proper and
 It will throws exception like:
`Error: java.lang.reflect.UndeclaredThrowableException (state=,code=0)`
**Actually it should tell "DataLoad failed" to user instead of throws such an exception.**

## How to solve:

Catch the  UndeclaredThrowableException and throw the cause wrapped in it which is carbon's real exception message to user.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
@@ -18,6 +18,7 @@
 package org.apache.spark.sql.execution.command
 
 import java.io.File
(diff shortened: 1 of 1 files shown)


sujith71955:
Please close this PR since this scenario will not occur in any scenario.

PR #86, opened 2016-08-23
[CARBONDATA-170] Delete the lock file once the unlocking is done.

ravikiran23:
Issue : The lock files which are used for locking purpose need to get deleted once lock is unlocked.

Solution : deleting the lock files after the unlock.


gvramana:
LGTM

PR #87, opened 2016-08-23
[CARBONDATA-171] Block distribution not proper when the number of active executors more than the node size

mohammadshahidkhan:
Scenario:
suppose block size : 10
Node size = 3
Number of executors = 8
then only 6 blocks are allocated remaining 4 blocks lefts un- assigned.

### DIFF
--- a/integration/spark/src/main/java/org/apache/carbondata/spark/load/CarbonLoaderUtil.java
+++ b/integration/spark/src/main/java/org/apache/carbondata/spark/load/CarbonLoaderUtil.java
@@ -25,15 +25,11 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.net.InetAddress;
-import java.net.InterfaceAddress;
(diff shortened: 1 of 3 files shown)


mohammadshahidkhan:
Fixed issue #CARBONDATA-171

integration/spark/src/main/java/org/apache/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -1204,12 +1208,14 @@ private static void assignLeftOverBlocks(Map<String, List<Distributable>> output
 
     if (activeNodes != null) {
       for (String activeNode : activeNodes) {
-        List<Distributable> blockLst = outputMap.get(activeNode);
```
gvramana: Don't use getBlockList here, it costly. Instead in previous output map population logic, always populate active node name only.

gvramana:
LGTM

PR #89, opened 2016-08-24
[CARBONDATA-174]Fix the problem of hdfs lock and move the lock file inside the table folder.

Zhangshunyu:
## Why raise this pr
1. The hdfs lock file for one table should be put inside this table's store path, this is more reasonable, and if the store path is not set, then we put it into hadoop.tmp.dir.
For example: if the store path of carbon on hdfs is /user/hive/warehouse/carbon.store, then the lock file for this table woud be: /user/hive/warehouse/carbon.store/default/table_name/meta.lock
2. This bug is found by : Some times, hadoop configured wrong hadoop.tmp.dir, hadoop can still work normally, but carbon's hdfs lock can not work normally, it will throws exception: "Table is locked for updation. Please try after some time".

So, we should use the table store path as the first choice and reduce the coupling between hadoop conf and carbon.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -142,6 +142,14 @@ public final class CarbonCommonConstants {
    */
   public static final String LOAD_FOLDER = "Segment_";
   /**
+   * HDFSURL_PREFIX
+   */
+  public static final String HDFSURL_PREFIX = "hdfs://";
+  /**
+   * FS_DEFAULT_FS
+   */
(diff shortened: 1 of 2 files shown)


chenliang613:
Passed CI test:

Module Builds
Success Apache CarbonData :: Assembly	3.7 秒
Success Apache CarbonData :: Common	2.6 秒
Success Apache CarbonData :: Core	8.7 秒
Success Apache CarbonData :: Examples	8.6 秒
Success Apache CarbonData :: Format	2.1 秒
Success Apache CarbonData :: Hadoop	7.4 秒
Success Apache CarbonData :: Parent	2.1 秒
Success Apache CarbonData :: Processing	4.6 秒
Success Apache CarbonData :: Spark	4 分 19 秒

ravipesala:
LGTM

PR #90, opened 2016-08-24
[CARBONDATA-175]Fix Bug: Load Decimal Data to BigInt Column, query is NULL

lion-x:
https://issues.apache.org/jira/browse/CARBONDATA-175

When Loading decimal Data into a BigInt Column,  the original flow just use Long.valueof to parse, so we get null value, now  add a preprocessing step to get the integer part of Decimal.


core/src/main/java/org/apache/carbondata/core/util/DataTypeUtil.java
```diff
@@ -62,7 +62,12 @@ public static Object getMeasureValueBasedOnDataType(String msrValue, DataType da
       case INT:
         return Double.valueOf(msrValue).longValue();
       case LONG:
-        return Long.valueOf(msrValue);
+        int index = msrValue.indexOf(".");
+        if (index != -1) {
+          return Long.valueOf(msrValue.substring(0,index));
+        } else {
+          return Long.valueOf(msrValue);
+        }
```
manishgupta88: @lion-x ...keep the code same for INT and LONG datatype case...i think instead of writing this code, that would be better. Sample code as below
case INT:
case LONG:
  return Double.valueOf(msrValue).longValue();

lion-x: @manishgupta88 Hi, in this PR, I want to solve the problem described as below:

when the Loding msrValue is decimal like "3.1415926" and the column type is Long, the original flow will not work well. It will return us a null value.

manishgupta88: @lion-x ...It is a correct behavior to return null value in this case as our validation is strict. If any column which has BigInt datatype is included as dimension then also we make decimal value as null because of strict validation. 
That is why I suggested that dont change the behavior for Long parsing, instead change the parsing logic for Int datatype to keep the strict validation behavior.

Zhangshunyu: @lion-x @manishgupta88 What about hive's behavior for this issue?

lion-x:
It is not a defect, no need to fixed. Close PR.

PR #92, opened 2016-08-24
[CARBONDATA-176] Deletion of compacted segment is not allowed.

ravikiran23:
once the segment is compacted then if user selects the compacted segment to delete then no need to delete the segment.

displaying error msg on this.

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
+++ b/processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
@@ -413,7 +413,7 @@ public class SegmentStatusManager {
           // if the segment is compacted then no need to delete that.
           if (CarbonCommonConstants.SEGMENT_COMPACTED
               .equalsIgnoreCase(loadMetadata.getLoadStatus())) {
-            LOG.error("Cannot delete the load which is compacted.");
+            LOG.error("Cannot delete the Segment which is compacted. Segment is " + loadId);
             loadFound = true;
             invalidLoadIds.add(loadId);
             break;
@@ -422,14 +422,14 @@ public class SegmentStatusManager {
             loadFound = true;
             loadMetadata.setLoadStatus(CarbonCommonConstants.MARKED_FOR_DELETE);
             loadMetadata.setModificationOrdeletionTimesStamp(readCurrentTime());
-            LOG.info("LoadId " + loadId + " Marked for Delete");
+            LOG.info("Segment ID " + loadId + " Marked for Delete");
           }
           break;
         }
       }
 
       if (!loadFound) {
-        LOG.audit("Delete load by Id is failed. No matching load id found.");
+        LOG.audit("Delete segment by ID is failed. No matching segment id found.");
         invalidLoadIds.add(loadId);
         return invalidLoadIds;
       }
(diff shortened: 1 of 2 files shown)


kumarvishal09:
LGTM

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
```diff
@@ -410,6 +410,14 @@ public void writeLoadDetailsIntoFile(String dataLoadLocation,
       for (LoadMetadataDetails loadMetadata : listOfLoadFolderDetailsArray) {
 
         if (loadId.equalsIgnoreCase(loadMetadata.getLoadName())) {
+          // if the segment is compacted then no need to delete that.
+          if (CarbonCommonConstants.SEGMENT_COMPACTED
+              .equalsIgnoreCase(loadMetadata.getLoadStatus())) {
+            LOG.error("Cannot delete the load which is compacted.");
```
Vimal-Das: In the log, mention the load id for which you are giving this error.

kumarvishal09:
@Vimal-Das it is really required? As we have one ddl to know how many compacted segments are present and as once segment are compacted it will not get deleted and in the log same segment id will be printed every time. Please comment.

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
ravikiran23: Here logs will be only in case of user is trying to delete the compacted loads intentionally using delete segment DDL.  so it is ok to add the segment ID in logs.
Fixing that.

kumarvishal09:
@Vimal-Das sorry for invalid commnet. @Vimal-Das @ravikiran23 As this is for segment deletion, so log will for the segment id which user wants to delete , so as user knows which segment he is deleting then printing the segment id in the log wont be useful :) anyway we can add the log in future may we will be requiring this.

kumarvishal09:
@Vimal-Das we can merge this PR

ravikiran23:
@kumarvishal09  @Vimal-Das Handled the scenario of deletion by date also. please check it once.

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
```diff
@@ -410,18 +410,26 @@ public void writeLoadDetailsIntoFile(String dataLoadLocation,
       for (LoadMetadataDetails loadMetadata : listOfLoadFolderDetailsArray) {
 
         if (loadId.equalsIgnoreCase(loadMetadata.getLoadName())) {
+          // if the segment is compacted then no need to delete that.
+          if (CarbonCommonConstants.SEGMENT_COMPACTED
+              .equalsIgnoreCase(loadMetadata.getLoadStatus())) {
+            LOG.error("Cannot delete the Segment which is compacted. Segment is " + loadId);
+            loadFound = true;
+            invalidLoadIds.add(loadId);
```
ManoharVanam: Above two lines are not required as we are deleting  all or none

ravikiran23: fixed.

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
```diff
@@ -449,6 +457,12 @@ public void writeLoadDetailsIntoFile(String dataLoadLocation,
     for (LoadMetadataDetails loadMetadata : listOfLoadFolderDetailsArray) {
       Integer result = compareDateValues(loadMetadata.getLoadStartTimeAsLong(), loadStartTime);
       if (result < 0) {
+        if (CarbonCommonConstants.SEGMENT_COMPACTED
```
ManoharVanam: Add test case for this scenario

ravikiran23: handled

ManoharVanam:
LGTM

Vimal-Das:
LGTM

gvramana:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/65/

PR #94, opened 2016-08-25
[CARBONDATA-178]Fix bug that table not exist when execute show segments using spark-sql and beeline the same time.

Zhangshunyu:
## Why raise this pr:
1 When using beeline and sparksql the same time, if create a table and load data into it by sparksql, beeline and sparksql would see the same table by show tables, but if execute show segments by beeline, it would throws exception that this table is not exists.
2. But if restart beeline or select the table before show segements, it is OK.

## How to solve this:
The problem is that beeline and sparksql using different process, they have different METADATAINSTANCE and tableInfoMap, beeline will not get the table from his own map, althouh sparksql put his table into the tableInfoMap.
**So, we can use tableExists to check, checkSchemasModifiedTimeAndReloadTables in tableExists would check the "modifiedTime.mdt", if it is change by different process(here is spark sql), the other process(here is beeline) should reload the metadata firstly and tableInfoMap would be refilled.**


Zhangshunyu:
CI has passed：
image

ravipesala:
LGTM

PR #95, opened 2016-08-25
[CARBONDATA-152]Double min max difference compression issue.

ashokblend:
1.While compressing maxdifference typecast with proper DataType
2.While writting we follow like
a. diff with max value , say d = max - x
b. multiply with 10^decimal i.e w = d*10^decimal
3.Similarly while reading we should follow like
a. divide written value 10^decimal i.e d = w/10^decimal
b. diff with max i.e x = max -d
4. There is problem in java with double subtraction
i.e double value = 3.141818 - 0.000610
java will result 3.1412080000000002
To avoid this we are converting to BigDecimal and then doing subtraction
5.Changed logic to calculate no of integer after decimal.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
@@ -19,6 +19,8 @@
 
 package org.apache.carbondata.core.datastorage.store.compression.type;
 
+import java.math.BigDecimal;
+
 import org.apache.carbondata.common.logging.LogService;
 import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.datastorage.store.compression.Compressor;
@@ -93,7 +95,9 @@ public class UnCompressNonDecimalMaxMinByte
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimalVal);
+        BigDecimal diff = BigDecimal.valueOf(value[i] / Math.pow(10, decimalVal));
+        BigDecimal max = BigDecimal.valueOf(maxValue);
+        vals[i] = max.subtract(diff).doubleValue();
       }
 
     }
--- a/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinDefault.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinDefault.java
@@ -19,6 +19,7 @@
 
 package org.apache.carbondata.core.datastorage.store.compression.type;
 
+import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 
 import org.apache.carbondata.common.logging.LogService;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
```diff
@@ -93,7 +95,9 @@
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimalVal);
+        BigDecimal diff = new BigDecimal(Double.toString(value[i] / Math.pow(10, decimalVal)));
```
Vimal-Das: use BigDecimal.valueOf(double) instead of converting to string for performance and dataloss concerns.refer http://www.javaworld.com/article/2073176/caution--double-to-bigdecimal-in-java.html

core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinDefault.java
```diff
@@ -95,7 +96,9 @@
       if (value[i] == 0) {
         vals[i] = maxVal;
       } else {
-        vals[i] = (maxVal - value[i]) / Math.pow(10, decimal);
+        BigDecimal diff = new BigDecimal(Double.toString(value[i] / Math.pow(10, decimal)));
```
Vimal-Das: use BigDecimal.valueOf(double) instead of converting to string for performance and dataloss concerns.refer http://www.javaworld.com/article/2073176/caution--double-to-bigdecimal-in-java.html

core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinFloat.java
```diff
@@ -98,7 +99,9 @@
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimal);
+        BigDecimal diff = new BigDecimal(Double.toString(value[i] / Math.pow(10, decimal)));
```
Vimal-Das: use BigDecimal.valueOf(double) instead of converting to string for performance and dataloss concerns.refer http://www.javaworld.com/article/2073176/caution--double-to-bigdecimal-in-java.html

core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinInt.java
```diff
@@ -93,7 +94,9 @@
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimal);
+        BigDecimal diff = new BigDecimal(Double.toString(value[i] / Math.pow(10, decimal)));
```
Vimal-Das: use BigDecimal.valueOf(double) instead of converting to string for performance and dataloss concerns.refer http://www.javaworld.com/article/2073176/caution--double-to-bigdecimal-in-java.html

Zhangshunyu:
If the column datatype is double and it data is like this:
-7489.7976
-11234567490
-11234567490
-1.2
-2
-11234567490
-11234567490
-11234567490
-11234567490
the query result would be all 0 using your pr, please verify this scenario.

Zhangshunyu:
LGTM，and i l have fixed
 and will raise another pr for the issue i mentioned above.

ashokblend:
@Vimal-Das review comment is handled. please check.

ravipesala:
LGTM

PR #96, opened 2016-08-25
[CARBONDATA-179] Describe table show old table schema

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-179

this is because after drop table , the old table info in cachedDataSourceTables(which is in spark part) is not cleaned, so we can clean it manually.


kumarvishal09:
LGTM

PR #98, opened 2016-08-26
[CARBONDATA-183] Blocks are allocated to single node when Executors configured is based on the ip address.

mohammadshahidkhan:


kumarvishal09:
LGTM

PR #99, opened 2016-08-26
[CARBONDATA-184] Fixed special char delimiter for complex data.

nareshpr:
Reason : When complex types data is having ^ delimiter, data is not getting loaded properly. When the data is queried, output is NULL
eg: array<string> and complex data level1 delimiter is '^' 
actual data is 
john^carter^justin
data is not getting split and not getting loaded & when queried, its give NULL value.


kumarvishal09:
LGTM

nareshpr:
Thanks for reviewing Vishal.
PR Builder URL
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/11/

PR #101, opened 2016-08-27
When query result is null, getPartitions function should return empty Arrary[Partition]

foryou2030:
# Why rasied this pr?
For query, it will call CarbonScanRdd, when used filter, if getSplits result is empty, then no need to compute

# How to solve?
when getSplits result is empty, no need to add "new CarbonSparkPartition" to result list


foryou2030:
Passed CI test, pls check   http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/29/

Vimal-Das:
LGTM

PR #102, opened 2016-08-29
[CARBONDATA-186] Except compaction all other alter operations on carbon table will be unsupported.

nareshpr:
Reason: As Carbon table will not support alter operations except compaction, all the alter operations on carbon table should be skipped and error message should be displayed as "Unsupported alter operation on carbon table"
Whereas if the alter operation is on hive table, it should be transferred to hive for performing the operation.


nareshpr:
PR Builder URL is 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/31/

gvramana:
LGTM

PR #106, opened 2016-08-29
[CARBONDATA-160]Data mismatch issue in case of multiple segment with different dictionary key size

kumarvishal09:
Problem:In case of multiple segment and each segment keygenertaor key size is not same in that case when we are converting the surrogate key to mdkey values which are not present in the older segments using keygenertaor in that case  keygenerator is giving some values which is less than column cardinality 

Solutions: Need add only those surrogate keys which is less than equal to cardinality of the segment

### DIFF
--- a/core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
+++ b/core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
@@ -23,18 +23,7 @@ import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
 
 import org.apache.carbondata.common.logging.LogService;
 import org.apache.carbondata.common.logging.LogServiceFactory;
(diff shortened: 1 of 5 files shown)


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/37/

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
```diff
@@ -623,9 +655,11 @@ public static DimColumnFilterInfo getFilterListForAllMembersRS(Expression expres
...
       for (Integer surrogate : dimColumnFilterInfo.getFilterList()) {
         try {
-          keys[carbonDimension.getKeyOrdinal()] = surrogate;
-          filterValuesList
-              .add(getMaskedKey(rangesForMaskedByte, blockLevelKeyGenerator.generateKey(keys)));
+          if (surrogate <= dimColumnsCardinality[carbonDimension.getKeyOrdinal()]) {
+            keys[carbonDimension.getKeyOrdinal()] = surrogate;
+            filterValuesList
+                .add(getMaskedKey(rangesForMaskedByte, blockLevelKeyGenerator.generateKey(keys)));
+          }
```
ravipesala: Please break the loop if it greater since the surrogates are always in sorted order here.

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
```diff
@@ -23,7 +23,18 @@
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.ArrayList;
```
ravipesala: use `import java.util.*;`

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
```diff
@@ -57,7 +68,17 @@
 import org.apache.carbondata.scan.expression.conditional.ListExpression;
 import org.apache.carbondata.scan.expression.exception.FilterIllegalMemberException;
 import org.apache.carbondata.scan.expression.exception.FilterUnsupportedException;
-import org.apache.carbondata.scan.filter.executer.*;
+import org.apache.carbondata.scan.filter.executer.AndFilterExecuterImpl;
```
ravipesala: Use `import org.apache.carbondata.scan.filter.executer.*;`

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
kumarvishal09: ok

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
kumarvishal09: ok

core/src/main/java/org/apache/carbondata/scan/filter/FilterUtil.java
kumarvishal09: ok

ravipesala:
LGTM except few minor comments

kumarvishal09:
@ravipesala Please check I have fixed all the review comments. http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/49/

ravipesala:
LGTM

Vimal-Das:
retest this please

PR #107, opened 2016-08-29
[CARBONDATA-191]Fix bug when quote char is single without new Line in the end of file

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-191

## Reason:
this bug is because that in dictionary part, in spark-csv reader, a new line '\n' is added to the end of every record:
    
     if (n == length - next) {
          str.getChars((next - start).toInt, (next - start + n - 1).toInt, cbuf, off)
          cbuf(off + n - 1) = '\n'
        } else {

and then in  univocity-parser,  quote char is handled in following logic: 

    while (!(prev == quote && (ch <= ' ' || ch == delimiter || ch == newLine))) {
			if (ch != quote && ch != quoteEscape) {
	                    if (prev == quote) { /
	  .......
						break;
          .....
	  			output.appender.append(ch);
	  			prev = ch;**

That means, only when get the second quote char and the latter char is delimiter or new line, then  the loop can be get out and stop the parse for quote char, then '\n' in the end is removed. but in the case provided in CARBONDATA-191,  delimiter and new line is not provided, loop can not get out, so the '\n' added in spark-csv parser will be treated as a char in record added to field.

but in the load csvInput step, the data is parsed without '\n' added automatically, so the actual data and dictionary value is different, then it becomes null.

## solution:
Rewrite the spark-csv parser, for the last line, do not add '\n' automatically.

### DIFF
--- a/integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithSingleQuotechar.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithSingleQuotechar.scala
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.carbondata.integration.spark.testsuite.dataload
+
+import org.apache.spark.sql.{DataFrame, Row}
+import org.apache.spark.sql.common.util.CarbonHiveContext._
+import org.apache.spark.sql.common.util.QueryTest
+import org.scalatest.BeforeAndAfterAll
+
+/**
+ * Test Class for data loading when there is single quote in fact data
+ *
+ */
+class TestLoadDataWithSingleQuotechar extends QueryTest with BeforeAndAfterAll {
+  override def beforeAll {
+    sql("DROP TABLE IF EXISTS carbontable")
+    sql(
+      "CREATE TABLE carbontable (id Int, name String) STORED BY 'carbondata'")
+  }
+
+  test("test data loading with single quote char") {
+    try {
+      sql(
+        "LOAD DATA LOCAL INPATH './src/test/resources/dataWithSingleQuote.csv' INTO TABLE " +
+          "carbontable OPTIONS('DELIMITER'= ',')")
+      checkAnswer(
+        sql("SELECT * from carbontable"),
+        Seq(Row("Tom",1),
+          Row("Tony\n3,Lily",2),
+          Row("Games\"",4),
+          Row("prival\"\n6,\"hello\"",5)
+        )
+      )
+    } catch {
+      case e: Throwable =>
+        assert(false)
+    }
+  }
+
+  override def afterAll {
+    sql("DROP TABLE carbontable")
+  }
+}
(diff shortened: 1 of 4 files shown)


Jay357089:
CI also passed.
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/40/

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithSingleQuotechar.scala
```diff
@@ -0,0 +1,60 @@
...
+      sql(
+        "LOAD DATA LOCAL INPATH './src/test/resources/dataWithSingleQuote.csv' INTO TABLE " +
+          "carbontable OPTIONS('DELIMITER'= ',')")
+      checkAnswer(
+        sql("SELECT * from carbontable"),
+        Seq(Row("Tom",1),
+          Row("Tony\n3,Lily",2),
+          Row("Games\"",4),
+          Row("prival\"\n6,\"hello\"",5)
+        )
```
manishgupta88: @Jay357089 ...Please perform the below mentioned activities.
1. Can you test the same scenario with same set of data in spark and hive to check their behavior and paste the result screenshot here.
2. What about cases when instead of default escape character (\), some other escape character is given (like @, ^, /, etc)? Test with some other escape characters in carbon, spark and hive and verify that behavior is same everywhere and paste the snapshots here.

Jay357089: ## i create hive hive table and load, just like below.

CREATE TABLE hiveTable(id int, name string)
ROW FORMAT
SERDE 'org.apache.hadoop.hive.serde2.OpenCSVSerde'
WITH SERDEPROPERTIES (
   "separatorChar" = ",",
   "quoteChar"     = "\"",
   "escapeChar"    = "\\"
)  
STORED AS TEXTFILE;

## and find that when data is like below, 
1,Tom
2,"Lily"
##  that query can get the correct result,
 +-----+-------+--+
| id  | name  |
+-----+-------+--+
| 1   | Tom   |
| 2   | Lily  |
+-----+-------+--+
## but if data is like below
1,Tom
2,"Tony
3,Lily"
## query result is like below
+-----+-------+--+
| id  | name  |
+-----+-------+--+
| 1   | Tom   |
| 2   | NULL  |
| 3   | NULL  |
+-----+-------+--+
## while carbon query is like below
+--------------+-----+--+
|     name     | id  |
+--------------+-----+--+
| Tom          | 1   |
| Tony
3,Lily  | 2   |
+--------------+-----+--+

so, now we can see that, hive can support quote char, but can not support multiline, while carbon can support both, this is the difference, and i think that carbon's query result is right, no need to keep same with hive.
## so, what we should only concern about is what i raise in [CARBONDATA-191].
@gvramana @manishgupta88 what is ur opionion?

manishgupta88:
LGTM...I think we can keep Carbon behavior different from hive as Carbon supports multiline....
@gvramana ...please comment...

ravipesala:
It seems many issues in spark-csv and we end up with copying most of there code to carbon and modify it. Instead we should directly use input format with univocity parser and use it for both generating dictionary and for loading data. 
Any way we can do this change while updating(removing kettle) the design of data loading.

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithSingleQuotechar.scala
gvramana: 1) previous version we were supporting multiline=true/false option in data load, but after moving to univocity parser, that option is discontinued and defaulted to multiline=true
2) What happens if this multiline is split across blocks, as textinputformat will skip first line of current block and will read one extra new line of next block. It might end up reading, same line twice. - can be checked once and can be tracked as separate issue.
3) future we can support option to set multiline=false for univocity parser, similar to hive.So this behavior of carbon is fine.
4) I think multiline+parse till quote+delimiter as expected behaviour needs to be mentioned in IDG document.

Jay357089: @gvramana for [CARBONDATA-191], this PR is OK, right? for other issue like multiline across blocks, multiline can be set and so on, we can track in another jira, yes?

sraghunandan:
+1
We go ahead with the current changes and take up Ravindra's suggestion in the req to remove kettle

gvramana:
LGTM, CSVInputFormat tracked through CARBONDATA-2

PR #108, opened 2016-08-29
[CARBONDATA-190]Data mismatch issue in case of filter query 

kumarvishal09:
Issue steps:1. create table , then restart the server and then do data load, in that case filter query record count is not matching.
Problem: When user is creating any table and if user has not disabled inverted index false for any key column we are setting the inverted index true in column schema object. As we are not persisting this information in schema file, so after restarting the server useInvertedIndex property is false in columnschema object and in data loading column data is not sorted and in filter execution we are doing binary search, as data is not sorted binary search is failing and it is skipping some of the record.
Solution : In this pr default value is set to true. One more PR will be raised to handle inverted index disabled sceario. By default Inverted index will be enabled for all the column for better query performance


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/38/

ravipesala:
LGTM

Vimal-Das:
LGTM

PR #109, opened 2016-08-30
[CARBONDATA-192] Invalidating the table from hive context while dropping the table

ManoharVanam:
Select Query is throwing exception in the below scenario : 
Create table table1
load table1
select * from table1
drop table1
create same table table1
load table1
select * from table1
throws exception


ravipesala:
@ManoharVanam Please rebase it

ManoharVanam:
@ravipesala done

ravipesala:
LGTM

PR #110, opened 2016-08-31
[CARBONDATA-193]Fix the bug that negative data compress is not properly when datatype is Double

Zhangshunyu:
## Why raise this pr?
**Fix bug: negative data compress is not properly when datatype is Double.**
For example, If the column datatype is double and it data is like this:
-7489.7976
-11234567490
-11234567490
-1.2
-2
-11234567490
-11234567490
-11234567490
-11234567490
**the query result would be all 0, this is a bug.**
## How to solve?
1. This bug is becasue we only consider the MAX value of this column is +values, we should use Math.max(Math.abs((double)maxValue), Math.abs((double)minValue)) to get the datatype instead of max, bacasue -1 > -1223434545454, but we shoudl use -1223434545454.
2. When we use the new max values, the former code only check +values, but never ensure the values should >= Datatype.Min, for example, when the max value is -200, it < Byte.MAX, the code let it use byte, but it <Byte.Min, this is wrong. We should both check the upper and lower limit for one datatype.

## How to test?
Added test case: test("When the values of Double datatype are negative values"), should pass all the exist cases and this new testcase, ensure carbon's output is the same to hive.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/ValueCompressionUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/ValueCompressionUtil.java
@@ -54,6 +54,7 @@ import org.apache.carbondata.core.datastorage.store.compression.type.UnCompressN
 import org.apache.carbondata.core.datastorage.store.compression.type.UnCompressNoneInt;
 import org.apache.carbondata.core.datastorage.store.compression.type.UnCompressNoneLong;
 import org.apache.carbondata.core.datastorage.store.compression.type.UnCompressNoneShort;
+import org.apache.carbondata.core.datastorage.store.impl.data.uncompressed.DoubleArrayDataFileStore;
 
 public final class ValueCompressionUtil {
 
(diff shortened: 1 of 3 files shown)


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/56/

Zhangshunyu:
CI has passed, please check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/58/

core/src/main/java/org/apache/carbondata/core/util/ValueCompressionUtil.java
```diff
@@ -78,26 +78,26 @@ private ValueCompressionUtil() {
   private static DataType getDataType(double value, int decimal, byte dataTypeSelected) {
     DataType dataType = DataType.DATA_DOUBLE;
     if (decimal == 0) {
-      if (value < Byte.MAX_VALUE) {
+      if (value < Byte.MAX_VALUE && value > Byte.MIN_VALUE) {
```
Vimal-Das: should we change to value <= Byte.MAX_VALUE && value >= Byte.MIN_VALUE ?

Zhangshunyu: Before i modified to use new "absMaxValue", it is needed, but now i can detele this code.

Zhangshunyu:
@Vimal-Das Fixed review comments, and re-build CI has passed, please check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/62/

Vimal-Das:
LGTM

PR #111, opened 2016-08-31
[CARBONDATA-194] ArrayIndexOfBoundException thrown when number of columns in row more than the max number of columns in univocity parser settings

manishgupta88:
ISSUE ID: https://issues.apache.org/jira/browse/CARBONDATA-194

Problem: When the number of columns in CSV data file while parsing a row are more than the max number of columns configured in the univocity parser settings, the parser throws array index of bound exception

Reason: Max number of columns in CSVParserSettings are set equivalent to the number of columns in schema with an addition of 10. if still the number of columns while parsing are more then the univocity parser throws array index of bound exception.

Solution: Configure a higher value for max number of columns and take the max of number of columns in schema and default max columns value while setting in univocity parser settings.

Impact: Data load flow

e.g Load DDL COmmand
**LOAD DATA LOCAL INPATH './src/test/resources/character_carbon.csv' into table valid_max_columns_test options('MAXCOLUMNS'='400')**


manishgupta88:
CI PAssed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/61/

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
```diff
@@ -41,6 +41,10 @@
 public class UnivocityCsvParser {
 
   /**
+   * Max number of columns that will be parsed for a row by univocity parsing
+   */
+  private static final int MAX_NUMBER_OF_COLUMNS_FOR_PARSING = 2000;
```
kumarvishal09: I think i will fail if csv file has more number of column than 2000 and schema you have selected less columns. Better expose one property so user can also configure max number of columns in csv file.

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
```diff
@@ -104,6 +108,20 @@ public void initialize() throws IOException {
...
+   *
+   * @param columnCountInSchema total number of columns in schema
+   * @return
+   */
+  private int getMaxColumnsForParsing(int columnCountInSchema) {
+    int maxNumberOfColumnsForParsing = columnCountInSchema;
+    if (columnCountInSchema < MAX_NUMBER_OF_COLUMNS_FOR_PARSING) {
+      maxNumberOfColumnsForParsing = MAX_NUMBER_OF_COLUMNS_FOR_PARSING;
+    }
+    return maxNumberOfColumnsForParsing;
```
gvramana: Add +10 if schema columns are considered.

gvramana:
Also have to check if any impact on reading performance

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
gvramana: If array creation is for every row, then we should surely control it.
If array creation is one time, then can give value of 2000 as default. and can add DataLoad command option to control it, mainly useful for wide tables

manishgupta88:
@gvramana 
1. Array creation will happen only one time for the size specified and every time data will be filled and read from that array. So looking at parsing code I dont think there will be any impact on reading performance.
2. If we dont set this option in univocity parser settings, then max columns will have a default value of 512.

manishgupta88:
CI passed : http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/70/

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
kumarvishal09: I added +10 because to avoid this bug, i think now no need to add 10 as we are allowing user to give max number of columns. @gvramana Please comment

kumarvishal09:
except one doubt(addning +10)..LGTM

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
sraghunandan: I think user will not be able to provide this.He may get a csv with more cols, but he is interested in few cols he is interested in.Depending on user would be bug prone

gvramana:
LGTM

PR #113, opened 2016-08-31
stop compaction flow after metadata updation failed

QiangCai:
**1 Cause Analysis**
When the metadata updation failed,  it will do compaction again and again.

**2 Solution**
Throwing a exception to stop compaction flow.


Vimal-Das:
@QiangCai Can you add a test for this?

QiangCai:
I think It is not easy to simulate the error.

Vimal-Das:
@QiangCai thats fine, LGTM

PR #117, opened 2016-09-01
[CARBONDATA-199]Fix the bug that when subquery with sort and filter the result is empty.

Zhangshunyu:
## Why raise this pr?
Fix this bug: When the query has subquery with sort and filter, it can not return resullt.
## How to solve?
When the query likes this, the optimized plan by spark never push down the filter, and as aresult the sort is not decoded by carbon, when use filter, the int values can not resolved as string values by spark.
So we shoud decode them earlier **when the child of filter is sort**.
## How to test?
Added new testcases and should pass them and pass CI.


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/80/

ravipesala:
LGTM

PR #121, opened 2016-09-02
[CARBONDATA-203]Use static string to set Hadoop configuration

zuowang:
Should use static string in FileInputFormat class as the configuration key to set the corresponding configuration value in Hadoop Configuration

Some places currently is not doing like this. This PR made a fix.


ravipesala:
LGTM 
@zuowang Thanks for working on it.

Vimal-Das:
LGTM

PR #123, opened 2016-09-02
[CARBONDATA-204] Clear queryStatisticsMap when timeout for query statistic

foryou2030:
# Why raise this pr?
I found Query statistics issue:
1. some query statistics that never be printed will be keeped into querystatisticsMap, so it will cause "out of memory" for long time running
2. in some sceniaro, the driver can't record "sql_parse_time" , the driver statistics logs will not be output, we should output  block_allocation_time and block_identification_time always.
# How to solve?
1. add function to check querystatistics timeout , once timeout, remove the queryId from the map.
2. add conditional detection for queryStatisticsMap size, if the queryStatistic only contain block_allocation_time and block_identification_time, then ouput them.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
+++ b/core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
@@ -102,7 +102,6 @@ public class DriverQueryStatisticsRecorder {
               queryStatisticsMap.remove(queryId);
             }
           }
-          
           // clear the timeout query statistics
           long interval = System.nanoTime() - Long.parseLong(queryId);
           if (interval > QueryStatisticsConstants.CLEAR_STATISTICS_TIMEOUT) {
--- a/core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
+++ b/core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
@@ -90,22 +90,23 @@ public class DriverQueryStatisticsRecorder {
         String queryId = entry.getKey();
         // clear the unknown query statistics
         if(StringUtils.isEmpty(queryId)) {
-          queryStatisticsMap.remove(queryId);
+          entries.remove();
         } else {
-          // print sql_parse_t,load_meta_t,block_allocation_t,block_identification_t
-          // or just print block_allocation_t,block_identification_t
-          if (entry.getValue().size() >= 2) {
-            String tableInfo = collectDriverStatistics(entry.getValue(), queryId);
-            if (null != tableInfo) {
-              LOGGER.statistic(tableInfo);
-              // once the statistics be printed, remove it from the map
-              queryStatisticsMap.remove(queryId);
-            }
-          }
           // clear the timeout query statistics
           long interval = System.nanoTime() - Long.parseLong(queryId);
           if (interval > QueryStatisticsConstants.CLEAR_STATISTICS_TIMEOUT) {
-            queryStatisticsMap.remove(queryId);
+            entries.remove();
+          } else {
+            // print sql_parse_t,load_meta_t,block_allocation_t,block_identification_t
+            // or just print block_allocation_t,block_identification_t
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
```diff
@@ -78,106 +82,142 @@ public synchronized void recordStatisticsForDriver(QueryStatistic statistic, Str
    */
   public void logStatisticsAsTableDriver() {
     synchronized (lock) {
-      String tableInfo = collectDriverStatistics();
-      if (null != tableInfo) {
-        LOGGER.statistic(tableInfo);
+      for (String key: queryStatisticsMap.keySet()) {
+        // print sql_parse_t,load_meta_t,block_allocation_t,block_identification_t
+        // or just print block_allocation_t,block_identification_t
+        if (queryStatisticsMap.get(key).size() >= 2) {
```
Vimal-Das: The call can return null, because the keyset obtained reflects the state of map when the .keySet() method was called, later changes in the map will not be reflected. So, if the element is removed in the meantime, get() can return null.

Solution: use an iterator over entrySet()

core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
```diff
@@ -78,106 +82,142 @@ public synchronized void recordStatisticsForDriver(QueryStatistic statistic, Str
...
+    String sql_parse_time = "";
+    String load_meta_time = "";
+    String block_allocation_time = "";
+    String block_identification_time = "";
+    Double driver_part_time_tmp = 0.0;
+    Double driver_part_time_tmp2 = 0.0;
+    String splitChar = " ";
+    try {
+      // get statistic time from the QueryStatistic
+      for (QueryStatistic statistic : queryStatisticsMap.get(key)) {
```
Vimal-Das: null possibility here also, pass entry.getValue() instead, once you iterate over entrySet() in the calling method

foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/105/

core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
foryou2030: ok, thanks
handled, pls check

core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
foryou2030: ok, handled, pls check

core/src/main/java/org/apache/carbondata/core/carbon/querystatistics/DriverQueryStatisticsRecorder.java
```diff
@@ -78,106 +83,148 @@ public synchronized void recordStatisticsForDriver(QueryStatistic statistic, Str
...
-      if (null != tableInfo) {
-        LOGGER.statistic(tableInfo);
+      Iterator<Map.Entry<String, List<QueryStatistic>>> entries =
+              queryStatisticsMap.entrySet().iterator();
+      while (entries.hasNext()) {
+        Map.Entry<String, List<QueryStatistic>> entry = entries.next();
+        String queryId = entry.getKey();
+        // clear the unknown query statistics
+        if(StringUtils.isEmpty(queryId)) {
+          queryStatisticsMap.remove(queryId);
```
Vimal-Das: use the Iterator.remove() for better safety

foryou2030: I tried this, but caused some exceptions

foryou2030: ok, handled

Vimal-Das:
@foryou2030  LGTM, Thanks for working on this.

PR #126, opened 2016-09-05
[CARBONDATA-212] Use SQLContext to read CarbonData files

jackylk:
In this PR, user can use SQLContext instead of CarbonContext to read CarbonData files.
Example:
```
    // Save a DataFrame to CarbonData files
    df.write
      .format("carbondata")
      .option("tableName", "carbon1")
      .mode(SaveMode.Overwrite)
      .save()

    // use SQLContext to read CarbonData files
    val sqlContext = new SQLContext(sc)
    sqlContext.sql(
      """
        | CREATE TEMPORARY TABLE source
        | (c1 string, c2 string, c3 long)
        | USING org.apache.spark.sql.CarbonSource
        | OPTIONS (path './examples/target/store/default/carbon1',       // file path
      """.stripMargin)
    sqlContext.sql("SELECT c1, c2, count(*) FROM source WHERE c3 > 100 GROUP BY c1, c2").show
```

User can also create table without providing schema, schema will be read from the CarbonData file:
```
    sqlContext.sql(
      """
        | CREATE TEMPORARY TABLE source
        | USING org.apache.spark.sql.CarbonSource
        | OPTIONS (path './examples/target/store/default/carbon1',       // file path
      """.stripMargin)
```

Save to CarbonData file is still using CarbonContext, but in read path, it can use SQLContext datasource API.


jackylk:
@ravipesala @QiangCai

examples/src/main/scala/org/apache/carbondata/examples/DatasourceExample.scala
```diff
@@ -0,0 +1,55 @@
...
+      .save()
+
+    // use SQLContext to read CarbonData files
+    val sqlContext = new SQLContext(sc)
+    sqlContext.sql(
+      """
+        | CREATE TEMPORARY TABLE source
+        | (c1 string, c2 string, c3 long)
+        | USING org.apache.spark.sql.CarbonSource
+        | OPTIONS (path './examples/target/store',
```
QiangCai: please change path to './target/store'

integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
```diff
@@ -123,24 +120,24 @@ class CarbonSource
       dataSchema: Option[StructType],
       partitionColumns: Option[StructType],
       parameters: Map[String, String]): HadoopFsRelation = {
-    CarbonDatasourceHadoopRelation(sqlContext, paths, parameters)
+    CarbonDatasourceHadoopRelation(sqlContext, paths, parameters, dataSchema)
```
QiangCai: You can validate the parameters here.

jackylk:
@QiangCai please check

jackylk:
@Vimal-Das can you check this?

jackylk:
Please do not merge now, still there are test case failure need to solve

chenliang613:
Looks great!
I have one query: Supposed there are many columns(100+), how to automatically get columns' definition, to avoid one by one input again with the below script, because load data already did the works.

        | CREATE TEMPORARY TABLE source
        | (c1 string, c2 string, c3 long)
        | USING org.apache.spark.sql.CarbonSource
        | OPTIONS (path './examples/target/store/default/carbon1',       // file path
      """.stripMargin)

JihongMA:
this is good idea to get rid of CarbonContext and go back to normal SqlContext..  for the writing part, the new design for data loading will handle the case of saving a DF to Carbon table without using CarbonContext? 

for the reading part, in the future, we can add following df API integration, so that there is no need to supply table schema/option at the time of querying it, those info (schema, carbon properties) can be retrieved from the meta data of the table root path. 
df = sqlContext.read.format("Carbon").options("Carbon options").load(" path/to/carbon file")
df.registerTempTable()

I noticed Spark 2.0 has introduced new API for registering temporary table..

jackylk:
@JihongMA Yes, for the write part we should refactor to use SqlContext after new flow of data loading is complete. For the read part, I plan to raise another PR to read schema from table path, such that user is not required to give schema in create table DDL

jackylk:
@JihongMA @chenliang613 added support for create table without providing schema
```
    sqlContext.sql(
      """
        | CREATE TEMPORARY TABLE source
        | USING org.apache.spark.sql.CarbonSource
        | OPTIONS (path './examples/target/store/default/carbon1',       // file path
      """.stripMargin)
```

chenliang613:
verified ok: don't need to provide schema,  can automatically get metadata information.

LGTM

PR #127, opened 2016-09-06
[CARBONDATA-213] Remove dependency: thrift complier

QiangCai:
[CARBONDATA-213] Remove dependency: thrift complier

**analysis**

I think it unnecessary for user/developer to download thrift complier When building CarbonData project.

**solution**

Provide the java code, generated by thrift complier.


Vimal-Das:
@QiangCai Good idea. 
+1 for having two maven profiles. One using thrift compiler and one using java code. Also a profile in "format" to copy generated code to "format/format-java"

jackylk:
+1
I think this will also solve the problem that currently IDE could not build carbondata project because of thrift tool dependency.

chenliang613:
@QiangCai please rebase this PR

QiangCai:
@chenliang613 @jackylk 
done

pom.xml
```diff
@@ -93,7 +93,7 @@
 
   <modules>
```
Vimal-Das: Please **add** a maven profile in the main pom which will involve the format/format instead of format/format-java
This will be helpful to run in CI where we would also want to ensure the thrift files and the source code are in sync.

Vimal-Das:
@QiangCai Thanks for talking this up, but it would be nice to have one additional maven profile for running in CI where we can first trigger compile of format/format, so that the CI can also verify that the thrift files are in sync with the code.

QiangCai:
@Vimal-Das 
Added format profile.
We can use it as following.
mvn package -DskipTests -Pformat 

It will refresh format-java module's code in generate-sources phase,  before format-java module's compilation.

jackylk:
@QiangCai I think this PR can be closed

PR #130, opened 2016-09-07
[CARBONDATA-217]Fixed data mismatch issue after compaction

kumarvishal09:
Step1: Run Data load and restart server
Step2: Start compaxction
Step3: Run Query
Query is giving different result after compaction
Problem:In compaction flow we are not setting the useInvertedIndex boolean for key column so it is false and it is not writing column data in sorted order and binary search is failing 
Solution: Need to set the useInvertedIndex array from carbon table


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/122/

ravipesala:
@kumarvishal09 Better save the `useInvertedIndex` to store.

kumarvishal09:
@ravipesala I have raised one jira for no inverted index scenario but this issue will still come if we fix inverted index scenario https://issues.apache.org/jira/browse/CARBONDATA-221. Please Assign this issue to some one

ravipesala:
LGTM

PR #133, opened 2016-09-07
[CARBONDATA-219] If the compaction is triggered without data load failing.

ravikiran23:
If the compaction is triggered without data load failing. as the array access is wrong.


sraghunandan:
LGTM

Vimal-Das:
@ravikiran23 LGTM, thanks for working on this.

PR #134, opened 2016-09-07
[CARBONDATA-220] TimeStampDirectDictionaryGenerator_UT.java is not running in the build

mohammadshahidkhan:
# **Why raise this pr?**
Our build script does not execute the test case if the class name does not have Test or TestCase 
There was no boundary value test case for the timestamp datatype.
# 
# **How to resolve?**
We have to refactor the class and have to add the boundary test cases.


mohammadshahidkhan:
PR builder passed 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/org.apache.carbondata$carbondata-processing/130

gvramana:
LGTM

PR #135, opened 2016-09-07
[CARBONDATA-226]changing the error message in delete by ID.

ravikiran23:
WHY RAISED THIS PR ?
when the delete load by ID DDL is called by the user on the compacted segment then the error message displayed in the beeline is wrong. 
There are 2 scenarios in this:
1. deletion of the compacted segment.
2. deletion of segment doesnt exist/deleted segment.

so in both cases the error message should be valid.

Solution : 
logging the specific error message in the logs.
giving a general message in the beeline error.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
@@ -935,7 +935,7 @@ private[sql] case class DeleteLoadsById(
     }
     else {
       sys.error("Delete segment by Id is failed. Invalid ID is :"
-                + invalidLoadIds.mkString(",") + ". Please check the log.")
+                + invalidLoadIds.mkString(","))
     }
 
     Seq.empty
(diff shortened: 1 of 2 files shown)


ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/141/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -931,11 +931,11 @@ private[sql] case class DeleteLoadsById(
...
     if (invalidLoadIds.isEmpty) {
 
-      LOGGER.audit(s"Delete load by Id is successfull for $databaseName.$tableName.")
+      LOGGER.audit(s"Delete segment by Id is successfull for $databaseName.$tableName.")
     }
     else {
-      sys.error("Delete load by Id is failed. No matching load id found. SegmentSeqId(s) - "
-                + invalidLoadIds)
+      sys.error("Delete segment by Id is failed. Invalid ID is :"
+                + invalidLoadIds.mkString(",") + ". Please check the log.")
```
gvramana: remove please check log message.

gvramana:
LGTM

PR #137, opened 2016-09-07
[CARBONDATA-222]  Handled query issue for all dimensions are no dictionary columns

foryou2030:
# Why raise this pr?
query failed when all dimensions are no-dictionary column, and the data contains one row that all no-dictionary column value are empty
# How to resolve?
when find First Data Block, we need to compare indexes, in this scenario,we should use other compareTo method.


foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/135/

core/src/main/java/org/apache/carbondata/core/carbon/datastore/impl/btree/BTreeDataRefNodeFinder.java
```diff
@@ -240,9 +240,16 @@ private int compareIndexes(IndexKey first, IndexKey second) {
               firstNoDictionaryKeyBuffer.getShort(nonDictionaryKeyOffset + SHORT_SIZE_IN_BYTES);
           secondNodeDictionaryLength =
               secondNoDictionaryKeyBuffer.getShort(nonDictionaryKeyOffset + SHORT_SIZE_IN_BYTES);
-          compareResult = ByteUtil.UnsafeComparer.INSTANCE
-              .compareTo(first.getNoDictionaryKeys(), actualOffset, firstNoDcitionaryLength,
-                  second.getNoDictionaryKeys(), actualOffset, secondNodeDictionaryLength);
+          int minLength = Math.min(firstNoDcitionaryLength, secondNodeDictionaryLength);
```
gvramana: @kumarvishal09 please comment on this. I am not getting how actualOffset is same for both keybuffers. 
Also as I know length can be calculated by nextOffset-currentOffset. Please give the structure of data stored and how this function works

kumarvishal09: @gvramana Yes you right current logic has some problem we need to maintain offset of both the keys and when it is not the last key we need to substract the secondNodeDictionaryLength - actualOffset1 and firstNodeDictionaryLength-actualOffset.

foryou2030:
CI passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/145/

kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/145/

kumarvishal09:
LGTM

core/src/main/java/org/apache/carbondata/core/carbon/datastore/impl/btree/BTreeDataRefNodeFinder.java
foryou2030: ok, handled. thanks for you @gvramana @kumarvishal09

PR #139, opened 2016-09-07
[CARBONDATA-224]Fixed data mismatch issue in case of Dictionary Exclude column for Numeric data type

kumarvishal09:
Problem: In case of greater than query on dictionary exclude column of for numeric data type
This is because data is sorted based on string because of this if data is 1,10,2,3 , data will be sorted like 1,10,2,3 but if we search greater than 3 then while applying min max will return false as last value is 3
Solution:we need to sort based on actual data type for this we should have chain comparator based on data type while loading the data, currently disabling DictionaryExclude column for numeric data type and will throw exception. Will raise jira issue to for sorting the based on actual data type


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/139/

kumarvishal09:
raise Jira for handling data type based support https://issues.apache.org/jira/browse/CARBONDATA-225

ravipesala:
@kumarvishal09 Please add testcase

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -864,8 +864,8 @@ class CarbonSqlParser()
    /**
     * detects whether double or decimal column is part of dictionary_exclude
     */
-  def isDoubleDecimalColDictionaryExclude(columnDataType: String): Boolean = {
-    val dataTypes = Array("double", "decimal")
+  def isNumberInColDictionaryExclude(columnDataType: String): Boolean = {
+     val dataTypes = Array("double", "decimal", "int")
```
gvramana: Add a reverse check that only string or (varchar-check) is only supported. As other datatypes like bigint etc will have problem in above check

kumarvishal09: ok

kumarvishal09:
@ravipesala Added testcase http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/195/

gvramana:
LGTM, merged to master, @kumarvishal09 Raise separate PR for branch_0.1 also

PR #140, opened 2016-09-08
[CARBONDATA-227] In block distribution parallelism is decided initial…

mohammadshahidkhan:
# **## Why raise this pr?**

In block distribution parallelism is decided initially and not re initialized after requesting new executors. Due to this task per node initialization is getting wrong.

# **# How to resolve?**
after requesting additional executors we have to reinitialize the defaultParallelism with the current parallelism value.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
@@ -73,9 +73,9 @@ class CarbonScanRDD[V: ClassTag](
     baseStoreLocation: String)
   extends RDD[V](sc, Nil) with Logging {
 
-  val defaultParallelism = sc.defaultParallelism
 
   override def getPartitions: Array[Partition] = {
+    var defaultParallelism = sparkContext.defaultParallelism
     val statisticRecorder = CarbonTimeStatisticsFactory.getQueryStatisticsRecorderInstance()
     val (carbonInputFormat: CarbonInputFormat[Array[Object]], job: Job) =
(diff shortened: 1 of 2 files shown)


integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -73,7 +73,7 @@ class CarbonScanRDD[V: ClassTag](
     baseStoreLocation: String)
   extends RDD[V](sc, Nil) with Logging {
 
-  val defaultParallelism = sc.defaultParallelism
+  var defaultParallelism = sc.defaultParallelism
```
gvramana: now member is not required, make it a private val.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -110,13 +110,15 @@ class CarbonScanRDD[V: ClassTag](
           new BlockletInfos(inputSplit.getNumberOfBlocklets, 0, inputSplit.getNumberOfBlocklets)
         )
       )
+      val activeNodes = DistributionUtil
+        .ensureExecutorsAndGetNodeList(blockListTemp.toArray, sparkContext)
```
gvramana: can u test it on dynamic executors case and confirm

mohammadshahidkhan:
PR builder Passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/144/

gvramana:
LGTM

PR #142, opened 2016-09-08
[CARBONDATA-221] Fix the bug of inverted index that store inverted index in metadata.

Zhangshunyu:
## Why raise this pr?
Inverted index in ddl info was not stored into store, and when we restart the culster, query might mismatch.
## How to solve?
Using the Encoding as the indentifier to check whether using inverted index.


Zhangshunyu:
CI passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/162
@ravipesala @kumarvishal09 @manishgupta88 
pls help to check this pr, whether this modifacation is proper that after restart cluster data mismatch would happen again?  Thank you!

PR #143, opened 2016-09-08
[CARBONDATA-229] Array Index of bound exception thrown from dictionary look up while writing sort index file

manishgupta88:
Problem: Array Index of bound exception thrown from dictionary look up while writing sort index file

Analysis: Whenever we load dictionary data into memory, then in case of populating reverse dictionary object sometimes a chunk which has no value is also getting added to the dictionary chunk list. This is happening because the logic for dictionary chunk distribution in case of forward dictionary is not implemented for reverse dictionary and 0 size dictionary chunks are not getting removed while adding to the list of dictionary chunks.

Solution: Add the same distribution logic we have in forward dictionary for populating reverse dictionary object

Impact area: Sort index generation


manishgupta88:
CI Passed:- http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/155/

gvramana:
LGTM

PR #146, opened 2016-09-09
[CARBONDATA-231] Rename repeated table names in same test file and add drop tables.

Zhangshunyu:
## Why raise this pr?
1. In the same test file, too much test cases using the same table name, for example t3, it is not reasonable and difficult to distinguish test cases.

2.Should drop tables in before all and after all.

## How to solve?
Rename the table name and add drop tables.


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/170

jackylk:
LGTM

PR #148, opened 2016-09-11
[CARBONDATA-233] bad record logger support for non parseable numeric and timestamp type data

mohammadshahidkhan:
# **Why raise this pr?**
Currently non parseable numeric data we were storing as null. 
So the stored data was not giving the right picture of the actual data being loaded.
Based on configuration bad record should be written in the raw csv and will log the bad records in a log file with proper reason.
# **How to resolve?**
Instead of changing the wrong data to null we have tell to the user which data is wrong in a log file.
Bellow property will be exposed to handle the same.
**bad_records_logger_enable** -- to enable the bad record logger to write the reason in the log file. By default    logging will be false.
**bad_records_logger_action** -- This property has 3 option
**force** -- bad record data will be converted to null and will load them 
                   no csv file  be written (old behaviour)
**redirect** -- bad record  will not be converted to null and will be written to the raw csv
**ignore** -- bad record  will not be converted to null and will not be written to the raw csv


processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1133,6 +1149,9 @@ else if(isComplexTypeColumn[j]) {
                       .getDirectDictionaryGenerator(details.getColumnType());
               surrogateKeyForHrrchy[0] =
                   directDictionaryGenerator1.generateDirectSurrogateKey(tuple);
+              if (badRecordsLoggerEnable && surrogateKeyForHrrchy[0] == 1) {
+                surrogateKeyForHrrchy[0] = -1;
```
manishgupta88: Replace -1 and 1 with proper identifiers from CarbonCommonConstants to give proper meaning and improve code readability

mohammadshahidkhan:
PR build passed 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/182/

mohammadshahidkhan:
PR builder passed after modification:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/335/

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/BadRecordslogger.java
```diff
@@ -89,41 +95,55 @@ public static String hasBadRecord(String key) {
     return badRecordEntry.remove(key);
   }
 
-  public void addBadRecordsToBilder(Object[] row, int size, String reason, String valueComparer) {
+  public void addBadRecordsToBuilder(Object[] row, int size, String reason, String valueComparer,
+      boolean badRecordsLogRedirect, boolean badRecordLoggerEnable) {
     StringBuilder logStrings = new StringBuilder();
+    size = row.length;
```
sujith71955: better to validate whether we need to log the bad record in the starting of method, and then continue the log string creation logic

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/BadRecordslogger.java
```diff
@@ -89,41 +95,55 @@ public static String hasBadRecord(String key) {
...
+    }
   }
 
   /**
    *
    */
   private synchronized void writeBadRecordsToFile(StringBuilder logStrings) {
-    String filePath = this.storePath + File.separator + this.fileName
-        + CarbonCommonConstants.FILE_INPROGRESS_STATUS;
+    String filePath =
```
ravipesala: This line should be moved down under `if` condition

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -439,6 +439,22 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
               .setDimensionOrdinalToDimensionMapping(populateNameToCarbonDimensionMap());
         }
         serializationNullFormat = meta.getTableOptionWrapper().get("serialization_null_format");
+        badRecordsLoggerEnable =
+            Boolean.parseBoolean(meta.getTableOptionWrapper().get("bad_records_logger_enable"));
+        String bad_records_logger_action =
+            meta.getTableOptionWrapper().get("bad_records_logger_action");
+
+        badRecordConvertNullDisable = true;
+        if("force".equalsIgnoreCase(bad_records_logger_action)) {
```
sujith71955: move the bad logger action variables to an Enum.

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -853,8 +873,10 @@ private String getCarbonLocalBaseStoreLocation() {
 
   private Object[] process(Object[] r) throws RuntimeException {
     try {
+      Object[] actualRow = new Object[r.length];
+      System.arraycopy(r, 0, actualRow, 0, r.length);
```
ravipesala: Can't we avoid this object[] instance and arraycopy? 
I guess we can merge `changeNullValueToNullString` and `populateOutputRow` and avoid arraycopy.

ravipesala:
There are few object creations, arraycopies and equalIgnorecase are added for each row/field. It might impact actual loading performance irrespective of badrecordlogger is enabled or not.  Please try to avoid these operations.

mohammadshahidkhan:
handled all the comments

ravipesala:
LGTM. But please handle the checkstyle issues.

mohammadshahidkhan:
PR builder passed 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/381/
Please review the same

PR #155, opened 2016-09-14
[CARBONDATA-239] Handling the exception cases in compaction.

ravikiran23:
Why raise this pr?
If a compaction is triggered by the user on table1 , and other requests will go to queue. and if the compaction is failed for table1 then the requests in queue is not getting completed.

How to resolve?
If the compaction is failed then the requests in queue should continue the compaction.

if compaction is failed for table, requested by user then need to show error message in the beeline.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -32,11 +32,11 @@ import org.apache.hadoop.mapreduce.lib.input.{FileInputFormat, FileSplit}
 import org.apache.spark.{util => _, _}
 import org.apache.spark.sql.{CarbonEnv, SQLContext}
 import org.apache.spark.sql.execution.command.{AlterTableModel, CompactionCallableModel, CompactionModel, Partitioner}
-import org.apache.spark.sql.hive.DistributionUtil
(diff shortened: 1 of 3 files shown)


ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/200/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -590,16 +605,25 @@ object CarbonDataRDDFactory extends Logging {
...
+                    logger.error("Exception in compaction thread for table " + tableForCompaction
+                      .carbonTable.getDatabaseName + "." +
+                                 tableForCompaction.carbonTableIdentifier
+                                   .getTableName)
+                  // not handling the exception. only logging as this is not the table triggered
+                  // by user.
+                }
                 finally {
-                  // delete the compaction required file
+                  // delete the compaction required file in case of failure or success also.
```
gvramana: If compact file deletion fails, add failed table to skipped list.

ravikiran23: fixed

ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/248/

gvramana:
LGTM

PR #157, opened 2016-09-14
[CARBONDATA-240] Use SQLContext to query CarbonData directly without creating table

jackylk:
When using Spark 1.6, user can use SQLContext to query CarbonData files directly without creating table:
```
    // This requires Spark 1.6 or later version to run
    val sqlContext = new SQLContext(sc)
    sqlContext.sql(
      """
        | SELECT c1, c2, count(*)
        | FROM carbondata.`./examples/target/store/default/table1`
        | WHERE c3 > 100
        | GROUP BY c1, c2
      """.stripMargin).show
```

### DIFF
--- a/examples/src/main/scala/org/apache/carbondata/examples/DirectSQLExample.scala
+++ b/examples/src/main/scala/org/apache/carbondata/examples/DirectSQLExample.scala
@@ -17,36 +17,26 @@
 
 package org.apache.carbondata.examples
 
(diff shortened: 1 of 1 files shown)


jackylk:
@Vimal-Das can you merge this PR?

chenliang613:
@jackylk @Vimal-Das  i will handle this PR

PR #159, opened 2016-09-16
[CARBONDATA-242] Filter result was not compatible with hive result when a null filter member is present in 'NOT IN' filter model

sujith71955:
[**Problem**] 
Filter result was not compatible with hive result when a null filter member is present in not in
 filter model, as per hive no result shall be return if a NOT IN filter model has null object for comparison.
eg: select country from t3 where country not in (null,'china','france') group by country
[**Description**] 
When user provides Null member inside NOT IN filter condition the resultset is not compatible with hive result.


sujith71955:
PR build status
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/217/testReport/

gvramana:
change the PR title

gvramana:
LGTM

PR #160, opened 2016-09-16
[CARBONDATA-245] Actual Exception is getting lost in case of data dictionary file generation.

ravikiran23:
PROBLEM : 
during data load if any exception comes when the dictionary file is getting generated.
then in the finally block trying to write the meta file but as the stream is null, it will throw null pointer.
  
solution : 
in the finally added a check whether the stream is null or not. then only writing data to the stream.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/writer/CarbonDictionaryWriterImpl.java
+++ b/core/src/main/java/org/apache/carbondata/core/writer/CarbonDictionaryWriterImpl.java
@@ -198,11 +198,9 @@ public class CarbonDictionaryWriterImpl implements CarbonDictionaryWriter {
    * @throws IOException if an I/O error occurs
    */
   @Override public void close() throws IOException {
-    if (null != dictionaryThriftWriter) {
+    if (null != dictionaryThriftWriter && dictionaryThriftWriter.isOpen()) {
       // if stream is open then only need to write dictionary file.
-      if(dictionaryThriftWriter.isOpen()){
         writeDictionaryFile();
-      }
       // close the thrift writer for dictionary file
       closeThriftWriter();
     }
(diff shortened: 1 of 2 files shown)


ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/226/

manishgupta88:
LGTM

core/src/main/java/org/apache/carbondata/core/writer/ThriftWriter.java
```diff
@@ -80,6 +80,17 @@ public void open() throws IOException {
   }
 
   /**
+   * This will check whether stream and binary out is open or not.
+   * @return
+   */
+  public boolean isOpen() {
+    if (null != binaryOut && null != dataOutputStream) {
```
vinodkc: need to set  binaryOut  and dataOutputStream to null in close() method

core/src/main/java/org/apache/carbondata/core/writer/CarbonDictionaryWriterImpl.java
```diff
@@ -199,7 +199,10 @@ public CarbonDictionaryWriterImpl(String hdfsStorePath,
    */
   @Override public void close() throws IOException {
     if (null != dictionaryThriftWriter) {
-      writeDictionaryFile();
+      // if stream is open then only need to write dictionary file.
+      if(dictionaryThriftWriter.isOpen()){
```
vinodkc: Combine both if conditions

core/src/main/java/org/apache/carbondata/core/writer/ThriftWriter.java
ravikiran23: in close  we are using util to close the stream so i think no need to set null explicitly.

ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/253/

gvramana:
LGTM

PR #161, opened 2016-09-16
[CARBONDATA-246] compaction is wrong in case if last segment is not assigned to an executor.

ravikiran23:
PROBLEM:

if during compaction of 4 loads, for any executor if only first 3 loads task is assigned then the col cardinality calculation based on the last segment info will become wrong.

in this case the cardinality will go wrong for that executor.

Solution : 

Pass the segment properties info from the driver using the last segment.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
@@ -104,8 +104,9 @@ class CarbonMergerRDD[K, V](
 
         // get destination segment properties as sent from driver which is of last segment.
 
-        val segmentProperties = new SegmentProperties(carbonMergerMapping.columnSchemaList.asJava,
-          carbonMergerMapping.colCardinality)
+        val segmentProperties = new SegmentProperties(carbonMergerMapping.maxSegmentColumnSchemaList
+          .asJava,
+          carbonMergerMapping.maxSegmentColCardinality)
 
         // sorting the table block info List.
         var tableBlockInfoList = carbonSparkPartition.tableBlockInfos
@@ -162,7 +163,7 @@ class CarbonMergerRDD[K, V](
             segmentProperties,
             tempStoreLoc,
             carbonLoadModel,
-            carbonMergerMapping.colCardinality
+            carbonMergerMapping.maxSegmentColCardinality
           )
         mergeStatus = merger.mergerSlice()
 
@@ -254,7 +255,9 @@ class CarbonMergerRDD[K, V](
       )
 
       // keep on assigning till last one is reached.
-      blocksOfLastSegment = blocksOfOneSegment.asJava
+      if (null != blocksOfOneSegment && blocksOfOneSegment.size > 0) {
+        blocksOfLastSegment = blocksOfOneSegment.asJava
+      }
 
       // populate the task and its block mapping.
       blocksOfOneSegment.foreach(tableBlockInfo => {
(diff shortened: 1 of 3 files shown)


ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/227/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/Compactor.scala
```diff
@@ -69,7 +69,9 @@ object Compactor {
       schemaName,
       factTableName,
       validSegments,
-      carbonTable.getAbsoluteTableIdentifier.getCarbonTableIdentifier.getTableId
+      carbonTable.getAbsoluteTableIdentifier.getCarbonTableIdentifier.getTableId,
+      colCardinality = Array[Int](0),
```
manishgupta88: Instead of creating an empty list and reassigning it you can create a reference of Array[Int] type like
var colCardinality: Array[Int] = null

ravikiran23: fixed. using the null.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -102,6 +102,11 @@ class CarbonMergerRDD[K, V](
         var dataloadStatus = CarbonCommonConstants.STORE_LOADSTATUS_FAILURE
         val carbonSparkPartition = theSplit.asInstanceOf[CarbonSparkPartition]
 
+        // get destination segment properties as sent from driver which is of last segment.
+
+        val segmentProperties = new SegmentProperties(carbonMergerMapping.columnSchemaList.asJava,
```
gvramana: rename it to maxSegmentcolumnSchemaList and maxSegmentColCardinality.
Write comment where it is declared

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -259,6 +253,9 @@ class CarbonMergerRDD[K, V](
         )
       )
 
+      // keep on assigning till last one is reached.
+      blocksOfLastSegment = blocksOfOneSegment.asJava
```
gvramana: add null check and size check

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -280,6 +277,25 @@ class CarbonMergerRDD[K, V](
           taskInfoList.add(new TableTaskInfo(entry._1, entry._2).asInstanceOf[Distributable])
       )
     }
+
+    // prepare the details required to extract the segment properties using last segment.
+
```
gvramana: Add empty and null check.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
ravikiran23: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
ravikiran23: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
ravikiran23: fixed

ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/250/

gvramana:
LGTM

PR #162, opened 2016-09-17
[CARBONDATA-247] Higher MAXCOLUMNS value in load DML options is leading to out of memory error

dhatchayani:
Problem: Higher MAXCOLUMNS value in load DML options is leading to out of memory error

Analysis: When a higher value lets say Integer max value is configured for maxcolumns option in load DML and executor memory is less, then in that case UnivocityCsvParser throws an out of memory error when it tries to create an array of size of maxColumns option value.

Fix: Set the threshold value for maxColumns option value that our system can support and if maxColumns option value is greater than threshold value then assign the threshold value to maxColumns option value

Impact: Data loading


dhatchayani:
CI Passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/229/

nareshpr:
LGTM

ManoharVanam:
LGTM

manishgupta88:
CI Passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/240/

PR #163, opened 2016-09-17
[CARBONDATA-248]Problem:Column heading TOTAL QUERY COST was missing in driver statistics table and SCAN BLOCK TIME was always zero in query statistics table

sujith71955:
problem:There was no column header for total query cost in driver statistics table and scan block time was always zero in query statistics table.

Analysis:While creating driver statistics table header was not created and the scan block time was not initialized before recording in statistics recorder.

Fix:One more line is added to print the column header in driver statistics table and the scan block time is initialized before recording the statistics.

Impact:Logs of all query


kumarvishal09:
LGTM. If possible please add the snap shot of driver statistics :)

sraghunandan:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/261/

PR #167, opened 2016-09-17
[CARBONDATA-250] Throw exception and fail the data load if provided MAXCOLUMNS value is not proper

ManoharVanam:
[Problem] : If user executes data load query with improper maxcolumns value, data load is showing success instead of error in the console
[Solution] : Throw exception and fail the data load if provided MAXCOLUMNS value is not proper.


ManoharVanam:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/236/

gvramana:
LGTM

PR #168, opened 2016-09-17
[CARBONDATA-251] making the auto compaction as blocking call.

ravikiran23:
Problem : 
the behaviour of auto compaction with spark-submit and thrift server is different.
auto compaction will not work in spark-submit as the compaction is background process.

Solution : 
Make the auto compaction feature as blocking call.
disable the system level compaction lock feature by default.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -911,6 +911,7 @@ public final class CarbonCommonConstants {
   public static String majorCompactionRequiredFile = "compactionRequired_major";
 
   /**
+   * @Deprecated : This property has been deprecated.
    * Property for enabling system level compaction lock.1 compaction can run at once.
    */
   public static String ENABLE_CONCURRENT_COMPACTION =
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -920,7 +920,7 @@ public final class CarbonCommonConstants {
    * Default value of Property for enabling system level compaction lock.1 compaction can run
    * at once.
    */
-  public static String DEFAULT_ENABLE_CONCURRENT_COMPACTION = "false";
+  public static String DEFAULT_ENABLE_CONCURRENT_COMPACTION = "true";
 
   /**
    * Compaction system level lock folder.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -920,7 +920,7 @@
    * Default value of Property for enabling system level compaction lock.1 compaction can run
    * at once.
    */
-  public static String DEFAULT_ENABLE_CONCURRENT_COMPACTION = "false";
+  public static String DEFAULT_ENABLE_CONCURRENT_COMPACTION = "true";
```
gvramana: Add comment that this property "carbon.concurrent.compaction" is @Depricated

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -623,14 +623,7 @@ object CarbonDataRDDFactory extends Logging {
...
       }
-      if(compactionModel.isDDLTrigger) {
-        // making this an blocking call for DDL
-        compactionThread.run()
-      }
-      else {
-        // non blocking call in case of auto compaction.
-        compactionThread.start()
-      }
+    compactionThread.run()
```
gvramana: Add a comment, why run is called instead of start, future will make it concurrent.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -1030,28 +1021,36 @@ object CarbonDataRDDFactory extends Logging {
...
-          logInfo("********Database updated**********")
+          logger.audit("Data load is successful for " +
+                       s"${carbonLoadModel.getDatabaseName}.${carbonLoadModel.getTableName}")
+        try {
+          // compaction handling
+          handleSegmentMerging(tableCreationTime)
+        }
+        catch {
+          case e: Exception =>
+            throw new Exception("Dataload is success. Compaction is failed. Please check logs.")
```
gvramana: Add Auto compaction into error message.

core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
ravikiran23: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
ravikiran23: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
ravikiran23: fixed

ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/251/

gvramana:
LGTM

PR #169, opened 2016-09-17
[CARBONDATA-250] Filter result is not proper when Double data type values with 0.0 and -0.0 will be used.

sujith71955:
[**Problem**] 
Equals filter query where -0.0 is used in filter model was not giving proper result since the system was comparing double value 0.0 and -0.0 as different and resultset was getting changed. 
eg: select log(c4_double,1) from Test_Boundary1 where log(c4_double,1)= -0.0
[**Solution**] 
System  has to compare double values for its equality and also it will preserve the -0.0 and 0.0 equality as same  == operator ,also preserve NaN equality check as per java.lang.Double.equals()


sujith71955:
PR Builder Status http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/258/

gvramana:
LGTM

PR #170, opened 2016-09-17
[CARBONDATA-253]OOM issue if distribution is based on blocklet duing query execution

kumarvishal09:
Problem:In case of query execution when distribution is based on blocklet same blocks are getting loaded multiple times this is because hash code and equals method contract is not same, this is can cause OOM issue if distribution is based on blocklet
Solution: As same class will be used to identify unique blocks while distribution and while loading so creating a wrapper class and implementing hash code and equals method based on filepath, offset and length, this will remove duplicate blocks and only one block's metadata will be loaded in memory


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/262/

gvramana:
LGTM

PR #172, opened 2016-09-18
[CARBONDATA-255]Fix DOC

Zhangshunyu:
## Why raise this pr?

1. To fix doc that in DML 'show loads/DELETE LOAD ' , the keyword SEGMENT should be used instead of LOAD, because LOAD is not supported now.
2. To fix the usage of ALL_DICTIONARY_PATH

## How to test?
Never modify the code.

### DIFF
--- a/docs/DML-Operations-on-Carbon.md
+++ b/docs/DML-Operations-on-Carbon.md
@@ -18,7 +18,7 @@
 -->
 
 * [LOAD DATA](#LOAD DATA)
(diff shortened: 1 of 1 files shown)


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/265/

chenliang613:
LGTM

PR #173, opened 2016-09-18
[CARBONDATA-256]Changed StringBuffer to StringBuilder

vinodkc:
Even with local variables of Stringbuffer type, unnecessary synchronization overhead will be added in string operation.Replacing StringBuffer with StringBuilder in such cases can avoid synchronization overhead

           >> Not added new TCs . Not modified  any new functional change


jackylk:
LGTM

PR #180, opened 2016-09-20
[CARBONDATA-260] Equal or lesser value of MAXCOLUMNS option than column count in CSV header results into array index of bound exception

manishgupta88:
Problem: Equal or lesser value of MAXCOLUMNS option than column count in CSV header results into array index of bound exception

Analysis: If column count in CSV header is more or equal to MAXCOLUMNS option value then array index out of bound exception is thrown by the Univocity CSV parser. This is because while parsing the row, parser adds each row to an array and increments the index and after incrementing it performs one more operation using the incremented index value which leads to array index pf bound exception. Code snipped as attached below for CSV parser.

**public void valueParsed() {
	this.parsedValues[column++] = appender.getAndReset();
	this.appender = appenders[column];
}**

e.g. In the above code if column value is 7 then array index will be from 0-6 and when column value becomes 6 then in the second line ArrayIndexOutOfBoundException will be thrown as column value will become 7.

Fix: Whenever Column count in CSV header is equal or more than MAXCOLUMNS option value or default value, increment it by 1.

Impact: Data load flow

### DIFF
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/dataload/TestDataLoadWithColumnsMoreThanSchema.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/dataload/TestDataLoadWithColumnsMoreThanSchema.scala
@@ -99,6 +99,38 @@ class TestDataLoadWithColumnsMoreThanSchema extends QueryTest with BeforeAndAfte
     }
   }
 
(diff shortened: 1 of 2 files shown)


manishgupta88:
CI Passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/287/

gvramana:
LGTM

PR #181, opened 2016-09-20
[CARBONDATA-261] clean files is updating the stale segment status.

ravikiran23:
PROBLEM:
in clean files it will read the table status first and it will delete the physical locations and update the table metadata  and then take the lock for writing and after that it will write the old read list to the file.
but this is wrong as in the time between the read and write the table status would have been changed.
so before writing it should read the table status again and update the changes and then write.

SOLUTION:
before writing the status read the table status again and update the changes and then write.


gvramana:
LGTM

ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/289/

PR #182, opened 2016-09-20
[CARBONDATA-262]Fixed limit query issue

kumarvishal09:
Problem: In case of limit query if limit value is 100 so after consuming 100 records executor service is not getting shutdown, and it may cause memory issue 

Solution: Add executor service in query model need to shutdown the executor after query execution in carbonscan rdd


gvramana:
LGTM

kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/295/

gvramana:
LGTM

gvramana:
@kumarvishal09 Please raise PR on 0.1.0 branch also

PR #183, opened 2016-09-20
[CARBONDATA-263]Configurable blocklet distribution

kumarvishal09:
Problem: Currently blocklet distribution is not configurable , in some case it may impact query performance
Solution: Expose parameter to enable and disable blocklet distribution


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/298/

gvramana:
LGTM

PR #184, opened 2016-09-21
[CARBONDATA-264]Fixed limit query scan time statistics issue

kumarvishal09:
Problem: Scan time is not logging in statistics in case of limit query 
Soluntion: Moved statistics recording from iterator to queryExecutor finish method as in case of limit query after consuming records call is not coming to hasNext method

### DIFF
--- a/core/src/main/java/org/apache/carbondata/scan/executor/impl/QueryExecutorProperties.java
+++ b/core/src/main/java/org/apache/carbondata/scan/executor/impl/QueryExecutorProperties.java
@@ -18,9 +18,7 @@
  */
 package org.apache.carbondata.scan.executor.impl;
 
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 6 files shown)


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/300/testReport/

core/src/main/java/org/apache/carbondata/scan/executor/impl/QueryExecutorProperties.java
```diff
@@ -91,5 +89,10 @@
    * list of blocks in which query will be executed
    */
   protected List<AbstractIndex> dataBlocks;
+  /**
+   * start time scanning
```
jackylk: it is the start time of scanning?

ravipesala:

PR #185, opened 2016-09-21
[CARBONDATA-266]Add a new feature that support delete all carbon tables under one database.

Zhangshunyu:
## Why rasie this pr?
Add a new feature that support delete all carbon tables under one database.
**Only delete all carbon tables, do not has effect on other tables.**

## How to test?
Pass all the testcases including the new test case.


Zhangshunyu:
CI has passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/303/testReport/

Zhangshunyu:
For user's convenience, user can delete all the carbon tables under one database one time instead of one by one, and it not delete the other tables.

jackylk:
I am of the opinion that we should add new syntax unless it is necessary, and find the standard syntax if there are any. If there are no such syntax in standard SQL, it probably means that it is not a common requirement to add such syntax.

manishgupta88:
@Zhangshunyu ...we already have a sql standard command "drop database database_name cascade", then we can use the same instead of giving one more command...we can enable cascade option in the code....
@jackylk @gvramana ...what is your opinion?

PR #189, opened 2016-09-22
[CARBONDATA-267] Set block_size for table on table level

Zhangshunyu:
## Why raised this pr?
To configure block file size for each table on table level.
## How to solve?
Add a new parameter in TableSchema, when create table, setting it in table properties and write this info into thrift file.
## How to test?
Pass all the test cases and the new test case.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
+++ b/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
@@ -21,6 +21,7 @@ package org.apache.carbondata.core.carbon.metadata.schema.table;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.carbondata.core.carbon.metadata.schema.SchemaEvolution;
 import org.apache.carbondata.core.carbon.metadata.schema.table.column.ColumnSchema;
@@ -57,9 +58,9 @@ public class TableSchema implements Serializable {
   private SchemaEvolution schemaEvalution;
 
   /**
-   * Used to record the block size of this table.
+   * contains all key value pairs for table properties set by user in craete DDL
    */
-  private int blocksize;
+  private Map<String, String> tableProperties;
 
   public TableSchema() {
     this.listOfColumns = new ArrayList<ColumnSchema>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);
@@ -105,14 +106,6 @@ public class TableSchema implements Serializable {
    */
   public void setSchemaEvalution(SchemaEvolution schemaEvalution) {
     this.schemaEvalution = schemaEvalution;
-  }
-
-  public int getBlocksize() {
-    return blocksize;
-  }
-
-  public void setBlocksize(int blocksize) {
-    this.blocksize = blocksize;
   }
 
   /**
@@ -196,4 +189,17 @@ public class TableSchema implements Serializable {
     return true;
   }
 
+  /**
+   * @return
+   */
+  public Map<String, String> getTableProperties() {
+    return tableProperties;
+  }
+
+  /**
+   * @param tableProperties
+   */
+  public void setTableProperties(Map<String, String> tableProperties) {
+    this.tableProperties = tableProperties;
+  }
 }
--- a/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
+++ b/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
@@ -59,7 +59,7 @@ public class TableSchema implements Serializable {
   /**
    * Used to record the block size of this table.
    */
-  private int blockszie;
+  private int blocksize;
 
   public TableSchema() {
     this.listOfColumns = new ArrayList<ColumnSchema>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);
@@ -107,12 +107,12 @@ public class TableSchema implements Serializable {
     this.schemaEvalution = schemaEvalution;
   }
 
-  public int getBlockszie() {
-    return blockszie;
+  public int getBlocksize() {
+    return blocksize;
   }
 
-  public void setBlockszie(int blockszie) {
-    this.blockszie = blockszie;
+  public void setBlocksize(int blocksize) {
+    this.blocksize = blocksize;
   }
 
   /**
--- a/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
+++ b/core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
@@ -56,6 +56,11 @@ public class TableSchema implements Serializable {
    */
   private SchemaEvolution schemaEvalution;
 
+  /**
+   * Used to record the block size of this table.
+   */
+  private int blockszie;
+
   public TableSchema() {
     this.listOfColumns = new ArrayList<ColumnSchema>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);
   }
@@ -100,6 +105,14 @@ public class TableSchema implements Serializable {
    */
   public void setSchemaEvalution(SchemaEvolution schemaEvalution) {
     this.schemaEvalution = schemaEvalution;
+  }
+
+  public int getBlockszie() {
+    return blockszie;
+  }
+
+  public void setBlockszie(int blockszie) {
+    this.blockszie = blockszie;
   }
 
   /**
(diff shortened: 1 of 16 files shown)


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/311/testReport/

format/src/main/thrift/schema.thrift
```diff
@@ -124,6 +124,7 @@ struct TableSchema{
 	1: required string table_id;  // ID used to
 	2: required list<ColumnSchema> table_columns; // Columns in the table
 	3: required SchemaEvolution schema_evolution; // History of schema evolution of this table
+	4: optional i32 block_size
```
jackylk: Why save it to the file? It should be stored as a table level metadata.

Zhangshunyu: @jackylk If dont save it into file, when we restart thirift-server, the message in memory object would be lost.

Zhangshunyu: @jackylk Futher more, all the meta data of table would also be stored into file by thrift, if not, when restart all info would be lost

jackylk: I think it is better to save it in the table metadata, through metastore for example. 
By concept, this should not be tied with the file.
@ravipesala @gvramana any suggestion?

Zhangshunyu: @jackylk Each table has his own metadata, this file is just that one......

manishgupta88:
@Zhangshunyu ...In which scenario this PR will be helpful.....I mean is there any use case for this PR?

Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/343/testReport/

processing/src/main/java/org/apache/carbondata/processing/store/writer/AbstractFactDataWriter.java
```diff
@@ -196,11 +197,22 @@ public AbstractFactDataWriter(String storeLocation, int measureCount, int mdKeyL
...
+    // If blocksize is defined by create table ddl, then use this value otherwise
+    // use system level property
+    if (blocksize >= CarbonCommonConstants.MAX_FILE_SIZE_DEFAULT_VAL_MIN_VAL &&
+        blocksize <= CarbonCommonConstants.MAX_FILE_SIZE_DEFAULT_VAL_MAX_VAL) {
+      LOGGER.audit("The max file size is set in table level. ");
+      this.fileSizeInBytes = blocksize * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR
+          * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR * 1L;
+    } else {
+      LOGGER.audit("The max file size in create table ddl is not set or is invalid, " +
+          "so here use " + CarbonCommonConstants.MAX_FILE_SIZE + "to set. ");
```
manishgupta88: @Zhangshunyu ...I am thinking instead of validating block size here specified by the user, we can validate at the time of creating the table and in error message we can display the range of allowed block size so that user is not confused.
If user does not specify the block size the consider the default value and update in the schema

format/src/main/thrift/schema.thrift
manishgupta88: @Zhangshunyu ...I agree with @jackylk ...just like we are storing the table path in metastore, same way block size can also be stored as it is metadata for the table. schema file we write should only describe the structure the table and not the metadata for table

sujith71955: i think we should initialize the block size with default value, Will be better.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -546,6 +548,18 @@ class CarbonSqlParser()
...
   }
 
+  protected def getTableBlockSize(tableProperties: Map[String, String]): Integer = {
+    var tableBlockSize: Integer = 0
+    if (tableProperties.get(CarbonCommonConstants.TABLE_BLOCKSIZE).isDefined) {
+      val blockSizeStr: String = tableProperties.get(CarbonCommonConstants.TABLE_BLOCKSIZE).get
+      try {
+        tableBlockSize = Integer.parseInt(blockSizeStr)
+      } catch {
+        case e: NumberFormatException => tableBlockSize = 0
```
sujith71955: better assign default value. so that system will be in a consistent state always

core/src/main/java/org/apache/carbondata/core/carbon/metadata/schema/table/TableSchema.java
```diff
@@ -102,6 +107,14 @@ public void setSchemaEvalution(SchemaEvolution schemaEvalution) {
     this.schemaEvalution = schemaEvalution;
   }
 
+  public int getBlockszie() {
+    return blockszie;
+  }
+
+  public void setBlockszie(int blockszie) {
+    this.blockszie = blockszie;
```
sujith71955: provide proper variable name 'blocksize'

Zhangshunyu:
@manishgupta88 @sujith71955 Fixed all review comments, pls check again. CI has passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/356/

format/src/main/thrift/schema.thrift
Zhangshunyu: @manishgupta88 As this point, i hold my own opinion.

Zhangshunyu: @manishgupta88 
1. Now we put all columnProperties in format.ColumnSchema, and all ColumnSchema would be stored into format.TableSchema, why not table property(block.size) it self can not stored into format.TableSchema?  I think its ok.
2. We must using file to store this info, if not, when restart the message would lost. Further more it is not reasonable to use one new file who only store this info.

manishgupta88: @Zhangshunyu ...if you register the property in hive metastore, the proeprty will never be lost even on restart of thrift server as we do for tablepath...you can create table flow

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -546,6 +548,30 @@ class CarbonSqlParser()
...
+        tableBlockSize = Integer.parseInt(blockSizeStr)
+      } catch {
+        case e: NumberFormatException =>
+          throw new MalformedCarbonCommandException("Invalid table_blocksize value found: " +
+            s"$blockSizeStr, only int value from 1 to 2048 is supported.")
+      }
+      if (tableBlockSize < CarbonCommonConstants.BLOCK_SIZE_MIN_VAL ||
+        tableBlockSize > CarbonCommonConstants.BLOCK_SIZE_MAX_VAL) {
+        throw new MalformedCarbonCommandException("Invalid table_blocksize value found: " +
+          s"$blockSizeStr, only int value from 1 to 2048 is supported.")
```
sujith71955: Please mention the Unit also in comment like GB or MB
"Invalid table_blocksize value found: " +
 +          s"$blockSizeStr, only int value from 1 MB to 2048 MB is supported"

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -419,6 +420,7 @@ class TableNewProcessor(cm: tableModel, sqlContext: SQLContext) {
     schemaEvol
       .setSchemaEvolutionEntryList(new util.ArrayList[SchemaEvolutionEntry]())
     tableSchema.setTableId(UUID.randomUUID().toString)
+    tableSchema.setBlocksize(Integer.parseInt(cm.tableBlockSize.getOrElse(0).toString))
```
sujith71955: What if user will provide value as 1024M or 1MB, you will get exception and the same has not been handled.
I think we need to handle

format/src/main/thrift/schema.thrift
gvramana: @Zhangshunyu Actually need not change Schema for such extra properties which are being added. We can directly store TABLE_PROPERTIES as a string key value pairs in thrift. So that all such properties configuration parameters can be effectively handled without modifying thrift in future. @jackylk please comment.

Zhangshunyu:
@gvramana yes，i agree with this idea that we can add a TABLE_PROPERTY in thrift.

manishgupta88:
@gvramana @jackylk  ...Please check and merge. Handle review comments.
CI Passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/370/

ravipesala:
LGTM

PR #193, opened 2016-09-22
Updated integration-test/pom.xml with correct version

ravipesala:
Updated integration-test/pom.xml with correct version


chenliang613:
LGTM

PR #196, opened 2016-09-23
[CARBONDATA-271]Fixed non filter data mismatch issue.

kumarvishal09:
Problem: While generating the default end key we are taking LONG.MAX key and using segment key generator we are generating the end key if cardinality is less than it will give some value with in its cardinality and btree searching will fail

Solution: From segment property get the dimension cardinality as this is the max value for segment


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/327/

kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/329/

gvramana:
So impact of the problem is:
Some of the block-lets can get skipped as filter key we are making is not proper.

gvramana:
LGTM

PR #200, opened 2016-09-27
[CARBONDATA-276]add trim property for cols

lion-x:
# Why raise this PR?

Add trim property for cols.
trim a col or not can be set during create table time, like,

CREATE TABLE IF NOT EXISTS t3 (ID Int, date Timestamp, country String, name String, phonetype String, serialname String, salary Int)
STORED BY 'carbondata'
TBLPROPERTIES("trim"="country")

In order to delete some meaningless white space in some fields, we do this PR. It will also improve the index finding performance with shorter string.
# How to test?

Pass all the test cases.

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/graphgenerator/GraphGenerator.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/graphgenerator/GraphGenerator.java
@@ -583,6 +583,7 @@ public class GraphGenerator {
     seqMeta.setHeirKeySize(graphConfiguration.getHeirAndKeySizeString());
     seqMeta.setColumnSchemaDetails(graphConfiguration.getColumnSchemaDetails().toString());
     seqMeta.setTableOption(graphConfiguration.getTableOptionWrapper().toString());
+    seqMeta.setIsUseTrim(graphConfiguration.getIsUseTrim());
     String[] aggType = graphConfiguration.getAggType();
     StringBuilder builder = new StringBuilder();
     for (int i = 0; i < aggType.length; i++) {
@@ -800,6 +801,7 @@ public class GraphGenerator {
     List<CarbonDimension> dimensions = carbonDataLoadSchema.getCarbonTable()
         .getDimensionByTableName(carbonDataLoadSchema.getCarbonTable().getFactTableName());
     prepareIsUseInvertedIndex(dimensions, graphConfiguration);
+    prepareIsUseTrim(dimensions, graphConfiguration);
     graphConfiguration
         .setDimensions(CarbonSchemaParser.getTableDimensions(dimensions, carbonDataLoadSchema));
     graphConfiguration
@@ -997,5 +999,25 @@ public class GraphGenerator {
     }
     graphConfig.setIsUseInvertedIndex(
         isUseInvertedIndexList.toArray(new Boolean[isUseInvertedIndexList.size()]));
+  }
+
+  /**
+   * Preparing the boolean [] to map whether the dimension use trim or not.
+   *
+   * @param dims
+   * @param graphConfig
+   */
+  private void prepareIsUseTrim(List<CarbonDimension> dims,
+                                GraphConfigurationInfo graphConfig) {
+    List<Boolean> isUseTrimList = new ArrayList<Boolean>();
+    for (CarbonDimension dimension : dims) {
+      if (dimension.isUseTrim()) {
+        isUseTrimList.add(true);
+      } else {
+        isUseTrimList.add(false);
+      }
+    }
+    graphConfig.setIsUseTrim(
+            isUseTrimList.toArray(new Boolean[isUseTrimList.size()]));
   }
 }
--- a/processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenMeta.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenMeta.java
@@ -386,6 +386,8 @@ public class CarbonCSVBasedSeqGenMeta extends BaseStepMeta implements StepMetaIn
    */
   private String columnsDataTypeString;
 
+  private String isUseTrim = "";
+
   public CarbonCSVBasedSeqGenMeta() {
     super();
   }
@@ -655,6 +657,7 @@ public class CarbonCSVBasedSeqGenMeta extends BaseStepMeta implements StepMetaIn
     columnsDataTypeString="";
     tableOption = "";
     dateFormat = "";
+    isUseTrim = "";
   }
 
   // helper method to allocate the arrays
@@ -721,6 +724,7 @@ public class CarbonCSVBasedSeqGenMeta extends BaseStepMeta implements StepMetaIn
         .append(XMLHandler.addTagValue("columnSchemaDetails", columnSchemaDetails));
     retval.append("    ")
         .append(XMLHandler.addTagValue("tableOption", tableOption));
+    retval.append("    ").append(XMLHandler.addTagValue("isUseTrim", isUseTrim));
     return retval.toString();
   }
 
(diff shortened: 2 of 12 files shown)


lion-x:
CI passed!
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/347/

hadoop/src/test/java/org/apache/carbondata/hadoop/test/util/StoreCreator.java
```diff
@@ -465,6 +466,7 @@ private static void generateGraph(IDataProcessStatus schmaModel, SchemaInfo info
     model.setEscapeCharacter(schmaModel.getEscapeCharacter());
     model.setQuoteCharacter(schmaModel.getQuoteCharacter());
     model.setCommentCharacter(schmaModel.getCommentCharacter());
+    model.setTrim(schmaModel.getTrim());
```
sujith71955: why we need to set this in schemamodel?

sujith71955: Not clear about the usecase, can you make it more clear by providing more details

QiangCai:
I think the dictionary generation is the same with the CSVInput, not ignore any white space.
It is unneccessary to trim data.

lion-x:
Trimming data feature is used to support some users  requirements. There is no  causality between the dictionary generation is the same with the CSVInput and triming data or not.

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/UnivocityCsvParser.java
```diff
@@ -102,8 +102,8 @@ public void initialize() throws IOException {
     parserSettings.setMaxColumns(
         getMaxColumnsForParsing(csvParserVo.getNumberOfColumns(), csvParserVo.getMaxColumns()));
     parserSettings.setNullValue("");
-    parserSettings.setIgnoreLeadingWhitespaces(false);
-    parserSettings.setIgnoreTrailingWhitespaces(false);
+    parserSettings.setIgnoreLeadingWhitespaces(csvParserVo.getTrim());
```
sujith71955: Same you need to handle in CarbonFilters.scala also, since while processing filter expressions the spaces are not getting trimmed, so here also you need to take care based on your feature

eason-lyx: hi sujith
i am thinking if user already trim the data with the option setting,then when user query with some space filter,it would no getting result.

lion-x: hi sujith,
I agree with Eason, when user query with some space filter, it should be considered as forbidden action.

sujith71955: Also one more point it will be better to set this property in column level while creating the table itself as its column properties , this will avoid user to provide this option every time while data loading

sujith71955: pros of this approach will be suppose in one load user loaded with dirty data and suddenly he realizes no i need to trim then in the next load he will enable the option and load the data, this will increase the dictionary space also, also in query dictionary lookup overhead will increase.

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/dataload/TestDataLoadWithTrimOption.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import org.apache.carbondata.core.constants.CarbonCommonConstants
+import org.apache.carbondata.core.util.CarbonProperties
+import org.apache.spark.sql.common.util.CarbonHiveContext._
+import org.apache.spark.sql.common.util.QueryTest
+import org.scalatest.BeforeAndAfterAll
+import org.apache.spark.sql.Row
+
+/**
+  * Created by x00381807 on 2016/9/26.
```
lion-x: Oh, my fault

lion-x:
@sujith71955 @jackylk pls review again. The code has been updated. The trim property for cols is set in Create.

jackylk:
Can you provide more information about this PR, like what kind of query and why is it a bug?

lion-x:
@jackylk there is no bug, I update my description.

processing/src/main/java/org/apache/carbondata/processing/graphgenerator/GraphGenerator.java
```diff
@@ -998,4 +1000,24 @@ private void prepareIsUseInvertedIndex(List<CarbonDimension> dims,
...
+   * Preparing the boolean [] to map whether the dimension use trim or not.
+   *
+   * @param dims
+   * @param graphConfig
+   */
+  private void prepareIsUseTrim(List<CarbonDimension> dims,
+                                GraphConfigurationInfo graphConfig) {
+    List<Boolean> isUseTrimList = new ArrayList<Boolean>();
+    for (CarbonDimension dimension : dims) {
+      if (dimension.isUseTrim()) {
```
sujith71955: Can we add this trim option as property of the column, i mean inside a column property rather than setting directly as CarbonDimension property, Already in CarbonColumn there is a property map which holds column related properties, i think we can use that, please check the feasibility.

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenMeta.java
```diff
@@ -1694,5 +1699,19 @@ public void setTableOption(String tableOption) {
...
   }
+
+  public String getIsUseTrim() {
+    return isUseTrim;
+  }
+
+  public void setIsUseTrim(Boolean[] isUseTrim) {
+    for (Boolean flag: isUseTrim) {
+      if (flag) {
+        this.isUseTrim += "T";
```
sujith71955: Use  TRUE/FALSE for better readability

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -472,6 +475,7 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
               break;
           }
         }
+<<<<<<< HEAD
```
sujith71955: is this file is having any conflict?

bill1208:
I suggestion the trim should not be the properties of the table. 
because in actually scenario the table designer is in front, then data of the this table is generated. 
so the designer may don not consider which column need trim that may only can know after data generated. 
So I suggestion this trim operation should be consider when data load. 
during the data load user can set all column trim or not and also set trim for special column both for datalod command and insert command

sujith71955:
@Bill, i think while creating table the user shall know the entire schema which he is trying to generate and what kind of data/datatype is required for his table based on the expected data, if we provide the support in data load there are few drawbacks which we discussed in the prior discussion thread, please check.

sujith71955:
i will repeat the cons again if we set this property while data loading.
cons of this approach will be suppose in one load user loaded with dirty data and suddenly he realizes no i need to trim then in the next load he will enable the option and load the data, this will increase the dictionary space also, also in query dictionary lookup overhead will increase.

sujith71955:
just a thinking you can add your points  :)

bill1208:
@sujith  I got your point. 
but I think trim data should belong to the step when process the data, which not belong to the table properties. 
the scenario what's your des is exist even if specify it when create table. 
For example when create the table they donot think it need trim the column1, but when actually data come or some business changed it need trim it. how to process this scenario? recreate the table or alter the table properties?   :)

sujith71955:
@Bill , I agree to your valid points but i am thinking  data load is a frequent operation compare to Create, and more over once you create table with the column properties specified, it will remain the same throughout the table life cycle, if we give in load command there is always a scope of inconsistency by keeping few loads with trimmed data and few without.  i think user shall able to judge/decide if data should contain white spaces or it shall be trimmed for particular column if he knows his schema well .
And more over if the initial load the column data is having white space and as you told above he thinks to trim it, for that again user will perform data loading, but in this case his column count/distinct count and all will be different compare to the actual  and also he has to see more query result with the data he is not interested, now if he delete the particular load also again he needs to reload the particular load data which will be again a cumbersome process. we shall try to make user to make his steps right and keep it simple.
You can put a view on my above points

lion-x:
@sujith71955 @bill1208 
actually, in my opinion, trimming data should belong to data preprocessing.  If we must put trim property in carbon, I think it is better to put trim option in loading data part. If user put unprepocessing data into table, it is user's fault.

ravipesala:

chenliang613:
@lion-x  please rebase it.

PR #202, opened 2016-09-27
[CARBONDATA-278] IS NULL and IS NOT NULL shall be push down to carbon

sujith71955:
**Current design** : As per the current design IS NULL and IS NOT NULL filter is been handled by spark layer
**Drawback** :  Block/Blocklet level pruning will not be possible if the filter is been processed by spark layer
**Proposed Improvement**:  IS NULL and IS NOT NULL shall be push down to carbon layer since carbon layer can process these filters faster using block/block-let pruning , also while processing filters in executers  carbon is applying binary search for applying filter values.

**UT/FT Testcases** : Existing testcases can verify the impact of this PR change since its not new usecase/scenario.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/scan/expression/conditional/BinaryConditionalExpression.java
+++ b/core/src/main/java/org/apache/carbondata/scan/expression/conditional/BinaryConditionalExpression.java
@@ -29,7 +29,7 @@ public abstract class BinaryConditionalExpression extends BinaryLogicalExpressio
    *
    */
   private static final long serialVersionUID = 1L;
-
+  public boolean isNull;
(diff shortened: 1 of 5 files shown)


sujith71955:
Please find the PR builder status
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/355/

jackylk:
@sujith71955 can you add testcase also?

sujith71955:
@Jacky Existing testcases can verify the impact of this PR change since its not new usecase/scenario.
There are test cases for IS NULL and IS NOT NULL in the following test classes FilterProcessorTestcase.scala , AllDataTypesTestCaseAggregate.scala. please let me know if i need to add more. Thanks for review.

PR #203, opened 2016-09-27
[CARBONDATA-279]Load RDD to carbondata file directly without writing CSV file

QiangCai:
1. Support dictionary generation read data from DataFrame
Add a dataframe parameter for GlobalDictionaryUtil.generateGlobalDictionary function.

2. Support Data Loading read data from DataFrame
Modify CsvInput step to read data from RDD.

3. Improve saveAsCarbonFile function 
Avoid to generate CSV file

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
@@ -17,8 +17,6 @@
 
 package org.apache.carbondata.spark
 
-import org.apache.spark.sql.{getDB, SQLContext}
-
 /**
  * Contains all options for Spark data source
  */
@@ -38,7 +36,7 @@ class CarbonOption(options: Map[String, String]) {
       "org.apache.carbondata.spark.partition.api.impl.SampleDataPartitionerImpl")
   }
 
-  def tempCSV: String = options.getOrElse("tempCSV", "false")
+  def tempCSV: String = options.getOrElse("tempCSV", "true")
 
   def compress: String = options.getOrElse("compress", "false")
 
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
@@ -22,10 +22,10 @@ import org.apache.spark.sql.{getDB, SQLContext}
 /**
  * Contains all options for Spark data source
  */
-class CarbonOption(options: Map[String, String], cc: SQLContext) {
+class CarbonOption(options: Map[String, String]) {
   def tableIdentifier: String = options.getOrElse("tableName", s"$dbName.$tableName")
 
-  def dbName: String = getDB.getDatabaseName(options.get("dbName"), cc)
+  def dbName: String = options.getOrElse("dbName", "default")
 
   def tableName: String = options.getOrElse("tableName", "default_table")
 
@@ -37,6 +37,8 @@ class CarbonOption(options: Map[String, String], cc: SQLContext) {
     options.getOrElse("partitionClass",
       "org.apache.carbondata.spark.partition.api.impl.SampleDataPartitionerImpl")
   }
+
+  def tempCSV: String = options.getOrElse("tempCSV", "false")
 
   def compress: String = options.getOrElse("compress", "false")
 
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
@@ -17,13 +17,15 @@
 
 package org.apache.carbondata.spark
 
+import org.apache.spark.sql.{getDB, SQLContext}
+
 /**
  * Contains all options for Spark data source
  */
-class CarbonOption(options: Map[String, String]) {
+class CarbonOption(options: Map[String, String], cc: SQLContext) {
   def tableIdentifier: String = options.getOrElse("tableName", s"$dbName.$tableName")
 
-  def dbName: String = options.getOrElse("dbName", "default")
+  def dbName: String = getDB.getDatabaseName(options.get("dbName"), cc)
 
   def tableName: String = options.getOrElse("tableName", "default_table")
 
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
@@ -19,6 +19,7 @@
 package org.apache.carbondata.spark.rdd
 
 import java.lang.Long
+import java.text.SimpleDateFormat
 import java.util
 import java.util.UUID
 
@@ -26,8 +27,10 @@ import scala.collection.JavaConverters._
 import scala.util.Random
 
 import org.apache.spark.{Logging, Partition, SerializableWritable, SparkContext, SparkEnv, TaskContext}
+import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.execution.command.Partitioner
+import org.apache.spark.sql.Row
 
 import org.apache.carbondata.common.logging.LogServiceFactory
 import org.apache.carbondata.common.logging.impl.StandardLogService
@@ -35,6 +38,7 @@ import org.apache.carbondata.core.constants.CarbonCommonConstants
 import org.apache.carbondata.core.load.{BlockDetails, LoadMetadataDetails}
 import org.apache.carbondata.core.util.{CarbonProperties, CarbonTimeStatisticsFactory}
 import org.apache.carbondata.processing.constants.DataProcessorConstants
+import org.apache.carbondata.processing.csvreaderstep.RddInputUtils
 import org.apache.carbondata.processing.etl.DataLoadingException
 import org.apache.carbondata.processing.graphgenerator.GraphGenerator
 import org.apache.carbondata.spark.DataLoadResult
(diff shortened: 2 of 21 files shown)


jackylk:
Please describe this PR @QiangCai

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -462,3 +478,133 @@ class CarbonDataLoadRDD[K, V](
...
+        carbonLoadModel.setSegmentId(String.valueOf(loadCount))
+        carbonLoadModel.setTaskNo(String.valueOf(theSplit.index))
+
+        storeLocation = CarbonDataLoadRDD.initialize(carbonLoadModel, theSplit.index)
+        loadMetadataDetails.setLoadStatus(CarbonCommonConstants.STORE_LOADSTATUS_SUCCESS)
+        val rddIteratorKey = UUID.randomUUID().toString
+        try{
+          RddInputUtils.put(rddIteratorKey,
+            new RddIterator(firstParent[Row].iterator(theSplit, context), carbonLoadModel))
+          carbonLoadModel.setRddIteratorKey(rddIteratorKey)
```
jackylk: Why can't you just set the iterator object itself in the carbonLoadModel instead of putting it in a global map?

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/CsvInput.java
```diff
@@ -85,6 +87,8 @@
 
   private ExecutorService exec;
 
+  private String rddIteratorKey = null;
```
jackylk: can we make another input class instead of CsvInput? Like SparkRowInput?

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/CsvInput.java
```diff
@@ -343,41 +349,57 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
...
     setOutputDone();
     return false;
   }
 
+  private void scanRddIterator() throws RuntimeException {
+    Iterator<String[]> iterator = RddInputUtils.getAndRemove(rddIteratorKey);
+    if (iterator != null) {
+      try{
+        while(iterator.hasNext()){
+          putRow(data.outputRowMeta, iterator.next());
```
jackylk: One suggestion:
Instead of calling putRow here, perhaps we can execute more steps here until reaching merge step which need to wait. 
I suggest to raise another PR to implement this improvement.

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -742,7 +742,8 @@ object GlobalDictionaryUtil extends Logging {
    */
   def generateGlobalDictionary(sqlContext: SQLContext,
                                carbonLoadModel: CarbonLoadModel,
-                               hdfsLocation: String): Unit = {
+                               hdfsLocation: String,
+                               dataFrame: Option[DataFrame] = None): Unit = {
```
jackylk: I think we can call dataFrame.rdd.cache inside this function, because there are 2 scan, dataframe can be re-used after the first scan.

jackylk:
@QiangCai thanks for working on this, with this feature, user can use dataframe as input to load carbondata files and without saving temporary CSV files. 
The only problem here is that it is still requiring 2 scan of the input dataframe, we probably should have another PR to explore the 1 scan approach.

ravipesala:
@QiangCai @jackylk  I think this approach requires to execute dataframe twice. It is not recommendable to execute twice  as we don' t know the cost of execution of dataframe twice and it may end up with taking long time. And also caching dataframe is also not a good idea as we don't know the size of out put.  I think we can better keep the csv (compressed csv)staging solution until we find the single pass data loading solution.

jackylk:
@ravipesala can we keep it as an alternative to the current loading process? You are right about the data frame execution cost, however, I can see some benefit:
1. no temporary CSV file is required, so that less disk space is required, if the user does not have enough disk space.
2. say, this data frame is from a hive table, ORC scan is much more efficient than compressed CSV scan due to better compression of ORC, and it also doesn't require the CSV data write cost, so data frame execution cost should be less in this case, which I think is typical use case nowadays
3. as we go forward, I think this is the first step toward non-kettle solution, we need to do this anyway in the future.

What is your idea? Do you think we should find a way to keep both approaches?

ravipesala:
@jackylk Both solutions have pros and cons, both are alternate solutions only.
In case of loading data from hive or orc or parquet etc.. to carbondata, we can use the respective InputFormats to load data to carbondata instead of using dataframe if there is no transformation involved.
But if the transformation is involved before loading to carbondata then executing dataframe twice may take very long time depends upon the transformation cost.
Each solution can solve some set of problems but both are not perfect solutions. I think the better solution is to make the data loading single pass.
@QiangCai  Can you move this discussion to dev mailing list with current opinions.

QiangCai:
@jackylk @ravipesala 
If the output set is small and select query take a long time, maybe compressed Csv solution is better
If the output set is big, maybe load dataframe solution is better.

Now We can keep two solutions. And carbon user can choose a better solution for the special dataframe.

jarray888:
@QiangCai carbon can not load two huge data one time currently , for example load 10 billion records one time .
So we seperate the input CSV to multi part , and load multi times to work around .
If customer create a 10 billion text table ,and insert into  carbon table , will face same problem ?

JihongMA:
@QiangCai @jackylk @ravipesala  This is good feature and long waiting one. I also have concern about executing DF twice, before we figuring out the detail of one single pass data loading,  we actually could first introduce a version without global dictionary generation, fall back to local dictionary encoding or alternatively the table might already  have global dictionary in place (under metadata) or allowing user to supply global dictionary if it is generated somewhere else, so that this piece of work can still align with the ultimate single pass solution with global dictionary generation on the fly.  it looks to me eliminating the intermediate csv file is urgent and has higher priority.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -462,3 +478,133 @@ class CarbonDataLoadRDD[K, V](
   }
 }
 
+class CarbonRDDDataLoadRDD[K, V](
```
chenliang613: Please add needful annotate for new class CarbonRDDDataLoadRDD. (it is difficult to know what is the different of CarbonRDDDataLoadRDD, CarbonDataLoadRDD)
Same comments for other place of this PR :  please provide needful annotate for new classes and new functions.

chenliang613:
@QiangCai  please put the discussion to mailing list, and let more users vote the different solutions.

I suggest : currently we can have two simple ways(dataframe and CSV) for transferring data to carbon data before we finish one scan data loading solution.  (Don't suggest implementing dataframe solution very complexly)

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
QiangCai: In the CSVInput Step, we cann't call carbonLoadModel too.

integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
```diff
@@ -17,13 +17,15 @@
...
+
 /**
  * Contains all options for Spark data source
  */
-class CarbonOption(options: Map[String, String]) {
+class CarbonOption(options: Map[String, String], cc: SQLContext) {
   def tableIdentifier: String = options.getOrElse("tableName", s"$dbName.$tableName")
 
-  def dbName: String = options.getOrElse("dbName", "default")
+  def dbName: String = getDB.getDatabaseName(options.get("dbName"), cc)
```
QiangCai: fixed

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/CsvInput.java
QiangCai: Now we still need keep kettle steps.

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/CsvInput.java
```diff
@@ -343,41 +349,57 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
     } catch (NumberFormatException exc) {
       numberOfNodes = NUM_CORES_DEFAULT_VAL;
     }
+    if ( rddIteratorKey == null ) {
```
QiangCai: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -72,6 +76,89 @@ class CarbonNodePartition(rddId: Int, val idx: Int, host: String,
   override def hashCode(): Int = 41 * (41 + rddId) + idx
 }
 
+object CarbonDataLoadRDD extends Logging{
```
QiangCai: Rename to CarbonDataLoadRDDFuncs

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
QiangCai: I will add some comment to describe the difference points.

QiangCai:
@jackylk please review.

QiangCai:
CI PASS
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/377/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -72,8 +76,91 @@ class CarbonNodePartition(rddId: Int, val idx: Int, host: String,
   override def hashCode(): Int = 41 * (41 + rddId) + idx
 }
 
+object CarbonDataLoadRDDFuncs extends Logging{
```
jackylk: Can we have a more meaningful name for this?
I suggest `SparkPartitionLoader` (it loads one spark partition into carbon using kettle), and make it a class instead of object, so that it can keep the storeLocation internally instead of returning it in initialize function. One is using the returned storeLocation variable other than run function.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
```diff
@@ -462,3 +478,155 @@ class CarbonDataLoadRDD[K, V](
...
+ * @param kettleHomePath
+ * @param columinar
+ * @param loadCount
+ * @param tableCreationTime
+ * @param schemaLastUpdatedTime
+ * @param prev
+ * @tparam K
+ * @tparam V
+ */
+class CarbonRDDDataLoadRDD[K, V](
```
jackylk: suggest to change to `DataFrameLoaderRDD`, it means the RDD to load the dataframe into carbon.
change `CarbonDataLoadRDD` to `DataFileLoaderRDD` or 'CSVFileLoaderRDD`. It matches loadDataFrame and loadDataFile function name accordinality.

integration/spark/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
jackylk: Can you also add a option in CarbonOption to cache the dataframe? So that user can decide whether to cache it. The name of the option can be `cache`, default is `false`

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataLoadRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
```diff
@@ -36,6 +38,8 @@ class CarbonOption(options: Map[String, String]) {
       "org.apache.carbondata.spark.partition.api.impl.SampleDataPartitionerImpl")
   }
 
+  def tempCSV: String = options.getOrElse("tempCSV", "false")
```
QiangCai: fixed

QiangCai:
CI PASS
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/382/
@jackylk please review

processing/src/main/java/org/apache/carbondata/processing/csvreaderstep/CsvInput.java
kumarvishal09: I feel putRow is not a overhead, We can take this as a part of kettle removal task.

kumarvishal09:
Falling back to local dictionary solution will cause feature like late materialization and push up would  have more suffles

sraghunandan:
How to support distributed and central cache if we go for local dictionary?

jackylk:
@sraghunandan @kumarvishal09 Please feel free to send a mail to dev mail list to discuss the local dictionary.

@QiangCai this PR looks good to me after fix comment, will merge it

PR #204, opened 2016-09-28
[CARBONDATA-280]Fix the bug that when table properties is repeated it only set the last one

Zhangshunyu:
## Why raise this pr?
When table properties is repeated it only set the last one, for example,
```
CREATE TABLE IF NOT EXISTS carbontable
(ID Int, date Timestamp, country String,
name String, phonetype String, serialname String, salary Int)
STORED BY 'carbondata'
TBLPROPERTIES('DICTIONARY_EXCLUDE'='country','DICTIONARY_INCLUDE'='ID',
'DICTIONARY_EXCLUDE'='phonetype', 'DICTIONARY_INCLUDE'='salary')
```
As we use map to store the properties, only salary is set to DICTIONARY_INCLUDE and only phonetype is set to DICTIONARY_EXCLUDE.

## How to solve?
**We should do restrict syntax check that 'DICTIONARY_EXCLUDE'='country,phonetype' , 'DICTIONARY_INCLUDE'='ID,salary**' and if table properties is repeated, throw an  MalformedCarbonCommandException to tell the user that Table properties is repeated, so that the user would not perform error operation.

## How to test?
Pass the exist test cases and the new test case for this bug.
## Test Result
CI has passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/354/testReport/

### DIFF
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/deleteTable/TestDeleteTableNewDDL.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/deleteTable/TestDeleteTableNewDDL.scala
@@ -97,7 +97,7 @@ class TestDeleteTableNewDDL extends QueryTest with BeforeAndAfterAll {
       "CREATE table CaseInsensitiveTable (ID int, date String, country String, name " +
       "String," +
       "phonetype String, serialname String, salary int) stored by 'org.apache.carbondata.format'" +
-      "TBLPROPERTIES('DICTIONARY_INCLUDE'='ID,salary')"
+      "TBLPROPERTIES('DICTIONARY_INCLUDE'='ID, salary')"
     )
     // table should drop wihout any error
     sql("drop table caseInsensitiveTable")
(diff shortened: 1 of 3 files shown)


ravipesala:
LGTM

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/deleteTable/TestDeleteTableNewDDL.scala
```diff
@@ -97,7 +97,7 @@ class TestDeleteTableNewDDL extends QueryTest with BeforeAndAfterAll {
       "CREATE table CaseInsensitiveTable (ID int, date String, country String, name " +
       "String," +
       "phonetype String, serialname String, salary int) stored by 'org.apache.carbondata.format'" +
-      "TBLPROPERTIES('DICTIONARY_INCLUDE'='ID', 'DICTIONARY_INCLUDE'='salary')"
+      "TBLPROPERTIES('DICTIONARY_INCLUDE'='ID,salary')"
```
jackylk: add space after `,`

Zhangshunyu:
@jackylk Fixed review comments and rebased, pls check.

jackylk:
LGTM, will merge

PR #206, opened 2016-10-02
[CARBONDATA-282] Add segment management in CarbonExample

jackylk:
Add segment delete using segment id and date example in CarbonExample
This PR will help user to understand how `DELETE SEGMENT` syntax is used


chenliang613:
@jackylk  Everything looks good, can you create another example for segment management.

jackylk:
@chenliang613 create a independent example is fine, it can show usage of compaction also

PR #208, opened 2016-10-03
[CARBONDATA-284] Abstracting index and segment interface

jackylk:
This PR adds new User API and Dev API for carbon-hadoop module:
### User API (org.apache.carbondata.hadoop.api package)
- `CarbonColumnarInputFormat/OutputFormat`: it uses current `CarbonInputFormat` as internal implementation. 
- `CarbonRowInputFormat/OutputFormat`: it needs to be implemented
- `CarbonOutputCommitter`: used for managing segment commit

They are based on `CarbonInputFormatBase/OutputFormatBase`
### Dev API (org.apache.carbondata.hadoop.internal package)
- Segment: an abstract class represents a single load of data,  used by CarbonInputFormatBase to get all InputSplit by matching QueryModel, and used by CarbonOutputCommitter to prepare for reading. Implementation examples are `IndexedSegment` and `StreamingSegment`.
- SegmentManager: an interface to manage segments. Current implementation is `ZkSegmentManager`, which need to be mapped to existing logic.
- Index: an interface that can is used by `IndexedSegment` to filter InputSplit. Current implementation is `InMemoryBTreeIndex` which load the index into driver's memory.

`CarbonInputFormatUtil` is modified so that it can also be used by both `CarbonInputFormat` and `CarbonColumnarInputFormat`.

Some of the implementation are still empty, please review on the interface to the get the abstraction right. Implementation will come in another PR.


hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/IndexedSegment.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.carbondata.hadoop.api.CarbonInputFormatBase;
+import org.apache.carbondata.hadoop.internal.index.Index;
+import org.apache.carbondata.hadoop.internal.index.IndexLoader;
+import org.apache.carbondata.scan.filter.resolver.FilterResolverIntf;
+import org.apache.carbondata.scan.model.QueryModel;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+
+public class IndexedSegment extends Segment {
```
ravipesala: How about segment with out index? do we should have one implementation class for that?

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/StreamingSegment.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.carbondata.hadoop.api.CarbonInputFormatBase;
+import org.apache.carbondata.scan.model.QueryModel;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+
+public class StreamingSegment extends Segment {
```
ravipesala: This segment is for appending data from stream right?  Why we should handle separately while reading, can't we handle as normal segment  outside but handle reading differently in carbon layer.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/row/CarbonRowInputFormat.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException;
+
+import org.apache.carbondata.hadoop.api.CarbonInputFormatBase;
+import org.apache.carbondata.hadoop.internal.segment.SegmentManager;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+public class CarbonRowInputFormat<T> extends CarbonInputFormatBase<T> {
```
ravipesala: is it for loading data row wise like old molap data? Then if we don't have plan to implement now better remove it.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/memory/InMemoryBTreeIndex.java
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.carbondata.scan.filter.FilterExpressionProcessor;
+import org.apache.carbondata.scan.filter.FilterUtil;
+import org.apache.carbondata.scan.filter.resolver.FilterResolverIntf;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+
+class InMemoryBTreeIndex implements Index {
```
ravipesala: Is this class used for pruning blocks based on secondary index?

ravipesala:
@jackylk Good work. All driver side Interfaces are clean apart from few comments and doubts.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormatBase.java
```diff
@@ -0,0 +1,69 @@
+/*
```
JihongMA: should we consolidate the interface as following, 
CarbonInputFormatBase
      --> CarbonInputFormat
                  --> CarbonColumnarInputFormat
                  --> CarbonRowInputFormat
or even further merging CarbonInputFormat into CarbonInputFormatBase

hadoop/src/main/java/org/apache/carbondata/hadoop/api/columnar/CarbonColumnarInputFormat.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.carbondata.hadoop.internal.segment.SegmentManager;
+import org.apache.carbondata.hadoop.util.CarbonInputFormatUtil;
+import org.apache.carbondata.scan.expression.Expression;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+public class CarbonColumnarInputFormat<T> extends CarbonInputFormatBase<T> {
+  private CarbonInputFormat<T> wrap;
```
JihongMA: suggest to change to a meaningful name: wrap --> carbonInputFormat

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/memory/InMemoryBTreeIndex.java
jackylk: No, it is the global B tree index for the table.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/row/CarbonRowInputFormat.java
jackylk: Row format is for streaming ingest. I can remove it now and submit it in future feature of streaming ingest

hadoop/src/main/java/org/apache/carbondata/hadoop/api/columnar/CarbonColumnarInputFormat.java
jackylk: The common logic of row & column format is the getSplit part, to iterate on all valid segment and getSplit of each one of them.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormatBase.java
jackylk: Yes, it is like that only

JihongMA:
we need interface for performing the actual filtering/scanning (file level/inputFormat level) as well as providing cost estimate for optimizer (table level/ Index) as discussed in the dev-mailing list. 

refer this link for interface reference:
https://www.postgresql.org/docs/9.0/static/index-cost-estimation.html
specificly btcostestimate in http://doxygen.postgresql.org/selfuncs_8c.html

 when it comes to the time to hook up with processing framework (like Spark), we might need to tweak here / there to fit it in.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/memory/InMemoryBTreeIndex.java
QiangCai: I understand InMemoryBTreeIndex  is segment level's index.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/memory/InMemoryBTreeIndex.java
```diff
@@ -0,0 +1,220 @@
...
+      TableBlockInfo tableBlockInfo = leafNode.getTableBlockInfo();
+      result.add(new CarbonInputSplit(segment.getId(), new Path(tableBlockInfo.getFilePath()),
+          tableBlockInfo.getBlockOffset(), tableBlockInfo.getBlockLength(),
+          tableBlockInfo.getLocations(), tableBlockInfo.getBlockletInfos().getNoOfBlockLets()));
+    }
+    return result;
+  }
+
+  private Map<String, AbstractIndex> getSegmentAbstractIndexs(JobContext job,
+      AbsoluteTableIdentifier identifier, String segmentId)
```
QiangCai: remove segmentId, directly use segment.id in this class

QiangCai:
I prefer to the new layout of carbondata files as following.
```
storelocation
    |__database_0
            |__table_0
                    |__partition_0
                            |__segment_0
                                    |__fact_0
                                    |__index_0
                                    |__......
```
Segment manager can include fact, index and so on. 
After this, we can extend carbon table on segment level.

jackylk:
@QiangCai I prefer not to modify the folder structure of Carbon, otherwise new version will not be backward compatible.

jackylk:
I have updated the interface, please review

jackylk:
Please review.
@ravipesala @QiangCai @JihongMA @chenliang613

jackylk:
@JihongMA  I have added a function in CarbonTableInputFormat to select the index to be used.
see https://github.com/apache/incubator-carbondata/pull/208/files#diff-1cb4c4e1751ea76cd7dcc596cde82abaR156

jackylk:
I have removed ColumnarInputFormat but keep a TableInputFormat only:
1. TableInputFormat contains multiple segments.
2. In getSplit function, it loop for all segments and delegate the getSplit to each segment. Segment can make use of the index to do file pruning.
3. In createRecordReader function, it create the record reader according to the format type in the split.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -0,0 +1,169 @@
...
+ * @param <T>
+ */
+public class CarbonTableInputFormat<T> extends FileInputFormat<Void, T> {
+
+  private static final String FILTER_PREDICATE =
+      "mapreduce.input.carboninputformat.filter.predicate";
+
+  private SegmentManager segmentManager;
+
+  public CarbonTableInputFormat(SegmentManager segmentManager) {
```
ravipesala: Better don't pass any thing from constructor, better create inside class by using job configuration. Because inputformat is mostly instantiated through reflection.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -0,0 +1,169 @@
...
+    this.segmentManager = segmentManager;
+  }
+
+  @Override
+  public RecordReader<Void, T> createRecordReader(InputSplit split,
+      TaskAttemptContext context) throws IOException, InterruptedException {
+    switch (((CarbonInputSplit)split).formatType()) {
```
ravipesala: Why don't you take the formatType from job conf? Better don't touch InputSplit as it comes from outside.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/CarbonFormat.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.carbondata.hadoop.internal;
+
+public enum CarbonFormat {
+  COLUMNR
```
ravipesala: typo : COLUMNAR

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/impl/InMemoryBTreeIndex.java
```diff
@@ -0,0 +1,215 @@
...
+    this.segment = segment;
+  }
+
+  @Override
+  public String getName() {
+    return null;
+  }
+
+  @Override
+  public List<InputSplit> filter(JobContext job, FilterResolverIntf filter)
```
ravipesala: It seems method return type is incompatible.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/index/impl/InMemoryBTreeIndex.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+
+class InMemoryBTreeIndex implements Index {
+
+  private static final Log LOG = LogFactory.getLog(InMemoryBTreeIndex.class);
+  private Segment segment;
+
+  InMemoryBTreeIndex(Segment segment) {
```
ravipesala: I guess we supposed to pass list of valid segments here.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
```diff
@@ -0,0 +1,73 @@
...
+  public List<InputSplit> getSplits(JobContext job, FilterResolverIntf filterResolver)
+      throws IOException {
+    // do as following
+    // 1. create the index or get from cache by the filter name in the configuration
+    // 2. filter by index to get the filtered block
+    // 3. create input split from filtered block
+
+    List<InputSplit> output = new LinkedList<>();
+    Index index = loader.load(job.getConfiguration());
+    List<Block> blocks = index.filter(job, filterResolver);
```
ravipesala: I guess either `loader.load(job.getConfiguration())` or `index.filter(job, filterResolver)`  should be passed with `segment`  to get the blocks.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
jackylk: accept

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
```diff
@@ -0,0 +1,73 @@
...
+  @Override
+  public List<InputSplit> getSplits(JobContext job, FilterResolverIntf filterResolver)
+      throws IOException {
+    // do as following
+    // 1. create the index or get from cache by the filter name in the configuration
+    // 2. filter by index to get the filtered block
+    // 3. create input split from filtered block
+
+    List<InputSplit> output = new LinkedList<>();
+    Index index = loader.load(job.getConfiguration());
```
jackylk: Yes, actually the IndexLoader is the factory, `load` function has the same functionality as `getInstance` in the factory. So I will keep the IndexLoader name?

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
jackylk: Actually this Index is per Segment.
You can see current implementation there is a map to map index to segment, so it is the same.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/Segment.java
```diff
@@ -0,0 +1,94 @@
...
+    return path;
+  }
+
+  /**
+   * return all InputSplit of this segment, each file is a InputSplit
+   * @param job job context
+   * @return all InputSplit
+   * @throws IOException
+   */
+  public List<InputSplit> getAllSplits(JobContext job) throws IOException {
```
QiangCai: I suggest to return List<CarbonInputSplit>

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
```diff
@@ -0,0 +1,73 @@
...
+ * under the License.
+ */
+
+package org.apache.carbondata.hadoop.internal.segment.impl;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.apache.carbondata.hadoop.CarbonInputSplit;
```
QiangCai: please use internal.CarbonInputSplit

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
ravipesala: if loader internally implement cache then we can keep as `IndexLoader` only.

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
ravipesala: My question was how `Segment` object is conveyed to `Index` like `InMemoryBTreeIndex`. Here `InMemoryBTreeIndex` takes `segment` object to constructor. How that `InMemoryBTreeIndex` without passing from loader.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
ravipesala: ok

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
jackylk: You are right, I will modify

hadoop/src/main/java/org/apache/carbondata/hadoop/internal/segment/impl/IndexedSegment.java
jackylk: ok

ravipesala:
LGTM

PR #211, opened 2016-10-04
[CARBONDATA-282] Add data management example

jackylk:
A new example for data management is added.
It will help user to understand how following syntax is used:
- DELETE SEGMENT BY ID
- DELETE SEGMENT BY DATE
- COMPACTION
- CLEAN FILES


chenliang613:
LGTM.
This new example is introducing detailed segment management.

PR #212, opened 2016-10-04
[CARBONDATA-285] Use path parameter in Spark datasource API

jackylk:
Currently, when using carbon with spark datasource API, it need to give database name and table name as parameter, it is not the normal way of datasource API usage. In this PR, database name and table name is not required to give, user need to specify the `path` parameter (indicating the path to table folder) only when using datasource API

before:

```
df.write
   .format("carbondata")
   .option("tableName", tableName)
   .mode(SaveMode.Overwrite)
   .save()
```

after:

```
df.write
   .format("carbondata")
   .mode(SaveMode.Overwrite)
   .save("store_path/default/table_name")
```

The same for read interface.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
@@ -36,7 +36,7 @@ import org.apache.carbondata.core.carbon.metadata.schema.table.column.CarbonDime
 import org.apache.carbondata.core.carbon.path.CarbonStorePath
 import org.apache.carbondata.core.datastorage.store.impl.FileFactory
 import org.apache.carbondata.lcm.status.SegmentStatusManager
-import org.apache.carbondata.spark.{CarbonOption, _}
+import org.apache.carbondata.spark._
 
 /**
  * Carbon relation provider compliant to data source api.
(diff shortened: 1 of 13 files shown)


chenliang613:
@jackylk  everything looks great.
One comment:Can you list what are the exact benefits for this feature?  main purpose is for saving as to different folder , not CarbonData's default folder ?

As per my understanding:
before : the database name is using the default, users are not aware of it, users don't need to input database as parameter.

So i feel that both solution just need to input table name.

jackylk:
Since CarbonData is a file format, so ideally it should support saving it to any folder. Currently we are forcing database and table folder structure, it makes the usage different from other Spark file datasource like parquet and orc. 
And I think by accepting only path parameter, it is easier to integrate with Spark 2.0 and other processing framework

integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
```diff
@@ -55,18 +55,11 @@ class CarbonSource extends RelationProvider
...
-      case _ =>
-        val options = new CarbonOption(parameters)
-        val tableIdentifier = options.tableIdentifier.split("""\.""").toSeq
-        val identifier = tableIdentifier match {
-          case Seq(name) => TableIdentifier(name, None)
-          case Seq(db, name) => TableIdentifier(name, Some(db))
-        }
-        CarbonDatasourceRelation(identifier, None)(sqlContext)
+    val options = new CarbonOption(parameters)
+    if (sqlContext.isInstanceOf[CarbonContext]) {
```
ravipesala: I don't think we can load the data using context type, earlier we are checking `path` parameter and then decides whether to use Hadoop relation or not. Because we could use dataframe.load(path). But now it can't work if we use carboncontext.load(path)

jackylk: There is no `load` method in dataframe, only in context class.

ravipesala: sorry, yes `carboncontext.load(path)` cannot work now right?

jackylk: It works, please check DataFrameAPIExample.scala

ravipesala: Ok, got it. You have added `path` in `options` while creating datasource table to make it work.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -861,9 +861,11 @@ private[sql] case class CreateTable(cm: tableModel) extends RunnableCommand {
       val tablePath = catalog.createTableFromThrift(tableInfo, dbName, tbName, null)(sqlContext)
       try {
         sqlContext.sql(
-          s"""CREATE TABLE $dbName.$tbName USING carbondata""" +
-          s""" OPTIONS (tableName "$dbName.$tbName", tablePath "$tablePath") """)
-              .collect
+          s"""
+             | CREATE TABLE $dbName.$tbName
+             | USING carbondata
+             | OPTIONS (path "$tablePath")
```
jackylk: Ok, I think we should not keep both approach neither. Then I think we can keep the current approach. So shall I close this PR?

jackylk:
This will be done in another PR for spark2 integration
closing it

PR #214, opened 2016-10-06
[WIP] Add a tool to print out schema of CarbonData

jackylk:
Add a tool to print out schema information, taking carbondata table folder as input parameter

```
usage: carbon-ls path
arguments:
    path: path to carbondatat files
```

For example, for the table generated using `ExampleUtils.writeSampleCarbonFile`, it will print following information:

```
======= table information =======
path of this table: /Users/jackylk/code/incubator-carbondata/tools/target/store/default/t1
last updated time: 1475755645305
partition count: 1
block size: 1024
======= schema information =======
dimension column: {name: c1, data type: STRING, columnar: true, encoding: [DICTIONARY], inverted index: true, ordinal: 0, id: d5345f2e-ce2c-40f0-99e8-39ee07f356da, default value: null}
dimension column: {name: c2, data type: STRING, columnar: true, encoding: [DICTIONARY], inverted index: true, ordinal: 1, id: 0ddb30d9-7d9f-4c2c-b253-9a5a4bdebbfb, default value: null}
measure column: {name: c3, data type: INT, columnar: true, encoding: [], inverted index: false, ordinal: 0, id: 5c9010de-0006-46dd-918a-72e7a2f0a2c8, default value: null}
```

I will create JIRA later


ravipesala:

PR #216, opened 2016-10-08
[CARBONDATA-289]Support MB/M for table block size and update the doc about this new feature.

Zhangshunyu:
## Why raise this pr?
1. Support MB/M for table blocksize.
2. To Update DOC about table level blocksize, the code has been merged, but doc is not updated.

## How to test?
Pass all the test cases.

### DIFF
--- a/docs/DDL-Operations-on-Carbon.md
+++ b/docs/DDL-Operations-on-Carbon.md
@@ -67,6 +67,14 @@ Here, DICTIONARY_EXCLUDE will exclude dictionary creation. This is applicable fo
   ```ruby
   TBLPROPERTIES ("COLUMN_GROUPS"="(column1,column3),(Column4,Column5,Column6)") 
   ```
+ - **Table Block Size Configuration**
+
+   The block size of one table's files on hdfs can be defined using an int value whose size is in MB, the range is form 1MB to 2048MB and the default value is 1024MB, if user didn't define this values in ddl, it would use default value to set.
+
+  ```ruby
+  TBLPROPERTIES ("TABLE_BLOCKSIZE"="512")
+  ```
+Here 512 means the block size of this table is 512MB.
  - **Inverted Index Configuration**
 
    Inverted index is very useful to improve compression ratio and query speed, especially for those low-cardinality columns who are in reward position.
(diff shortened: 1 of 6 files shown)


docs/DDL-Operations-on-Carbon.md
```diff
@@ -67,6 +67,14 @@ Here, DICTIONARY_EXCLUDE will exclude dictionary creation. This is applicable fo
   ```ruby
   TBLPROPERTIES ("COLUMN_GROUPS"="(column1,column3),(Column4,Column5,Column6)") 
   ```
+ - **Table Block Size Configuration**
+
+   The block size of one table's files on hdfs can be defined using an int value whose size is in MB, the range is form 1MB to 2048MB and the default value is 1024MB, if user didn't define this values in ddl, it would use default value to set.
+
+  ```ruby
+  TBLPROPERTIES ("TABLE_BLOCKSIZE"="512")
```
jackylk: I think it is better to have the MB inside the value, it is more clear to the user. For example:
`TBLPROPERTIES ("TABLE_BLOCKSIZE"="512MB")`
And we can also add support for GB in the future, if required. 
Can you modify the code accordingly

Zhangshunyu: @jackylk 
Fixed, support MB/M now, pls check. 
As to GB, i think no need to support now because the range is 1M-2048M if set in GB it would be 0.000976GB-2GB.

Zhangshunyu:
CI has passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/395/testReport/

Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/396/testReport/

docs/DDL-Operations-on-Carbon.md
```diff
@@ -67,6 +67,14 @@ Here, DICTIONARY_EXCLUDE will exclude dictionary creation. This is applicable fo
   ```ruby
   TBLPROPERTIES ("COLUMN_GROUPS"="(column1,column3),(Column4,Column5,Column6)") 
   ```
+ - **Table Block Size Configuration**
+
+   The block size of one table's files on hdfs can be defined using an int value whose size is in MB, the range is form 1MB to 2048MB and the default value is 1024MB, if user didn't define this values in ddl, it would use default value to set.
+
+  ```ruby
+  TBLPROPERTIES ("TABLE_BLOCKSIZE"="512 MB")
```
jackylk: please remove the space, I think `512MB` is ok

Zhangshunyu: @jackylk here we support 512mb/512m/512 mb/ etc.

jackylk: ok

PR #217, opened 2016-10-08
[CARBONDATA-287]Using multi local directorys to improve dataloading perfomance

QiangCai:
**1  Analysis**

Now for each dataloading, we use only one different local directory to save the sorted temp files. I think it is neccessary to use multi local directorys in multi disks for each dataloading to improve dataloading performance.

**2 Solution**

Modify SortDataRows class to separate the sorted temp files into different local directorys in different disks..


jackylk:
@QiangCai I think it is preferable to make this work with new loading flow without kettle

QiangCai:
close this pr

PR #220, opened 2016-10-09
[CARBONDATA-290]Fix path not found when part of table name  has database Name

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-290

this issue is because when part of table name has database name, then substring(0, tablePath.lastIndexOf(dbName))  can not get the correct store path, so will cause segment path not found.
regarding that table name will have no '/', so we can use this to get substring.

testcase addded, CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/397/


jackylk:
LGTM

PR #225, opened 2016-10-10
[CARBONDATA-295]Abstract Compressor interface and let Snappy interface extends from Compressor

Zhangshunyu:
## Why raise this pr?
Currently, we only have snappy compressor who extends form Compressor interface, for future expansion, we need to abstract Snappy interface and seperate it from Compressor interface, it means `Compressor interface is the parent of all compressors, and SnappyCompressor interface and the other compressor's interface(or abstract class) should extends Compressor interface, as to different data type for different compressor, it would extend its own interface/abstract class.`

For example: Compressor -> SnappyCompressor ->  SnappyDoubleCompression.

In the future, we can define abstract class B for other compressor who extends from Compressor A, and define compressors implimentations C1/C2/C3/C4 .etc for this kind of compressor.

## How to test?
Pass all the test cases.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/SnappyCompression.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastorage/store/compression/SnappyCompression.java
@@ -36,7 +36,7 @@ public class SnappyCompression {
   /**
    * SnappyByteCompression.
    */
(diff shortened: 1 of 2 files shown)


Zhangshunyu:
CI has passed, pls check:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/401/testReport/

PR #227, opened 2016-10-11
[CARBONDATA-304] Fixed data loading failure when set table_blocksize=2048

foryou2030:
# Why raise this pr?
Load data failure when set table_blocksize=2048
# How to solve?
if blocksize=2048, then `2048 * 1024 * 1024` will beyond the range of Int, the result is `-2147483648`, so it will use the file size as block size. But once the file size is smaller than `1Mb (dfs.namenode.fs-limits.min-block-size 1048576) ` , create hdfs file failed.
So use `1L * 2048 * 1024 * 1024` instead of` 2048 * 1024 * 1024 * 1L`

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/store/writer/AbstractFactDataWriter.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/store/writer/AbstractFactDataWriter.java
@@ -197,8 +197,9 @@ public abstract class AbstractFactDataWriter<T> implements CarbonFactDataWriter<
     blockIndexInfoList = new ArrayList<>();
     // get max file size;
     CarbonProperties propInstance = CarbonProperties.getInstance();
-    this.fileSizeInBytes = blocksize * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR
-        * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR * 1L;
+    // if blocksize=2048, then 2048*1024*1024 will beyond the range of Int
+    this.fileSizeInBytes = 1L * blocksize * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR
+        * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR;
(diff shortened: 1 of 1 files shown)


foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/408/

Zhangshunyu:
LGTM

processing/src/main/java/org/apache/carbondata/processing/store/writer/AbstractFactDataWriter.java
```diff
@@ -197,8 +197,9 @@ public AbstractFactDataWriter(String storeLocation, int measureCount, int mdKeyL
     blockIndexInfoList = new ArrayList<>();
     // get max file size;
     CarbonProperties propInstance = CarbonProperties.getInstance();
-    this.fileSizeInBytes = blocksize * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR
-        * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR * 1L;
+    // if blocksize=2048, then 2048*1024*1024 will beyond the range of Int
+    this.fileSizeInBytes = 1L * blocksize * CarbonCommonConstants.BYTE_TO_KB_CONVERSION_FACTOR
```
jackylk: instead of multiple by `1L`, you can just convert `blocksize` to `(long)blocksize` instead

foryou2030: fixed

foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/417/

jackylk:
LGTM

manishgupta88:
@foryou2030 ...please add a test case for this scenario

Zhangshunyu:
@manishgupta88 Hi manish, as our CI in run in local mode, and it does not use block to create file, this parameter only runs on hdfs, so test case in useless on CI, and it is no need to and this test case.

gvramana:
LGTM

PR #229, opened 2016-10-11
[CARBONDATA-297]Added interface for carbon data loading

ravipesala:
Interface DataLoadProcessorStep and related supported classes are added in this PR.


processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  void intialize(CarbonDataLoadConfiguration configuration, DataLoadProcessorStep child) throws
+      CarbonDataLoadingException;
+
+  /**
+   * Tranform the data as per the implemetation.
+   * @return Iterator of data
+   * @throws CarbonDataLoadingException
+   */
+  Iterator<Object[]> execute() throws CarbonDataLoadingException;
```
jackylk: I think `execute()` is called for every parallel unit of the input, right? For example, when using spark to load from dataframe, `execute()` is called for every spark partition (execute one task for one partition). When loading from CSV HDFS file, `execute()` is called for every HDFS block. So I do not think returning array of iterator is required.

The loading process of carbon in every executor, some of the step can be parallelized, but sort step need to be synchronized (potential bottle net), since we need datanode-scope sorting. Am I correct?

processing/src/main/java/org/apache/carbondata/processing/newflow/CarbonDataLoadConfiguration.java
```diff
@@ -0,0 +1,185 @@
+package org.apache.carbondata.processing.newflow;
```
jackylk: add license header

processing/src/main/java/org/apache/carbondata/processing/newflow/CarbonDataLoadConfiguration.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.Iterator;
+
+import org.apache.carbondata.core.carbon.AbsoluteTableIdentifier;
+
+public class CarbonDataLoadConfiguration {
```
jackylk: It seems this configuration is quite complex, I think it is because it contains configuration for all steps. 
Can we just have a simple `Map` as the configuration and let the `Step` decide what to keep in it?

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Iterator;
+
+import org.apache.carbondata.processing.newflow.exception.CarbonDataLoadingException;
+
+/**
+ * This base interface for data loading. It can do transformation jobs as per the implementation.
+ *
+ */
+public interface DataLoadProcessorStep {
```
jackylk: I think each implementation of this interface have similar logic in the execute function, can we create a abstract class to implement the common logic? 
The common logic like:
```
Iterator<Object[]> execute() throws CarbonDataLoadingException {
    Iterator childIter = child.execute();
    return new Iterator<Object[]> {
          public boolean hasNext() { 
                return childIter.hasNext();
          }
          public Object[] next() {
                // processInput is the abstract func in this class
                return processInput(childItor.next());
          }
     }
}
```

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  DataField[] getOutput();
+
+  /**
+   * Intialization process for this step.
+   * @param configuration
+   * @param child
+   * @throws CarbonDataLoadingException
+   */
+  void intialize(CarbonDataLoadConfiguration configuration, DataLoadProcessorStep child) throws
```
jackylk: If there is a abstract class, it can have the child as its member variable, then this `initialize` function takes no parameter as input

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
```diff
@@ -0,0 +1,40 @@
...
+   * Tranform the data as per the implemetation.
+   * @return Iterator of data
+   * @throws CarbonDataLoadingException
+   */
+  Iterator<Object[]> execute() throws CarbonDataLoadingException;
+
+  /**
+   * Any closing of resources after step execution can be done here.
+   */
+  void finish();
```
jackylk: This is called when the step successfully finished. But what about failure case, should there be a 
`void close();` 
interface for failure case?

processing/src/main/java/org/apache/carbondata/processing/newflow/iterators/RecordReaderIterator.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.carbondata.common.CarbonIterator;
+import org.apache.carbondata.common.logging.LogService;
+import org.apache.carbondata.common.logging.LogServiceFactory;
+
+import org.apache.hadoop.mapred.RecordReader;
+
+/**
+ * This iterator iterates RecordReader.
+ */
+public class RecordReaderIterator extends CarbonIterator<Object[]> {
```
jackylk: why this is carbon-processing but not carbon-hadoop module?

jackylk: what is it used for?

processing/src/main/java/org/apache/carbondata/processing/newflow/iterators/CarbonArrayWritable.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+import org.apache.hadoop.io.Writable;
+
+/**
+ * It is hadoop's writable value wrapper.
+ */
+public class CarbonArrayWritable implements Writable {
```
jackylk: Why this is carbon-processing but not carbon-hadoop module? All hadoop related class should be in carbon-hadoop module

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
ravipesala: For suppose if we are loading 50GB of csv files and each HDFS block size is 256MB then total number of partitions are 200. If we allow one task per partition then it would be 200 tasks. In carbondata one btree is created for each task. So if we allow all 200 tasks then it would be massively 200 btrees and it is not effective both in performance and memory wise. 
That is the reason why we pool multiple blocks per task in the current kettle implementation. And these blocks are processed parallely. We can take the same way and use iterator for each thread and returns array of iterator.

What do you mean by datanode-scope sorting? how to synchronize between multiple tasks?

processing/src/main/java/org/apache/carbondata/processing/newflow/CarbonDataLoadConfiguration.java
ravipesala: Ok. I will keep the configuration but only keep important and global configurations I will keep in it, remaining configurations we can move to `Map` and keep it inside configuartion it self.  what do you say?

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
ravipesala: ok

processing/src/main/java/org/apache/carbondata/processing/newflow/DataLoadProcessorStep.java
jackylk: Then how implementation differentiate success or failure case? I think it need to release resource in failure case.

ravipesala: Ok. I will add close method along with finish method

processing/src/main/java/org/apache/carbondata/processing/newflow/iterators/RecordReaderIterator.java
ravipesala: It is used for iterating RecordReader. I can move it to carbon-hadoop module but processing module need to be dependent on it. Already processing module is dependent on hadoop module so it becomes dependent on each other.

jackylk: I thought this PR is for loading part, why RecordReader related abstraction is required?

ravipesala:
@jackylk Fixed review comments, please review.

processing/src/main/java/org/apache/carbondata/processing/newflow/AbstractDataLoadProcessorStep.java
```diff
@@ -73,15 +72,15 @@ public AbstractDataLoadProcessorStep(CarbonDataLoadConfiguration configuration,
    * Create the iterator using child iterator.
    *
    * @param childIter
-   * @return
+   * @return new iterator with step specific processing.
    */
-  protected Iterator<Object[]> getIterator(final Iterator<Object[]> childIter) {
-    return new CarbonIterator<Object[]>() {
+  protected Iterator<CarbonRow> getIterator(final Iterator<CarbonRow> childIter) {
```
ravipesala: ok. Added

jackylk:
LGTM

PR #231, opened 2016-10-13
[CARBONDATA-311]Log the data size of blocklet during data load.

Zhangshunyu:
## Why raise this pr?
The blocklet size is an important parameter for analyzing data load and query, this info should be logged.
## How to test?
Pass all the test case.


processing/src/main/java/org/apache/carbondata/processing/store/writer/CarbonFactDataWriterImplForIntIndexAndAggBlock.java
```diff
@@ -250,6 +254,7 @@ public NodeHolder buildDataNodeHolder(IndexStorage<int[]>[] keyStorageArray, byt
     }
     long blockletDataSize =
         holder.getKeyArray().length + holder.getDataArray().length + indexBlockSize;
+    LOGGER.info("A new blocklet is added, its data size is: " + blockletDataSize + " Byte");
```
Vimal-Das: Move this line after call to writeDataToFile, because if writeDataToFile fails, we should not log that a blocklet is added.

Zhangshunyu:
@Vimal-Das Thanks for your kindly suggestion, fixed already, pls check.

jackylk:
LGTM

PR #232, opened 2016-10-13
[CARBONDATA-310]Fixed compilation failure when using spark 1.6.2

foryou2030:
# Why raise this pr?
Compilation failed when using spark 1.6.2, because class not found: AggregateExpression
# How to solve?
Once Removing the import "import org.apache.spark.sql.catalyst.expressions.aggregate._" will cause compilation failure when using Spark 1.6.2, in which AggregateExpression is moved to subpackage "aggregate". So neeed changing it back.

Thanks for you remind, @harperjiang


foryou2030:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/420/

PR #234, opened 2016-10-13
[CARBONDATA-315] Data loading fails if parsing a double value returns infinity

manishgupta88:
Problem: Data loading fails if parsing a double value returns infinity

Analysis: During data load, if a value specified is too big for a double DataType column then while parsing that value as double result is returned as "Infinity". Due to this while we calculate min and max value for measures in carbon data writer step it throws an exception.

Fix: If result is Infinity or NAN for double value parsing then make the value as null and add it to bad records.

Impact area: Data load which contains non parseable values for a datatype.


manishgupta88:
CI Passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/421/

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -994,29 +994,31 @@ private String getCarbonLocalBaseStoreLocation() {
...
-                    null :
-                    DataTypeUtil
-                        .getMeasureValueBasedOnDataType(msr, msrDataType[meta.msrMapping[msrCount]],
-                            meta.carbonMeasures[meta.msrMapping[msrCount]]);
-          } catch (NumberFormatException e) {
-            try {
-              msr = msr.replaceAll(",", "");
-              out[memberMapping[dimLen + index] - meta.complexTypes.size()] = DataTypeUtil
+            if (!isNull && null != msr && msr.length() > 0) {
+              Object measureValueBasedOnDataType = DataTypeUtil
```
jackylk: put `DataTypeUtil` in next line

jackylk:
LGTM
CI passed: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/432/

PR #236, opened 2016-10-13
[CARBONDATA-299] Add dictionary interface for new data load flow

jackylk:
In this PR, following is added:
- a BiDictionary interface which can retrieve key by value, or retrieve value by key.
- DictionaryGenerator interface that generates dictionary key from value
- an implementation (InMemBiDictionary) that uses BiMap from Google Guava library as backend data structure. This implementation also supports pre-created dictionary.

Testcase is added for InMemBiDictionary


core/src/main/java/org/apache/carbondata/core/devapi/BiDictionary.java
```diff
@@ -0,0 +1,53 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.carbondata.core.devapi;
+
+public interface BiDictionary<K, V> {
```
ravipesala: what is the meaning of `Bi`, not understandable

processing/src/main/java/org/apache/carbondata/processing/newflow/dictionary/InMemBiDictionary.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Map;
+
+import org.apache.carbondata.core.devapi.DictionaryGenerator;
+import org.apache.carbondata.core.devapi.GeneratingBiDictionary;
+
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+
+public class InMemBiDictionary<K, V> extends GeneratingBiDictionary<K, V> {
```
ravipesala: we can add DictionaryCache based implementation. we may not require this implementation now. We can add when we need it.

core/src/main/java/org/apache/carbondata/core/devapi/BiDictionary.java
jackylk: It mean Bidirectional, I borrow the term from Guava library.
see [BiMap](https://github.com/google/guava/wiki/NewCollectionTypesExplained#bimap)

ravipesala:
@jackylk size() or maxDictionaryValue() is required in dictionary interface to compose mdkey

jackylk:
@ravipesala ok, will add it

core/src/main/java/org/apache/carbondata/core/devapi/GeneratingBiDictionary.java
```diff
@@ -0,0 +1,49 @@
...
+  private DictionaryGenerator<K, V> generator;
+
+  public GeneratingBiDictionary(DictionaryGenerator<K, V> generator) {
+    this.generator = generator;
+  }
+
+  @Override
+  public K getOrGenerateKey(V value) throws Exception {
+    if (getKey(value) != null) {
+      return getKey(value);
```
ravipesala: `getKey` called twice. Please assign to local variable.

ravipesala:
LGTM except one minor comment.

jackylk:
fixed comment
rebased
squashed commits

ravipesala:
LGTM

PR #237, opened 2016-10-13
[CARBONDATA-317] - CSV having only space char is throwing NullPointer…

mohammadshahidkhan:
Problem: Data loading fails if csv is having only empty chars

Analysis: During data load, in global dictionary generation step while reading the first line for the headers 
its getting null value from the LineCsvReader.

Solution : Added null check over first row and throwing exception asking to provide valid csv.
"Please check your input path and make sure csv file contains valid header."
Impact area: Data load with csv having only space char in first line.


integration/spark/src/main/scala/org/apache/carbondata/spark/csv/CarbonCsvRelation.scala
```diff
@@ -148,6 +150,10 @@ case class CarbonCsvRelation protected[spark] (
           .withSkipHeaderRecord(false)
         CSVParser.parse(firstLine, csvFormat).getRecords.get(0).asScala.toArray
       }
+      if(null == firstRow) {
+        throw new DataLoadingException("Please check your input path and make sure " +
```
kumarvishal09: May be csv file does not have header and user is passing the header from load command in that case is this a valid message??

gvramana: "Give a message First line cannot be empty",  irrespective of header present or not.

gvramana:
LGTM

mohammadshahidkhan:
PR builder passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/456/

gvramana:
LGTM

PR #241, opened 2016-10-15
[CARBONDATA-319] Bad Records logging for column data type is not proper

sujith71955:
**Problem:**
Bad Records logging for column data type is not proper in case of long, carbon system while creating the table metadata uses BIGINT instead of LONG , internally it converts the bigint to long type and processes , while processing the data if any long type data is having issue it will be logged into the bad record with data type Long which is not proper since as per the metadata the datatype of column is BIGINT.
**Solution :**
Pass name for each data type ENUM constant in constructor, while logging the bad record get the column data type name value from the data type ENUM constant, for LONG the name will be BIGINT.

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
@@ -1237,6 +1237,7 @@ public class CarbonCSVBasedSeqGenStep extends BaseStep {
 
   private void addEntryToBadRecords(Object[] r, int j, String columnName,
       String dataType) {
+    dataType= DataTypeUtil.getColumnDataTypeDisplayName(dataType);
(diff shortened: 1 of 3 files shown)


sujith71955:
please find the pr builder result.
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/443/

processing/src/main/java/org/apache/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1237,6 +1237,9 @@ else if(isComplexTypeColumn[j]) {
 
   private void addEntryToBadRecords(Object[] r, int j, String columnName,
       String dataType) {
+    if (DataType.LONG.toString().equals(dataType)) {
```
gvramana: This check is not required.

sujith71955:
Comments handled, Please find the PR Build status.
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/450/

gvramana:
LGTM

PR #242, opened 2016-10-16
[CARBONDATA-318] Implement in memory sort for data load

jackylk:
Changed as following:
1. Change SortDataRows.java to keep rows and sort in memory if memory is sufficient, otherwise spill to disk.
2. Change SortKeyStep and MdkeyGenStep to support both in memory sort and merge sort.

To choose between these two approaches, user can set SORT_SIZE in carbon property, like set it to 3 million rows:

```
  // Number of rows to keep in memory when loading data, if number of input row exceeds this value,
  // carbon will use merge sort instead of in memory sort
CarobonPropery.getInstance().addProperty(CarbonCommonConstants.SORT_SIZE, "3000000")
```


kumarvishal09:
@jackylk I have one doubt at a time number of records to sort is very less suppose 1 million and in that case we can sort data in memory , but what about scenario like when multiple table is getting loaded at the same time in that case memory consumption will be two high if all the tables are keeping all the data  in memory  and java Arrays sorting is out-of-place sorting so at a time for suppose 1 millions records per table then 10 tables data will be 10 million and as out-of-sorting will make it 20 million so in that case it may go out of memory or it may cause to much gc

jackylk:
Yes, so actually I was thinking a online sorting algorithm is better for our scenario, like TreeMap. And we can allocate the buffer offheap to reduce GC. 
But these technique reqires more effort. So other PR is needed in the future.

kumarvishal09:
Better to have One interface for sorting based on the memory we can store the data in file or in memory , move the merge sort part from mdkey and add  in sort step once all the data is processed we can passed row by row to mdkey and mdkey will generate the mdkey and passed it to writer

kumarvishal09:
we can do the same way how spark does unsafe sorting but I am not clear about some part in it , while sorting we need to move and swap rows in that case row will be copied from non-heap space to heap space??

jackylk:
@kumarvishal09 I agree with your points. 
1. Let's move all sort logic into a SortStep including the merge part then pass the row to MDKeyGenStep
2. We can do the similar way as Spark. In spark, there are [`ExternalSorter`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/collection/ExternalSorter.scala) and [`UnsafeInMemorySorter`](https://github.com/apache/spark/blob/master/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/UnsafeInMemorySorter.java) for safe and unsafe version. `ExternalSorter` internally uses [`TimSort`](https://github.com/apache/spark/blob/master/core/src/main/java/org/apache/spark/util/collection/TimSort.java), you  can see descrption of `sort` function about the space usage:
```
In the worst case, this sort requires
 temporary storage space for n/2 object references; in the best case,
  it requires only a small constant amount of space.
```

And for `UnsafeInMemorySorter`, it is sorting using a key prefix to approve CPU cache efficiency:
```
/**
 * Sorts records using an AlphaSort-style key-prefix sort. This sort stores pointers to records
 * alongside a user-defined prefix of the record's sorting key. When the underlying sort algorithm
 * compares records, it will first compare the stored key prefixes; if the prefixes are not equal,
 * then we do not need to traverse the record pointers to compare the actual records. Avoiding these
 * random memory accesses improves cache hit rates.
 */
```

jackylk:
This will be done in another PR
closing it

PR #243, opened 2016-10-16
[CARBONDATA-323] load data not support local

scwf:
Seems we do not support `load data local`, so fix it.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -922,7 +922,7 @@ class CarbonSqlParser()
   }
 
   protected lazy val loadDataNew: Parser[LogicalPlan] =
-    LOAD ~> DATA ~> opt(LOCAL) ~> INPATH ~> stringLit ~ opt(OVERWRITE) ~
+    LOAD ~> DATA ~> INPATH ~> stringLit ~ opt(OVERWRITE) ~
       (INTO ~> TABLE ~> (ident <~ ".").? ~ ident) ~
       (OPTIONS ~> "(" ~> repsep(loadOptions, ",") <~ ")").? <~ opt(";") ^^ {
         case filePath ~ isOverwrite ~ table ~ partionDataOptions =>


kumarvishal09:
Can u tell me the load query.

scwf:
@kumarvishal09 in the parser   `opt(LOCAL) ~>`  this option never used.

Vimal-Das:
@scwf Can you add test case as well for the LOCAL scenario.

scwf:
@Vimal-Das  from the sytax carbon could not support local, it is necessary to add a test for this?

jackylk:
Yes, local is not supported. But your modification may intoduce incompatibility, can you modify it as parsing failure (unsupported feature) in case of local?

scwf:
It will parsing failure with this modification.

jackylk:
Thanks, and please fire a Jira issue and add it to the PR title

scwf:
And i do not think this is related to compatibility, actually now the `load data local` is wrong in syntax since the load action is not done from local data. If users have used the syntax, they just used a wrong syntax that not run as they expected, we should fix this.

jackylk:
Agree. Please fire a Jira issue. Thanks for fixing it.

scwf:
ok, later file a jira for this issue:)

scwf:
@jackylk carbon did not have a jenkens for CI

jackylk:
@scwf please refer to other PR how they do it.
Apache CI is not ready for Carbon yet, meanwhile we have setup a CI enviornment in India as a temporary solution, please use this [CI] (http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/)

modify your PR number and paste the result in your PR

chenliang613:
@scwf Thanks for your PR
please find the CI result : http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/470/
There are two errors.

Zhangshunyu:
There are 2 test cases in TestLoadDataWithFileHeaderException file，who used the ddl including “local”，pls also delete the local in them，so that the test case can pass.

ravipesala:

scwf:
i will close this, and in spark2 integration we do not support load local for carbon, in spark1 i think there is no problem

PR #245, opened 2016-10-17
[CARBONDATA-320] problem during drop of table when all datanodes are down.

ravikiran23:
Problem : 

when all the data nodes are down and user executes drop table. then drop table will some times fail to delete the files from hdfs. but it will remove the hive entry. 

Solution : 

Removed the unnecessary checks present in the drop table . so that the deletion of the files wont be skipped. 

changed the lock type of the drop to new lock type.


ravikiran23:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/447/

vinodkc:
LGTM

gvramana:
LGTM

PR #246, opened 2016-10-17
[CARBONDATA-321] If user changes the blocklet size the queries will b…

sujith71955:
**Problem:**
If user changes the blocklet size (50) the queries will be failed.
Currently byte size calculation in number compressor is 
int array length * bit length=x
bytesize= x/8+x%8;
when cardinality is 50
then byte size will be 3
but to calculate the actual int array size of index reverse based on the reverse calculation it is coming 4
int arrayLength=(3*8)/6
**Solution:**
We need to round off bit length to nearest value which should be divisible by 8


kumarvishal09:
LGTM.Please add test case for multiple scenario like blocklte size 50, 10,100

sujith71955:
Added testcase, Please find the PR status
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/org.apache.carbondata$carbondata-spark/45

gvramana:
This fix will make previous store not readable, Does not support backward compatibility.

PR #251, opened 2016-10-20
[CARBONDATA-302]Added Writer processor step for dataloading.

ravipesala:
Add DataWriterProcessorStep which reads the data from sort processor step and apply mdk generator on key and creates carbondata files.

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
@@ -110,17 +110,7 @@ public class DataWriterProcessorStepImpl extends AbstractDataLoadProcessorStep {
     return child.getOutput();
   }
 
-  @Override public void intialize() throws CarbonDataLoadingException {
-    setStepConfiguration();
-  }
-
-  /**
-   * This method will be used to get and update the step properties which will
-   * required to run this step
-   *
-   * @throws CarbonUtilException
-   */
-  private void setStepConfiguration() {
+  @Override public void initialize() throws CarbonDataLoadingException {
     CarbonTableIdentifier tableIdentifier =
         configuration.getTableIdentifier().getCarbonTableIdentifier();
 
@@ -365,10 +355,6 @@ public class DataWriterProcessorStepImpl extends AbstractDataLoadProcessorStep {
 
   @Override protected CarbonRow processRow(CarbonRow row) {
     return null;
-  }
-
-  @Override public void finish() {
-
   }
 
 }
--- a/processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
@@ -40,7 +40,6 @@ import org.apache.carbondata.core.keygenerator.KeyGenerator;
 import org.apache.carbondata.core.util.CarbonProperties;
 import org.apache.carbondata.core.util.CarbonTimeStatisticsFactory;
 import org.apache.carbondata.core.util.CarbonUtil;
-import org.apache.carbondata.core.util.CarbonUtilException;
 import org.apache.carbondata.processing.datatypes.GenericDataType;
 import org.apache.carbondata.processing.newflow.AbstractDataLoadProcessorStep;
 import org.apache.carbondata.processing.newflow.CarbonDataLoadConfiguration;
@@ -50,9 +49,9 @@ import org.apache.carbondata.processing.newflow.exception.CarbonDataLoadingExcep
 import org.apache.carbondata.processing.newflow.row.CarbonRow;
 import org.apache.carbondata.processing.newflow.row.CarbonRowBatch;
 import org.apache.carbondata.processing.store.CarbonDataFileAttributes;
-import org.apache.carbondata.processing.store.CarbonFactDataHandlerColumnar;
 import org.apache.carbondata.processing.store.CarbonFactDataHandlerModel;
 import org.apache.carbondata.processing.store.CarbonFactHandler;
+import org.apache.carbondata.processing.store.CarbonFactHandlerFactory;
 import org.apache.carbondata.processing.store.writer.exception.CarbonDataWriterException;
 import org.apache.carbondata.processing.util.CarbonDataProcessorUtil;
 
@@ -199,7 +198,8 @@ public class DataWriterProcessorStepImpl extends AbstractDataLoadProcessorStep {
     } else {
       carbonFactDataHandlerModel.setMdKeyIndex(measureCount);
     }
-    dataHandler = new CarbonFactDataHandlerColumnar(carbonFactDataHandlerModel);
+    dataHandler = CarbonFactHandlerFactory.createCarbonFactHandler(carbonFactDataHandlerModel,
+        CarbonFactHandlerFactory.FactHandlerType.COLUMNAR);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
```diff
@@ -0,0 +1,374 @@
...
+
+  @Override public void intialize() throws CarbonDataLoadingException {
+    setStepConfiguration();
+  }
+
+  /**
+   * This method will be used to get and update the step properties which will
+   * required to run this step
+   *
+   * @throws CarbonUtilException
```
jackylk: wrong description

jackylk:
Actually I think we should have an BlockletEncoder in this step to expose developer API to make it easier to extend encode method. Since this is another work, we can merge this PR first and refactory on it later.

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
```diff
@@ -0,0 +1,374 @@
...
+import org.apache.carbondata.processing.store.CarbonFactDataHandlerModel;
+import org.apache.carbondata.processing.store.CarbonFactHandler;
+import org.apache.carbondata.processing.store.writer.exception.CarbonDataWriterException;
+import org.apache.carbondata.processing.util.CarbonDataProcessorUtil;
+
+/**
+ * It reads data from sorted files which are generated in previous sort step.
+ * And it writes data to carbondata file. It also generates mdk key while writing to carbondata file
+ */
+public class DataWriterProcessorStepImpl extends AbstractDataLoadProcessorStep {
```
chenliang613: Why need to add "ProcessorStep" words to make the class name too long ?
How about directly use this class name: CarobnDataWriter ?

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
```diff
@@ -0,0 +1,360 @@
...
+        dataHandler.addDataToStore(outputRow);
+        writeCounter++;
+      }
+    } catch (Exception e) {
+      throw new CarbonDataLoadingException("unable to generate the mdkey", e);
+    }
+  }
+
+  @Override protected CarbonRow processRow(CarbonRow row) {
+    return null;
```
jackylk: if this is returning null, in `AbstractDataLoadProcessorStep.processRowBatch` it is unnecessarily adding many null

jackylk:
Can you separate out the creation of `dataHandler` into a Factory class,  and use that factory in `DataWriterProcessorStepImpl`. The logic will be cleaner

jackylk:
Suggest to modify the implementation name to `PipelinedWriterProcessorStepImpl` since it is assuming last step is a merge sort step and using producer-consumer to write in a pipelined manner.
In future, if sort step implements external sort (in memory sort), we could add a parallel write implementation for this step.

ravipesala:
@jackylk  This data writer step is not assuming merge sort with producer-consumer. Even we did not integrate producer-consumer for merge sort in our code. This step can receive sorted data from any type of step(even for in memory sort step.)
The producer-consumer present in CarbonFactHandler implementation is for parallel processing the block after it receives blocklet size of rows.

ravipesala:
@jackylk Added factory datahandler. Please review

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
```diff
@@ -0,0 +1,360 @@
...
+      simpleDimsCount++;
+      List<GenericDataType> primitiveTypes = new ArrayList<GenericDataType>();
+      complexDataType.getValue().getAllPrimitiveChildren(primitiveTypes);
+      for (GenericDataType eachPrimitive : primitiveTypes) {
+        eachPrimitive.setSurrogateIndex(surrIndex++);
+      }
+    }
+  }
+
+  private void initDataHandler() {
```
jackylk: can we move all these initialization logic into CarbonFactHandlerFactory? I think the handlerModel is created from configuration, so passing configuration to the factory can create both model and the handler.

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
```diff
@@ -0,0 +1,360 @@
...
+    carbonFactDataHandlerModel.setPrimitiveDimLens(simpleDimsLen);
+    carbonFactDataHandlerModel.setCarbonDataFileAttributes(carbonDataFileAttributes);
+    carbonFactDataHandlerModel.setCarbonDataDirectoryPath(carbonDataDirectoryPath);
+    carbonFactDataHandlerModel.setIsUseInvertedIndex(isUseInvertedIndex);
+    if (noDictionaryCount > 0 || complexDimensionCount > 0) {
+      carbonFactDataHandlerModel.setMdKeyIndex(measureCount + 1);
+    } else {
+      carbonFactDataHandlerModel.setMdKeyIndex(measureCount);
+    }
+    dataHandler = CarbonFactHandlerFactory.createCarbonFactHandler(carbonFactDataHandlerModel,
```
jackylk: I think we should make it work then refactory the `CarbonFactHandler` later. We should decouple the producer-consumer and actual writing logic. There should be abstraction like `CarbonFileWriter` and `CarbonBlockletWriter`, and it can be used to write single carbon data file by java application.

processing/src/main/java/org/apache/carbondata/processing/store/CarbonFactHandlerFactory.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public final class CarbonFactHandlerFactory {
+
+  /**
+   * Creating fact handler to write data.
+   * @param model
+   * @param handlerType
+   * @return
+   */
+  public static CarbonFactHandler createCarbonFactHandler(CarbonFactDataHandlerModel model,
```
ravipesala: Yes, I don't see the advantage of using semaphore here because we are already using fixed thread pool to control the threads. I will discuss with team and confirm whether it is needed.

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
ravipesala: Ok, I will move handlerModel creation to handlerModel itself to keep it simple.

processing/src/main/java/org/apache/carbondata/processing/util/CarbonDataProcessorUtil.java
```diff
@@ -304,4 +311,92 @@ public static String getLocalDataFolderLocation(String databaseName, String tabl
...
+  }
+
+  /**
+   * This method will return all the child dimensions under complex dimension
+   *
+   */
+  private static void addAllComplexTypeChildren(CarbonDimension dimension, StringBuilder dimString,
+      String parent) {
+    dimString.append(
+        dimension.getColName() + CarbonCommonConstants.COLON_SPC_CHARACTER + dimension.getDataType()
```
ravipesala: ok

processing/src/main/java/org/apache/carbondata/processing/newflow/steps/writer/DataWriterProcessorStepImpl.java
ravipesala: ok

jackylk:
There is one compilation error in `CarbonDataProcessorUtil` because of other merged PR

ravipesala:
But there is no compilation issue for me.
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/517/

jackylk:
I rebase locally then do the local mvn clean verify, it does has compilation error. I guess CI is not doing rebase before testing?

ravipesala:
Ok. Fixed, Please review and merge.

PR #252, opened 2016-10-20
[CARBONDATA-331] Modify Compressor interface and add no compression option

jackylk:
Modification in this PR:
1. Modify the compressor inteface to make it easier to add new compressors.
2. And add a DummyCompressor for not doing compression, user can choose the compressor by setting `CarbonProperty`, for example:

```
CarbonProperties.getInstance().addProperty(CarbonCommonConstants.COMPRESSOR, "snappy")
CarbonProperties.getInstance().addProperty(CarbonCommonConstants.COMPRESSOR, "dummy")
```


jackylk:
This PR have two benefit:
1. Modify the compressor inteface to make it easier to add new compressors, like adding no compression option, and other compressor like lz4, gzip, etc.
2. If the query is CPU bound but not IO bound, using no compression will be faster because decompress time is saved.

ravipesala:

jackylk:
This PR is invalid now as the interface got changed already.

PR #253, opened 2016-10-21
trivial fix in documentation

phalodi:

This is too trivial so i didn't create JIRA task.
In Quick Start Guide there is use of string interpolation to create the file but while we are using it in SQL without using 's' before the string to apply string interpolation.

## How was this patch tested?
Manually tested it by run the string manipulation on carbon-spark-shell


Zhangshunyu:
LGTM

chenliang613:
LGTM

PR #255, opened 2016-10-22
CARBONDATA-336 Align the the name description

chenliang613:
Align the the name description for databaseNameOp and tableName.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -926,14 +926,14 @@ class CarbonSqlParser()
       (INTO ~> TABLE ~> (ident <~ ".").? ~ ident) ~
       (OPTIONS ~> "(" ~> repsep(loadOptions, ",") <~ ")").? <~ opt(";") ^^ {
         case filePath ~ isOverwrite ~ table ~ partionDataOptions =>
-          val (schema, tablename) = table match {
+          val (databaseNameOp, tableName) = table match {
             case databaseName ~ tableName => (databaseName, tableName.toLowerCase())
           }
           if(partionDataOptions.isDefined) {
             validateOptions(partionDataOptions)
           }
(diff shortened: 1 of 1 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -926,14 +926,14 @@ class CarbonSqlParser()
       (INTO ~> TABLE ~> (ident <~ ".").? ~ ident) ~
       (OPTIONS ~> "(" ~> repsep(loadOptions, ",") <~ ")").? <~ opt(";") ^^ {
         case filePath ~ isOverwrite ~ table ~ partionDataOptions =>
-          val (schema, tablename) = table match {
+          val (databaseNameOp, tableName) = table match {
```
jackylk: change to databaseName

chenliang613: 1. Why add "Op", because this is option[], so all places using the below : 
case class LoadTable(
    databaseNameOp: Option[String],
    tableName: String,
2.Using options.dbName to get databaseName.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -926,14 +926,14 @@ class CarbonSqlParser()
...
             case databaseName ~ tableName => (databaseName, tableName.toLowerCase())
           }
           if(partionDataOptions.isDefined) {
             validateOptions(partionDataOptions)
           }
           val patitionOptionsMap = partionDataOptions.getOrElse(List.empty[(String, String)]).toMap
-          LoadTable(schema, tablename, filePath, Seq(), patitionOptionsMap, isOverwrite.isDefined)
+          LoadTable(databaseNameOp, tableName, filePath, Seq(), patitionOptionsMap, isOverwrite.isDefined)
```
jackylk: exceed 100 characters

PR #262, opened 2016-10-27
[CARBONDATA-308] Use CarbonInputFormat in CarbonScanRDD compute

jackylk:
Use CarbonInputFormat in CarbonScanRDD compute function
1. In driver side, only getSplit is required, so only filter condition is required, no need to create full QueryModel object, so creation of QueryModel is moved from driver side to executor side.
2. use CarbonInputFormat.createRecordReader in CarbonScanRDD.compute instead of use 
   QueryExecutor directly
3. A new input split class called CarbonMultiBlockSplit is added, CarbonInputFormat uses it to scan multiple block in one task.
4. Distribution of blocklet is removed in CarbonScanRdd.getPartition.

After these changes, CarbonInputFormat can be used in both driver side and executor side. Since no scanning logic is changed, there should not be changes in term of performance.


jackylk:
CI passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/537/
@QiangCai @kumarvishal09 @ravipesala please review

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputSplit.java
```diff
@@ -22,28 +22,44 @@
...
    */
   private int numberOfBlocklets = 0;
 
-  public CarbonInputSplit() {
-    super(null, 0, 0, new String[0]);
+  public  CarbonInputSplit() {
   }
 
-  public CarbonInputSplit(String segmentId, Path path, long start, long length,
+  private void parserPath(Path path) {
```
QiangCai: please use CarbonTablePath.DataFileUtil.getTaskNo

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputSplit.java
```diff
@@ -22,28 +22,44 @@
...
+    String[] nameParts = path.getName().split("-");
+    if (nameParts != null && nameParts.length >= 3) {
+      this.taskId = nameParts[2];
+    }
+  }
+
+  private CarbonInputSplit(String segmentId, Path path, long start, long length,
```
QiangCai: please initialize taskId

jackylk:
@QiangCai fixed

core/src/main/java/org/apache/carbondata/scan/filter/FilterExpressionProcessor.java
```diff
@@ -352,4 +354,18 @@ private FilterResolverIntf getFilterResolverBasedOnExpressionType(
     return new RowLevelFilterResolverImpl(expression, false, false, tableIdentifier);
   }
 
+  public static FilterResolverIntf getResolvedFilter(AbsoluteTableIdentifier identifier,
```
ravipesala: Why it was added?

integration/spark/src/main/java/org/apache/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -953,66 +959,6 @@ public static void checkAndCreateCarbonDataLocation(String carbonStorePath, Stri
   }
 
   /**
-   * method to distribute the blocklets of a block in multiple blocks
```
ravipesala: May be we should take a call for removing blocklet distribution. For filter queries with small number of blocks to scan it is very helpful to process faster.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -1,311 +0,0 @@
-/*
```
ravipesala: Better we just move the class to desired package and update it instead of removing the class and adding it in new package. It will maintain history.

integration/spark/src/main/java/org/apache/carbondata/spark/load/CarbonLoaderUtil.java
jackylk: @kumarvishal09 can you have a look and probably verify any performance change for filter queries?

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
```diff
@@ -177,6 +178,13 @@ public ValidAndInvalidSegmentsInfo getValidAndInvalidSegments() throws IOExcepti
       }
 
     }
+
+    // remove entry in the segment index if there are invalid segments
+    if (listOfInvalidSegments.size() > 0) {
```
kumarvishal09: segment status manager responsibility is to give list of valid and invalid segment, removing invalid segments from SegmentTaskIndexStore in SegmentStatusManager is not a good idea, I think class which is loading blocks is a better place to add this code.

core/src/main/java/org/apache/carbondata/core/carbon/datastore/block/Distributable.java
```diff
@@ -16,10 +16,12 @@
...
+
 /**
- * Abstract class which is maintains the locations of node.
+ * interface to get the locations of node. Used for making task distribution based on locality
  */
-public abstract class Distributable implements Comparable<Distributable> {
+public interface Distributable extends Comparable<Distributable> {
 
-  public abstract String[] getLocations();
+  String[] getLocations() throws IOException;
```
kumarvishal09: Any reason to throw IOException form this method, I think this is not required ??

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -311,80 +278,6 @@ private void addSegmentsIfEmpty(JobContext job, AbsoluteTableIdentifier absolute
     return result;
   }
 
-  /**
-   * get total number of rows. Same as count(*)
-   *
-   * @throws IOException
-   * @throws IndexBuilderException
-   */
-  public long getRowCount(JobContext job) throws IOException, IndexBuilderException {
```
kumarvishal09: This method is useful for count(*) query as we can return number of rows from driver itself , currently we are pushing down to executor, better keep this method it will be useful.

core/src/main/java/org/apache/carbondata/core/carbon/datastore/block/Distributable.java
jackylk: Because CarbonInputSplit need to implement Distributable, and InputSplit has a getLocation function that throws IOException

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
jackylk: ok

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
jackylk: ok, modified

jackylk: ok, modified

jackylk:
CI passed 
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/606/

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -130,41 +130,27 @@ public static CarbonTable getCarbonTable(Configuration configuration) throws IOE
...
   /**
-   * It sets the resolved filter expression
+   * It sets unresolved filter expression.
    *
    * @param configuration
    * @param filterExpression
    */
-  public static void setFilterPredicates(Configuration configuration,
-      FilterResolverIntf filterExpression) {
+  public static void setFilterPredicates(Configuration configuration, Expression filterExpression) {
```
ravipesala: Can't the `filterExpression` null in any case? Don't require null check?

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -244,68 +195,40 @@ class CarbonScanRDD[V: ClassTag](
...
 
-    iter
+  private def createInputFormat(conf: Configuration): CarbonInputFormat[V] = {
+    val format = new CarbonInputFormat[V]
+    CarbonInputFormat.setTablePath(conf, identifier.getTablePath)
+    CarbonInputFormat.setFilterPredicates(conf, filterExpression)
+    val projection = new CarbonProjection
+    columnProjection.foreach { attr =>
+      projection.addColumn(attr.asInstanceOf[AttributeReference].name)
+    }
```
ravipesala: I think we can simply write as below
```
columnProjection.foreach(attr => projection.addColumn(attr.name))
```

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
```diff
@@ -97,15 +91,14 @@ public long getTableStatusLastModifiedTime() throws IOException {
    * @return
    * @throws IOException
    */
-  public ValidAndInvalidSegmentsInfo getValidAndInvalidSegments() throws IOException {
+  public static ValidSegmentsInfo getValidSegments(AbsoluteTableIdentifier identifier)
```
ravipesala: In old code we are getting both valid and invalid segments at once time by reading status file one time. Now we need to read twice for getting the same. Is there any reason of changing the old code?

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -224,42 +221,29 @@ private void addSegmentsIfEmpty(JobContext job, AbsoluteTableIdentifier absolute
    * @return List<InputSplit> list of CarbonInputSplit
    * @throws IOException
    */
-  @Override public List<InputSplit> getSplits(JobContext job) throws IOException {
-    try {
+  @Override
+  public List<InputSplit> getSplits(JobContext job) throws IOException {
+    AbsoluteTableIdentifier identifier = getAbsoluteTableIdentifier(job.getConfiguration());
+    addSegmentsIfEmpty(job, identifier);
+    if (getSegmentsToAccess(job).length == 0) {
```
ravipesala: Is this check required here? Any way in the down method it does again right?

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -61,7 +61,6 @@
 import org.apache.carbondata.lcm.status.SegmentStatusManager;
```
ravipesala: I guess `getRowCount` can be removed as we are not using it.

integration/spark/src/main/scala/org/apache/spark/mapreduce/CarbonHadoopMapReduceUtil.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.mapreduce
+
+/**
+ * Utility that enable the usage of SparkHadoopMapReduceUtil out side of spark package
+ */
+trait CarbonHadoopMapReduceUtil extends SparkHadoopMapReduceUtil {
```
ravipesala: This class is already added in latest master

jackylk: ok

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
jackylk: ok

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -224,42 +221,29 @@ private void addSegmentsIfEmpty(JobContext job, AbsoluteTableIdentifier absolute
...
+  public List<InputSplit> getSplits(JobContext job) throws IOException {
+    AbsoluteTableIdentifier identifier = getAbsoluteTableIdentifier(job.getConfiguration());
+    addSegmentsIfEmpty(job, identifier);
+    if (getSegmentsToAccess(job).length == 0) {
+      return new ArrayList<>(0);
+    }
+
+    FilterResolverIntf filterInterface;
+    Expression filter = getFilterPredicates(job.getConfiguration());
+    if (filter == null) {
```
jackylk: ok

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
jackylk: If this check is removed, some test case of MergeRDD is failing

processing/src/main/java/org/apache/carbondata/lcm/status/SegmentStatusManager.java
jackylk: ok, added back.
invalid segments need to pass to task side, executor need to remove them in the index store

hadoop/src/main/java/org/apache/carbondata/hadoop/CarbonInputFormat.java
jackylk: ok, added

jackylk:
CI testing
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/609/

jackylk:
CI running
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/659/

jackylk:
CI running
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/661/

PR #267, opened 2016-10-28
 [CARBONDATA-340] implement test cases for core.load module

anuragknoldus:
Be sure to do all of the following to help us incorporate your contribution
quickly and easily:
- [x] Make sure the PR title is formatted like:
  `[CARBONDATA-<Jira issue #>] Description of pull request`
- [x] Make sure tests pass via `mvn clean verify`. (Even better, enable
     Travis-CI on your fork and ensure the whole test matrix passes).
- [x] Replace `<Jira issue #>` in the title with the actual Jira issue
     number, if there is one.
- [ ] If this contribution is large, please file an Apache
     [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.txt).
- [x] Testing done
  
  ```
  Please provide details on 
  - What manual testing you have done?
  ```

---

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/load/LoadMetadataDetails.java
+++ b/core/src/main/java/org/apache/carbondata/core/load/LoadMetadataDetails.java
@@ -150,7 +150,7 @@ public class LoadMetadataDetails implements Serializable {
    * return loadStartTime
    * @return
    */
-  public long getLoadStartTimeAsLong() {
+  public Long getLoadStartTimeAsLong() {
     return getTimeStamp(loadStartTime);
   }
 
--- a/core/src/test/java/org/apache/carbondata/core/load/LoadMetadataDetailsUnitTest.java
+++ b/core/src/test/java/org/apache/carbondata/core/load/LoadMetadataDetailsUnitTest.java
@@ -22,6 +22,10 @@ package org.apache.carbondata.core.load;
 import org.junit.Before;
 import org.junit.Test;
 
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertNotSame;
+import static junit.framework.Assert.assertNull;
+
 public class LoadMetadataDetailsUnitTest {
 
   private LoadMetadataDetails loadMetadataDetails;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/carbondata/core/load/LoadMetadataDetails.java
```diff
@@ -150,7 +150,7 @@ public String getLoadStartTime() {
    * return loadStartTime
    * @return
    */
-  public long getLoadStartTimeAsLong() {
+  public Long getLoadStartTimeAsLong() {
```
ravipesala: Why it is required to change to `Long`

ravipesala:
LGTM

core/src/test/java/org/apache/carbondata/core/load/LoadMetadataDetailsUnitTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+package org.apache.carbondata.core.load;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class LoadMetadataDetailsUnitTest {
+
+
+    private LoadMetadataDetails loadMetadataDetails;
```
jackylk: wrong indentation

anuragknoldus:
OK @ravipesala I will format the class using formatter.

core/src/test/java/org/apache/carbondata/core/load/LoadMetadataDetailsUnitTest.java
anuragknoldus: Sorry, I will manage indention.

anuragknoldus:
@ravipesala and @jackylk, I have made required changes. Could you please look into the changes.

core/src/test/java/org/apache/carbondata/core/load/LoadMetadataDetailsUnitTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * This method will test Hashcode which will return 31 if we don't set loadName.
+   *
+   * @throws Exception
+   */
+
+  @Test public void testHashCodeLoadNameNull() throws Exception {
+    Integer data = loadMetadataDetails.hashCode();
+    assert (data == 31);
```
jackylk: better to use assertEquals, so that log will give the actual value in case of failure

anuragknoldus: OK @jackylk.

anuragknoldus:
I have made the required changes. @jackylk, Could you please look into the changes.

jackylk:
LoadMetadataDetailsUnitTest.testGetTimeStampWithDate failed
expected:<1451586600000000> but was:<1451577600000000>

anuragknoldus:
@jackylk, I am not getting any error. I have run **mvn clean verify** before generatng the PR and  project build successfully.
I think this is Epoch and its taking your time zone while parsing the date when test cases run.

jackylk:
Did you rebase before testing? I have rebased before testing.
And if time zone is the problem, please make the test case not depend on time zone, mvn verify should  run globally.

anuragknoldus:
Yes @jackylk , I always rebase before implementing test case as well as before generate PR and run **mvn clean verify** to execute all the test cases.
Yes, problem was related to time zone. I have checked both value on Epoch convertter and get the time difference and that was my mistake to use static value in expected result.
Now I have made a function in test file which will work same as the original method and than we compare both value so it will not be time zone dependent.

PR #273, opened 2016-10-31
[CARBONDATA-332]Prohibit to use reserved words in database\table\col name

lion-x:
# Why raise this PR?
Carbon should prohibit to use reserved words in database\table\col name.  Using reserved words will cause many unnecessary troubles.

# How to solve?
Adding  reserved words validation for database, table, cols when they are creating.

# How to test?
Pass all test cases. 

# CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/543/


ravipesala:

PR #280, opened 2016-11-02
[CARBONDATA-363]fixed block loading issue in case of blocklet distribution

kumarvishal09:
**Problem**: In case of blocklet distribution same block is getting loaded multiple times this is because when blocklet distribution is enabled same block will divided inside a task so there is not synchronisation as block loading is done in different thread because of this same block is getting read from carbon data file footer multiple times and it is hitting the first time query performance.
**Solution**: Need to add locking for above issue, if one thread is loading particulate block other thread need and once block loaded other thread need to use same reference of the block


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/558/

ravipesala:
LGTM

PR #282, opened 2016-11-03
[CARBONDATA-364]Fixed drop and load table concurrent issue

ashokblend:
**Problem**
When load and drop command being run parallely or concurrently. Load table command will acquire lock on table so that further modification on table will be locked.
Since load in table is in progess, when user runs drop table command, we check if table is locked do not perform any operation and send message as table is locked for updation. 
But in finally block we are not checking if lock is acquired or not, because of which system is deleting all files from store.

**Solution**
Before deleting data from store in finally block, check if lock is acquired. 
If lock is acquired then only proceed to delete table related data.


ashokblend:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/561/

ravipesala:
LGTM

PR #283, opened 2016-11-03
[CARBONDATA-365]Pass block size to model during compaction.

ashokblend:
**Problem**
When user creates table by passing configured block size. After multiple load, compaction is triggered but it fails. After analysis, it is found that configured block size is not passed to processing layer.

**Solution**
Pass configured block size in model to next step in processing layer.

CI is passed for this PR
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/562/


ravipesala:
LGTM

PR #284, opened 2016-11-03
[CARBONDATA-366]Throw error when stream is not closed during load

ashokblend:
**Problem**
In some scenario, when user loads data, it shows successfull but while querying it gives error. After analysing, we saw carbon index file is of 0 byte. During log analysis, it is found that there was error while closing stream.

**Solution**
We should throw the error if occured while closing stream, so that either executer can relaunch the task or propogate the error to user. Same has been handled in this pull request.

CI is passed for this PR
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/563/

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
@@ -117,17 +117,26 @@ public final class CarbonUtil {
     // Added if to avoid NullPointerException in case one stream is being passed as null
     if (null != streams) {
       for (Closeable stream : streams) {
-        if (null != stream) {
-          try {
-            stream.close();
-          } catch (IOException e) {
-            LOGGER.error("Error while closing stream" + stream);
-          }
+        try {
+          closeStream(stream);
+        } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
```diff
@@ -117,18 +117,27 @@ public static void closeStreams(Closeable... streams) {
...
-        if (null != stream) {
-          try {
-            stream.close();
-          } catch (IOException e) {
-            LOGGER.error("Error while closing stream" + stream);
-          }
+        try {
+          closeStream(stream);
+        } catch (IOException e) {
+          LOGGER.error("Error while closing stream:" + e);
```
ravipesala: But here you are not throwing any error. I don't see much difference from old and new code.

ashokblend: This is an optimizaion. This method calls another method which tthrows IOException. During load we call the method which throws IO. Please check my other changes.

ravipesala:
LGTM

PR #292, opened 2016-11-03
[CARBONDATA-375] Dictionary cache not getting cleared after task completion in dictionary decoder

manishgupta88:
Problem: Dictionary cache not getting cleared after task completion in dictionary decoder

Analysis: Currently LRU cache eviction policy is based on dictionary access count. For cache to remove a entry its access count must be 0. In dictionary decoder after conversion of surrogate key to actual value the access count for dictionary columns in query is not getting decremented due to which it will never be cleared from memory when LRU cache size is configured.

Fix: Add a task completion listener which will take care of clearing the dictionary in case of both success and failure

Impact area: LRU cache eviction policy which can lead to query and data load failure

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonDictionaryDecoder.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonDictionaryDecoder.scala
@@ -17,6 +17,7 @@
 
 package org.apache.spark.sql
 
+import org.apache.spark.TaskContext
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.errors.attachTree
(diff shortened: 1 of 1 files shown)


manishgupta88:
CI PAss: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/569/

kumarvishal09:
@manishgupta88 I think same scenario we need to handle for compaction as query flow and compaction flow is same, can u please check and add the code if not handled.

manishgupta88:
@kumarvishal09 ....This change is for dictionary decoder that will convert the final result from surrogate to actual values. Compaction will not hit this flow. In compaction clearing dictionary is already handled where we create the query model and clean the dictionary cache after query result processing is complete.

gvramana:
LGTM

kumarvishal09:
Same flow I meant loading the dictionary values which is not required
during compaction, and handling cleaning of dictionary cache scenario

PR #294, opened 2016-11-04
[CARBONDATA-381] Unnecessary catalog metadata refresh and array index of bound exception in drop table

manishgupta88:
Problem:
1. Whenever a catalog metadata is refreshed it modified the timestamp of modifiedTime.mdt file which leads to unnecessary refreshing the complete catalog metadata.
2. Array Index of bound exception is thrown on failure of table creation.

Analysis:
1. Whenever carbon environment gets initialized it loads the table metadata in the catalog and changes the timestamp of modifiedTime.mdt file. If a parallel beeline session is in progress then it will cause unnecessary refreshing of the catalog metadata.
2. For the very first time if table creation fails then in the exception block it tries to drop that table and clear its metadata. In drop table filter API is used which throws array index out of bound exception if metadata array is empty.

Fix:
1. No need to update the timestamp of modifiedTime.mdt file while loading metadata. It should only be refreshed on create and drop table operations.
2. Instead of filter API use find API which will return an Option object.

Impact: carbon catalog refresh which will impact query and load flow.


manishgupta88:
CI Pass: http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/573/

gvramana:
LGTM

PR #299, opened 2016-11-07
[CARBONDATA-384][WIP]Add Table Properties Options Validation

lion-x:
# Why raise this PR?
Now, Carbon does not validate the table properties options. This will cause below problem.
For example,
Create table carbontable (.......)
TBLPROPERTIES ('DICTIONARY_EXELUDE'='colname');
user wants to use DICTIONARY_EXCLUDE property, but he types a wrong option name, the setting will be ignored with no error throw.

# How to test?
Pass all test case.


integration/spark/src/main/scala/org/apache/carbondata/spark/util/CommonUtil.scala
```diff
@@ -122,14 +122,29 @@ object CommonUtil {
...
           isValid = false
           throw new MalformedCarbonCommandException(s"Invalid table properties ${ key }")
         }
     }
     isValid
   }
 
+  def validateTblOption(option: String): Boolean = {
+    val supportedOptions = Seq[String]("DICTIONARY_EXCLUDE", "DICTIONARY_INCLUDE", "COLUMN_GROUPS",
+      "TABLE_BLOCKSIZE", "NO_INVERTED_INDEX", "PARTITIONCLASS", "PARTITIONCOUNT",
```
jackylk: By this way, it is easy to forget to update this list if later more options are added. Better to refactory the options, like make it a Enum, and use it in both parser and here

ravipesala:

PR #301, opened 2016-11-07
[CARBONDATA-385]Select query in case of join is throwing Classcast exception

ashokblend:
**Problem**
This problem can be better understood by example.  
Let's say we have two table
1. employee(eid,mgrid,city)
2. manager(eid,mgrid,no)
query is 
select e.empid from employee e inner join manager m on e.mgrid=m.empid
In this case user did select on e.empid and join condition on e.mgrid and m.empid(manager table)
As per our decoder plan, for relation e mgrid and for relation m empid will get decoded first and hence
it will be included in Exclude profile of Decoder. So exclude profile of decoder will have
mgrid#<expid of relation e>,empid#<expid of relation m>. 
In our code we are comparing only name and hence decoder will assume empid, which user has selected, is already decoded and hence it will not try to decode.

**Solution**
Change is made to compare both name and attribute expid.

CI pass link
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/588/


ravipesala:
LGTM

PR #304, opened 2016-11-08
[WIP]Fixed issue with coalasce function.

ashokblend:
**Problem**
Added testcase for query select * from table where coalasce(attr,1)!='value', this should result null record also


ashokblend:
Closing this, as its not required here.

PR #306, opened 2016-11-09
Update readme as per IPMC comments

chenliang613:
Update readme as per IPMC comments:
1.Adds incubating.
2.Removes the original contributed company.


ravipesala:
LGTM

PR #308, opened 2016-11-09
[CARBONDATA-398] In DropCarbonTable flow, Metadata lock should be acquired only if tab…

nareshpr:
Issue : In drop table flow, we acquiring metadata lock even if table not exist which is creating table folder and creating meta.lock file.
Solution : Check and acquire lock only if carbon table exist.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
@@ -1284,10 +1284,10 @@ private[sql] case class DropTableCommand(ifExistsSet: Boolean, databaseNameOp: O
       .getCarbonLockObj(carbonTableIdentifier, LockUsage.DROP_TABLE_LOCK)
     val storePath = CarbonEnv.getInstance(sqlContext).carbonCatalog.storePath
     var isLocked = false
-    val tmpTable = org.apache.carbondata.core.carbon.metadata.CarbonMetadata.getInstance()
-      .getCarbonTable(dbName + '_' + tableName)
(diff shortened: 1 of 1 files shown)


nareshpr:
CI PASS - http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/598/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1284,14 +1284,18 @@ private[sql] case class DropTableCommand(ifExistsSet: Boolean, databaseNameOp: O
...
+      .getCarbonTable(dbName + '_' + tableName)
     try {
-      isLocked = carbonLock.lockWithRetries()
-      if (isLocked) {
-        logInfo("Successfully able to get the lock for drop.")
-      }
-      else {
-        LOGGER.audit(s"Dropping table $dbName.$tableName failed as the Table is locked")
-        sys.error("Table is locked for deletion. Please try after some time")
+      if (null != tmpTable) {
```
ravipesala: It is better to use `CarbonEnv.getInstance(sqlContext).carbonCatalog.tableExists`

nareshpr:
Fixed review comments.
CI Pass - http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/599/

ravipesala:
@nareshpr Actually there is already one check exists in CarbonStrategies 
```
if (CarbonEnv.getInstance(sqlContext).carbonCatalog
            .isTablePathExists(toTableIdentifier(tableName.toLowerCase))(sqlContext))
```
why this check is not working? And why we need to add extra check in this PR?

PR #311, opened 2016-11-10
[CARBONDATA-403] Add example for data load without using kettle

jackylk:
In this PR, example SQL and dataframe usage is added for loading data without kettle

### DIFF
--- a/examples/src/main/scala/org/apache/carbondata/examples/CarbonExample.scala
+++ b/examples/src/main/scala/org/apache/carbondata/examples/CarbonExample.scala
@@ -32,6 +32,7 @@ object CarbonExample {
 
     cc.sql("DROP TABLE IF EXISTS t3")
 
+    // create table, 6 dimensions, 1 measure
     cc.sql("""
            CREATE TABLE IF NOT EXISTS t3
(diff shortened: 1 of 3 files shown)


examples/src/main/scala/org/apache/carbondata/examples/CarbonExample.scala
```diff
@@ -32,24 +32,34 @@ object CarbonExample {
...
            name String, phonetype String, serialname String, salary Int)
            STORED BY 'carbondata'
            """)
 
+    // load data using kettle
     cc.sql(s"""
            LOAD DATA LOCAL INPATH '$testData' into table t3
            """)
 
+    // load data without kettle
```
chenliang613: suggest either create a different example for without kettle or do the different query respectively for two type of load data.

lion-x: I quite agree with @chenliang613 . That is a better way.

jackylk:
CI passed:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/610/

chenliang613:
LGTM

PR #319, opened 2016-11-16
[CARBONDATA-411] Test

Zhangshunyu:
test


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -233,7 +233,7 @@ class CarbonSqlParser()
         node match {
           case Token("TOK_DROPDATABASE", children) =>
             dbName = BaseSemanticAnalyzer.unescapeIdentifier(children(0).getText)
-            // check whether cascade drop db
+            // check whether cascade drop database
```
Zhangshunyu: Why change this parameter?

PR #321, opened 2016-11-16
[CARBONDATA-414] Access array elements using index than For Loop

PallaviSingh1992:
Replaced the accessing of array elements using for loop with Index based accessing.
                 This has reduced the complexity of the code by removing the for loop traversal and
                 making index based access.

### DIFF
--- a/integration/spark/src/main/java/org/apache/carbondata/spark/merger/CarbonDataMergerUtil.java
+++ b/integration/spark/src/main/java/org/apache/carbondata/spark/merger/CarbonDataMergerUtil.java
@@ -510,19 +510,10 @@ public final class CarbonDataMergerUtil {
 
     int level1Size = 0;
     int level2Size = 0;
-    boolean first = true;
 
-    for(int levelCount : noOfSegmentLevelsCount){
-      if(first){
-        level1Size = levelCount;
-        first = false;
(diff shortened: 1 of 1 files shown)


integration/spark/src/main/java/org/apache/carbondata/spark/merger/CarbonDataMergerUtil.java
```diff
@@ -510,19 +510,10 @@ private static long getSizeOfOneSegmentAcrossPartition(int partitionCount, Strin
...
-      }
-      else{
-        level2Size = levelCount;
-        break;
-        // breaking as we are doing only 2 levels
-      }
 
+    if (noOfSegmentLevelsCount.length != 0) {
+      level1Size = noOfSegmentLevelsCount[0];
+      level2Size = noOfSegmentLevelsCount[1];
```
Hexiaoqiao: this could throw IndexOutOfBoundsException.

Hexiaoqiao:
1.Please rename title following format as "[CARBONDATA-(Jira issue#)] Description of pull request"
2.It is not necessary to include template info in PR description, please remove it and replace by this PR own descriptions.

jackylk:
LGTM

PR #324, opened 2016-11-17
[CARBONDATA-420] Remove unused parameter in config template file

HuangxiaoxiaBonnie:
## Why rasie this pr?
To remove unused parameter in config template file: carbon.max.file.size is removed now.
## How to test?
Not about function.

### DIFF
--- a/conf/carbon.properties.template
+++ b/conf/carbon.properties.template
@@ -52,7 +52,7 @@ carbon.compaction.level.threshold=4,3
 #default size (in MB) for major compaction to be triggered
 carbon.major.compaction.size=1024
 ######## Query Configuration ########
-#Number of cores to be used for loading blocks when execute query
+#Number of cores to be used for loading index into memory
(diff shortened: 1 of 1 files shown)


Zhangshunyu:
LGTM

conf/carbon.properties.template
```diff
@@ -52,7 +52,7 @@ carbon.compaction.level.threshold=4,3
 #default size (in MB) for major compaction to be triggered
 carbon.major.compaction.size=1024
 ######## Query Configuration ########
-#Number of cores to be used while querying
+#Number of cores to be used for loading blocks when execute query
```
jackylk: It is not just for query, compaction also need this parameter. 
I think a better description is "Number of cores to be used for loading index into memory"

HuangxiaoxiaBonnie:
Fixed review comments, please check. @jackylk

jackylk:
LGTM

PR #325, opened 2016-11-17
[CARBONDATA-418]Fixed data loading performance issue

kumarvishal09:
Problem: In CarbonCSVBasedSeqGenStep for each row dimension column ids string converted from string to String array . As split function being called it will create string object for each row and it will impact the data loading performance.
Soluntion: Create a instance variable and in process row method get in column id when first row is passed to this step


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/621/

jackylk:
LGTM
verified with -Pno-kettle also

PR #327, opened 2016-11-17
[CARBONDATA-421]Time Stamp Filter issue with other than yyyy-mm-dd format

kumarvishal09:
Problem: When time format is yyyy/mm/dd other than "-" filter query is not working , As in filter only "-" is allowed user need to give the filter value is "-" but as data loaded in in "/" filter is not working and returning 0 result. 
Soluntion: Problem is in filter we are taking default format but we need to take format used during data loaded while converting the filter values to surrogate key


kumarvishal09:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/622/

jackylk:
LGTM

PR #328, opened 2016-11-17
[CARBONDATA-374] Support smallint

cenyuhai:
Why raise this pull request?
Support smallint.

How to test?
Test with TestLoadDataWithHiveSyntax


Hexiaoqiao:
@cenyuhai 
Do you think we need to keep SMALLINT, and SHORT as well? Maybe it makes confusion since some users have used SHORT, then occur some Incompatible issue after upgrade. 
Please correct me if i am wrong.

cenyuhai:
@Hexiaoqiao There is a doc about what datatype carbon are supported.
https://cwiki.apache.org/confluence/display/CARBONDATA/Carbon+Data+Types

Short is not working because there is a bug in carbon, so it will not be a problem.

PR #337, opened 2016-11-21
[CARBONDATA-428] Removed Redundant Condition Checks

PallaviSingh1992:
Remove the redundant conditional checking of dictionaryCacheAlreadyExists(cacheType) in CacheProvider.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/cache/CacheProvider.java
+++ b/core/src/main/java/org/apache/carbondata/core/cache/CacheProvider.java
@@ -84,14 +84,13 @@ public class CacheProvider {
     //check if cache is null for given cache type, if null create one
     if (!dictionaryCacheAlreadyExists(cacheType)) {
       synchronized (lock) {
-        if (!dictionaryCacheAlreadyExists(cacheType)) {
-          if (null == cacheTypeToLRUCacheMap.get(cacheType)) {
-            createLRULevelCacheInstance(cacheType);
-          }
-          createDictionaryCacheForGivenType(cacheType, carbonStorePath);
+        if (null == cacheTypeToLRUCacheMap.get(cacheType)) {
+          createLRULevelCacheInstance(cacheType);
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/carbondata/core/cache/CacheProvider.java
```diff
@@ -84,14 +84,13 @@ public static CacheProvider getInstance() {
...
       synchronized (lock) {
-        if (!dictionaryCacheAlreadyExists(cacheType)) {
-          if (null == cacheTypeToLRUCacheMap.get(cacheType)) {
-            createLRULevelCacheInstance(cacheType);
-          }
-          createDictionaryCacheForGivenType(cacheType, carbonStorePath);
+        if (null == cacheTypeToLRUCacheMap.get(cacheType)) {
+          createLRULevelCacheInstance(cacheType);
         }
+        createDictionaryCacheForGivenType(cacheType, carbonStorePath);
```
manishgupta88: @PallaviSingh1992 ....dictionaryCacheAlreadyExists will check if for a given cache type (Forward/Reverse), the object has already been created. Now this check is before and inside synchronized block to handle concurrent access (double locking mechanism). If this check is removed inside the synchronized block then it can happen that in concurrent scenarios, multiple threads end up in creating multiple objects for the same cache type and the object in the map keeps getting overridden. 
Therefore I think that this double check is required. Please correct me if I am wrong.

PallaviSingh1992: synchronized blocks synchronized on the same object can only have one thread executing inside them at the same time.

manishgupta88: Yes I agree...but it can happen that multiple threads are waiting on the synchronized block and when one completes the next one enters, then for that thread also dictionary cache object is non existent and it will also create a new object.
May be you can once go through double locking mechanism and try to relate it with that.

PR #352, opened 2016-11-25
[CARBONDATA-448] Solve compilation error for spark2 integration

jackylk:
Solve compilation error in core for spark2, after this PR, master can be compile with -Pspark-2.0
1. Add example/spark and example/spark2 module
2. Make assembly include spark or spark2 jar according to profile
3. modify carbon core to use GenericInternalRow to make to compilable with spark2


jackylk:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/654/

pom.xml
```diff
@@ -291,27 +286,36 @@
       </activation>
       <properties>
         <spark.version>1.5.2</spark.version>
+        <scala.binary.version>2.10</scala.binary.version>
+        <scala.version>2.10.4</scala.version>
       </properties>
       <modules>
         <module>integration/spark</module>
+        <module>examples/spark</module>
```
scwf: should also include spark common module

PR #357, opened 2016-11-26
[CARBONDATA-451]Fix the bug of storePath.substring on windows which lead to query can not run

Zhangshunyu:
## Why rasie this pr?
To fix the bug that when query execute on windows can not run, this is becasue when table path is got on windows, the patch might contains '\\' but current code substring the path not format it.
## How to solve?
Using the formatted table path.
## How to test?
Pass all the test cases.


Zhangshunyu:
@jackylk pls review this pr, thank you!

manishgupta88:
@Zhangshunyu ...LGMT....only thing is you can put some comments with example for better understanding like what can be tablePath and derived formattedTablePath.

jackylk:
LGTM

Zhangshunyu:
Added comments and rebased.

PR #359, opened 2016-11-28
[CARBONDATA-456] Fixing select count(*) slow issue

ravipesala:
In the current master select count(*) query is slower because it try to retrieve all columns out of store and count it. But it is not needed using metadata could count the number of rows. So in driver side we no need to add all columns when projection is null, just don't add any columns and execute to return the empty rows  but count could be calculated from those empty rows.


ravipesala:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/675/

jackylk:
LGTM

PR #365, opened 2016-11-28
[CARBONDATA-463] Extract code to spark-common

jackylk:
Move reusable code from carbon-spark to carbon-spark-common module, for preparation of spark2 integration.
No functionality is modified in this PR


jackylk:
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/689/

scwf:
i think we should merge this into our spark2 branch, otherwise it will leads to large conflicts

jackylk:
@scwf
I have compared two branches and extracted out the common files, when we push spark2 integration code, common files will be excluded. I think in this way the the total effort will be less.

jackylk:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/696/

integration/spark-common/src/main/java/org/apache/carbondata/integration/spark/merger/CarbonCompactionUtil.java
```diff
@@ -38,7 +38,6 @@
 import org.apache.carbondata.core.util.CarbonUtil;
 import org.apache.carbondata.core.util.CarbonUtilException;
```
ravipesala: I guess better remove `integration` from package for all classes

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/NewCarbonDataLoadRDD.scala
```diff
@@ -51,7 +51,6 @@ class NewCarbonDataLoadRDD[K, V](
     sc: SparkContext,
```
jackylk: I think we can try it later. Currently because of newTaskAttemptID and some other function signature is changed in spark2, so need to duplicate code temporarily

ravipesala:
LGTM except few comments.
It seems there is an issue in pom files while using maven default build and using format. I will fix them in new PR.

integration/spark-common/src/main/java/org/apache/carbondata/integration/spark/merger/CarbonCompactionUtil.java
jackylk: fixed

ravipesala:
LGTM

PR #366, opened 2016-11-28
[CARBONDATA-368]Insert into carbon table feature

ashokblend:
**Problem**
If user wants to load data from other table, carbon doesn't support this.

**Solution**
User will be able to load data in carbon table from other table.
Below command can be used to insert data in carbon table through select query.
**insert into <carbon table> <select query>**
<carbon table> : name of carbon table where data has to be loaded
<select query> :  query which results data which has to be loaded into carbon.

Please check **InsertIntoCarbonTableTestCase.scala** for examples.
CI Url
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/703/

**Note**
Since there is change in schema.thrift and hence carbon-format jar has to be build and deployed.


QiangCai:
please rebase

pom.xml
```diff
@@ -93,6 +93,7 @@
 
   <modules>
     <module>common</module>
+	<module>format</module>
```
jackylk: should not add format dependency

ashokblend:
i have removed changes in pom.xml, but before merging we need to deploy carbon-format as there is change in schema.thrift.

jackylk:
please rebase, thanks

ashokblend:
rebase done. please merge it.

jackylk:
LGTM

PR #378, opened 2016-12-01
[CARBONDATA-480] Add file format version enum

jackylk:
Add file format version enum instead of using short value


jackylk:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/730/

ravipesala:
LGTM

PR #392, opened 2016-12-04
change snapshot version number due to new structure changes as per mailing proposal

chenliang613:
A rich set of features are planned to be included into next release, and more importantly there will be external API changes introduced as will integrate with Spark 2.x, remove kettle etc. 

The next version number will use 1.0.0 for distinguishing major API changes, it would be much helpful to reduce maintenance cost through distinguishing the major different version due to major API changes.

Please find the detailed discussion on dev@mailing list : http://apache-carbondata-mailing-list-archive.1130556.n5.nabble.com/CarbonData-propose-major-version-number-increment-for-next-version-to-1-0-0-td3131.html

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>org.apache.carbondata</groupId>
     <artifactId>carbondata-parent</artifactId>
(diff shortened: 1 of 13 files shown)


ravipesala:
LGTM

PR #396, opened 2016-12-05
[CARBONDATA-493]fix bug for insert into select from empty table 

QiangCai:
InsertInto Feature:
If there is no partition in select sub-query dataframe , data loading will throw dataloadingexception

Solution:
The dataframe without partition, no need to do data loading.


jackylk:
CI passed
http://136.243.101.176:8080/job/ApacheCarbonManualPRBuilder/750/

jackylk:
LGTM

PR #400, opened 2016-12-05
[SPARK2] Fix bug in spark2 module

jackylk:
CarbonExample in spark2 is failing because not setting current read support class.


ravipesala:
LGTM

PR #432, opened 2016-12-13
[CARBONDATA-531] remove spark dependency in carbon core

jackylk:
Carbon-core module should not depend on spark, this PR removes this dependency

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonScan.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonScan.scala
@@ -27,6 +27,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.execution.LeafNode
 import org.apache.spark.sql.hive.CarbonMetastore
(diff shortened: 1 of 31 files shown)


CarbonDataQA:
Build Failed  with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/162/

integration/spark/src/main/scala/org/apache/spark/sql/CarbonScan.scala
```diff
@@ -151,8 +153,15 @@ case class CarbonScan(
 
         override def next(): InternalRow = {
           val value = iter.next
+          val converted = value.map {
```
ravipesala: There was performance problem earlier with map if number of scan rows are more, that was the main reason we moved it to carbon layer.

jackylk: understood, I will wait for the performance check in cluster environment before reworking on this

CarbonDataQA:
Build Failed  with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/518/

PR #441, opened 2016-12-18
[CARBONDATA-539] Fiix column projection bug

jackylk:
Currently when using CarbonInputFormat in map reduce app, if projection columns are not set, then  Carbon will return empty row, but the correct behavior should read all columns. This PR fix this bug.

`CarbonInputFormat` are refactored to make cleaner API to application. 
Only 4 settings are now public to application. 
1. TablePath
2. Column Projection
3. Filter Expression
4. ReadSupport class
Function descriptions are added in `CarbonInputFormat`

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/util/CarbonInputFormatUtil.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/util/CarbonInputFormatUtil.java
@@ -45,15 +45,14 @@ import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
  */
 public class CarbonInputFormatUtil {
 
-  public static CarbonQueryPlan createQueryPlan(CarbonTable carbonTable, String columnString) {
+  public static CarbonQueryPlan createQueryPlan(CarbonTable carbonTable, String[] projection) {
     String[] columns = null;
-    if (columnString != null) {
-      columns = columnString.split(",");
+    if (projection != null) {
+      columns = projection;
     }
     String factTableName = carbonTable.getFactTableName();
     CarbonQueryPlan plan = new CarbonQueryPlan(carbonTable.getDatabaseName(), factTableName);
-    // fill dimensions
-    // If columns are null, set all dimensions and measures
+    // fill dimensions, if projection columns are null, set all dimensions and measures
     int i = 0;
     if (columns != null) {
       for (String column : columns) {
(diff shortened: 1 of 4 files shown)


hadoop/src/main/java/org/apache/carbondata/hadoop/util/CarbonInputFormatUtil.java
```diff
@@ -70,6 +69,18 @@ public static CarbonQueryPlan createQueryPlan(CarbonTable carbonTable, String co
           i++;
         }
       }
+    } else {
+      // query all columns
+      List<CarbonMeasure> tableMsrs = carbonTable.getMeasureByTableName(factTableName);
```
jackylk: This is the bug, when input projection is null, earlier it does not add all columns to read

CarbonDataQA:
Build Success with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/218/

hadoop/src/main/java/org/apache/carbondata/hadoop/util/CarbonInputFormatUtil.java
ravipesala: It was an performance issue if all columns are added when projection is null.  That was the reason it was removed.  Any reason it is added back?

jackylk: Because when I run HadoopFileExample, it is returning empty row. Then I found out that it is because projection columns are not set in `CarbonInputFormat`. So I have added it back. If it is removed, then it is mandatory that user set projection columns in `CarbonInputFormat`

ravipesala: yes, But we cannot make it mandatory because in count(*) case it sets null projection so that just returns empty rows but can calculate the count using that.

jackylk: ok, I got the issue here. Can we do like as following
1. if it is come from count(*), we set the projection column to a special string (like empty string "") in configuration, then we can process it differently in `createQueryPlan`. 
2. If user want to project all column, projection column should be set to null or list all columns.
3. If user want to project specific columns, then set it to specific columns

jackylk: ok, I will add a special column name set to projection

jackylk:
I think it is better let user set the column projection explicitly. So closing this PR

PR #459, opened 2016-12-23
[CARBONDATA-558] Fix load performace issue when use_kettle=false

foryou2030:
# Why raise this pr?
When I load a data file, the measure column contains many empty strings, if use_kettle=false, the load performance has a sharp decline
I checked the logs of executor, many warnnings printed like below:
16/12/22 07:03:12 WARN MeasureFieldConverterImpl: pool-22-thread-6 Cant not convert : to Numeric type value. Value considered as null.

# How to solve it?
When measureValue = "", we should set it as null directly, no need to do datatype conversion

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/newflow/converter/impl/MeasureFieldConverterImpl.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/newflow/converter/impl/MeasureFieldConverterImpl.java
@@ -22,6 +22,7 @@ import org.apache.carbondata.common.logging.LogService;
 import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.carbon.metadata.datatype.DataType;
 import org.apache.carbondata.core.carbon.metadata.schema.table.column.CarbonMeasure;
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Success with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/301/

lion-x:
LGTM

eason-lyx:
@foryou2030 please add test case for this scenario.
thanks.

manishgupta88:
@foryou2030 As a suggestion whenever you make any changes in the parsing behavior of data in the load flow without kettle, please validate the same behavior for load using the kettle flow considering that as a base. That would give the clear picture for defects.
Also this practice will make clear what all things need to be merged in the new load flow without kettle to keep the same behavior for load with kettle.

foryou2030:
@manishgupta88 , yes , I  compared the code of  new flow with the kettle flow,  and found the difference. We can check the method populateOutputRow in class "CarbonCSVBasedSeqGenStep" for more details.

PR #482, opened 2016-12-30
[CARBONDATA-552] Catch the FilterUnsupportedException in RowLevelRangeFilterResolverImpl

phalodi:
Catch the FilterUnsupportedException in RowLevelRangeFilterResolverImpl.scala to handle and log the error for invalid filter condition.

Test by running all the test case with `mvn clean verify`


CarbonDataQA:

core/src/main/java/org/apache/carbondata/scan/filter/resolver/RowLevelRangeFilterResolverImpl.java
```diff
@@ -187,7 +185,11 @@ public void resolve(AbsoluteTableIdentifier absoluteTableIdentifier) {
           dimColumnEvaluatorInfo.setDimension(columnExpression.getDimension());
           dimColumnEvaluatorInfo.setDimensionExistsInCurrentSilce(false);
           if (columnExpression.getDimension().hasEncoding(Encoding.DIRECT_DICTIONARY)) {
-            filterInfo.setFilterList(getDirectSurrogateValues(columnExpression));
+            try {
+              filterInfo.setFilterList(getDirectSurrogateValues(columnExpression));
+            } catch (FilterUnsupportedException e) {
+              FilterUtil.logFilterError(e, false);
```
manishgupta88: In the earlier code, if any exception was thrown it was not caught and delegated back to the caller but that behavior will be changed with this modified code. So in the catch block please rethrow the exception so that the behavior remains the same as old one

PR #487, opened 2016-12-30
[CARBONDATA-581] Wrong number of executors requested when preferred locations are not obtained

sraghunandan:
Wrong number of executors requested when preferred locations are not obtained.
In case of groupby in subquery, we'll not get the preferred locations from the RDD as data is shuffled.
In such cases we need to use number of executors available.
This is applicable in case of dynamic executors


CarbonDataQA:

ravipesala:
add to whitelist

CarbonDataQA:
Build Success with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/389/

PR #517, opened 2017-01-10
[CARBONDATA-621]Fixed compaction with multiple blocklet issue

kumarvishal09:
Problem: Compaction is failing in case of multiple blocklet and when each segment dictionary column size is changing
Reason: this is because during compaction we need to update the dictionary byte value with last segment key generator, in carbon merge rdd we are passing the oldest segment cardinality value we need to pass latest segment cardinality
Solution: Pass the latest segment cardinality.

### DIFF
--- a/integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datacompaction/DataCompactionBlockletBoundryTest.scala
+++ b/integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datacompaction/DataCompactionBlockletBoundryTest.scala
@@ -36,7 +36,7 @@ class DataCompactionBlockletBoundryTest extends QueryTest with BeforeAndAfterAll
       .addProperty(CarbonCommonConstants.CARBON_TIMESTAMP_FORMAT, "mm/dd/yyyy")
     CarbonProperties.getInstance()
       .addProperty(CarbonCommonConstants.BLOCKLET_SIZE,
-        "55")
+        "120")
     sql(
       "CREATE TABLE IF NOT EXISTS blocklettest (country String, ID String, date Timestamp, name " +
         "String, " +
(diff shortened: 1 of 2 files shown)


CarbonDataQA:
Build Success with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/541/

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datacompaction/DataCompactionBlockletBoundryTest.scala
```diff
@@ -36,7 +36,7 @@ class DataCompactionBlockletBoundryTest extends QueryTest with BeforeAndAfterAll
       .addProperty(CarbonCommonConstants.CARBON_TIMESTAMP_FORMAT, "mm/dd/yyyy")
     CarbonProperties.getInstance()
       .addProperty(CarbonCommonConstants.BLOCKLET_SIZE,
-        "55")
+        "120")
```
jackylk: Do you want to set the value that make the file containing more than 1 blocklet? I think it is better to ensure it by calculating from the input CSV file.

integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -318,7 +318,7 @@ class CarbonMergerRDD[K, V](
 
     // prepare the details required to extract the segment properties using last segment.
     if (null != carbonInputSplits && carbonInputSplits.nonEmpty) {
-      val carbonInputSplit = carbonInputSplits.last
+      val carbonInputSplit = carbonInputSplits(0)
```
jackylk: use head instead of (0), and why head is the latest one?

kumarvishal09: ok

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datacompaction/DataCompactionBlockletBoundryTest.scala
kumarvishal09: Yes this is for having multiple blocklet in one carbon data file.

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/584/

jackylk:
LGTM

PR #520, opened 2017-01-11
fix dependency issue for IntelliJ IDEA

QiangCai:
When using profile spark-2.1, can not run test case of spark-common-test in IntelliJ IDEA.


QiangCai:
close this pr. I didn't reproduce this issue.

PR #524, opened 2017-01-11
 [CARBONDATA-627]fix union test case for spark2

QiangCai:
Analyze:
Union test case failed in spark2. The result of union query is twice of the result of left query.

Root Cause:
CarbonLateDecodeRule only use union.children.head plan to build all CarbonDictionaryTempDecoder.

Changes:
Use child plan to build each CarbonDictionaryTempDecoder.

### DIFF
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/allqueries/AllDataTypesTestCaseAggregate.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/allqueries/AllDataTypesTestCaseAggregate.scala
@@ -59,21 +59,4 @@ class AllDataTypesTestCaseAggregate extends QueryTest with BeforeAndAfterAll {
       Seq(Row(100005.8)))
   })
 
-  test("CARBONDATA-60-union-defect")({
-    sql("drop table if exists carbonunion")
-    import sqlContext.implicits._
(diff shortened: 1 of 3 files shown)


CarbonDataQA:
Build Success with Spark 1.5.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/560/

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/allqueries/AllDataTypesTestCaseAggregate.scala
```diff
@@ -59,21 +59,4 @@ class AllDataTypesTestCaseAggregate extends QueryTest with BeforeAndAfterAll {
       Seq(Row(100005.8)))
   })
 
-  test("CARBONDATA-60-union-defect")({
```
jackylk: here, only one test case is removed from carbon-spark module, but in test report, it says two are deleted, can you check why?
http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/560/testReport/

QiangCai: Because the previous builder 559 added one test case, so the builder 560 has two deleted test case.

jackylk:
LGTM

PR #535, opened 2017-01-13
[CARBONDATA-542] Abnormal behaviour of Int datatype

anuragknoldus:
**Analysis:** Currently Short and Int datatypes are treated as long and at the time of storing in carbon data files delta compression is used which compresses the data based on min and max values of the column.
While parsing the values for these datatypes, we use Double data type parser and extract long value from that. Code snippet as below. Double.valueOf(msrValue).longValue(). During Double parsing, the value precision is lost. Integers greater than +9007199254740992 or less than -9007199254740992 may not have an exact representation and the sparseness increases exponentially as integers increase above the positive or decrease below the negative values, respectively.

**Fix:** Parse the values based on their datatypes to avoid this problem


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/592/

ravipesala:
LGTM

PR #542, opened 2017-01-16
[CARBONDATA-440] Update with no kettle fix

ravikiran23:
For iud data load flow will be used. so in the case of NO-KETTLE, need to handle data load.

load count/ segment count should be string because in compaction case it will be 2.1


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/626/

PR #548, opened 2017-01-18
[CARBONDATA-634] Load Query options invalid values are not consistent behavior.

mohammadshahidkhan:
By default if no value is set by default BAD_RECORDS_ACTION will be initialized with FORCE.
But if the value for BAD_RECORDS_ACTION is set then the the value should be validated.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/650/

ravipesala:
LGTM

PR #549, opened 2017-01-18
[CARBONDATA-561]Merge the two CarbonOption.scala into one under spark-common

Hexiaoqiao:
Class CarbonOption defined under Spark2 is exactly same as it defined under Spark-common, so as the package path, so i think we can delete CarbonOption in Spark2, and merge the below two CarbonOption.scala into single under spark-common:
/integration/spark-common/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala
/integration/spark2/src/main/scala/org/apache/carbondata/spark/CarbonOption.scala


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/655/

chenliang613:
LGTM

QiangCai:
LGTM

PR #554, opened 2017-01-19
[CARBONDATA-661] misc cleanup in carbon core

JihongMA:
cleanup un-exercised code/field/functions as well as minor fix.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/keygenerator/columnar/impl/MultiDimKeyVarLengthEquiSplitGenerator.java
+++ b/core/src/main/java/org/apache/carbondata/core/keygenerator/columnar/impl/MultiDimKeyVarLengthEquiSplitGenerator.java
@@ -18,7 +18,6 @@
 package org.apache.carbondata.core.keygenerator.columnar.impl;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -143,8 +142,9 @@ public class MultiDimKeyVarLengthEquiSplitGenerator extends MultiDimKeyVarLength
 
   private int[] convertToArray(List<Integer> list) {
     int[] ints = new int[list.size()];
-    for (int i = 0; i < ints.length; i++) {
-      ints[i] = list.get(i);
+    int i = 0;
+    for (Integer ii : list) {
+      ints[i++] = ii;
     }
     return ints;
   }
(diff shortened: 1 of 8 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/677/

core/src/main/java/org/apache/carbondata/core/keygenerator/columnar/impl/MultiDimKeyVarLengthEquiSplitGenerator.java
```diff
@@ -143,17 +142,18 @@ private void intialize() {
 
   private int[] convertToArray(List<Integer> list) {
```
jackylk: After second thought, I think these kind of functions are unnecessary, we should use
```
    List<Integer> list = new LinkedList<>();
    Integer[] array = x.toArray(new Integer[list.size()]);
```
from JDK instead of writing our own code

JihongMA: this is all depending on if we want to hold a primitive type array not not. these type of boxing/unboxing for primitive array type is annoying

jackylk: Yes, but this is another issue, which can be fixed in another PR. What I mean here is caller of `convertToArray` can use `collection.toArray` from java.util instead.

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/745/

core/src/main/java/org/apache/carbondata/core/keygenerator/columnar/impl/MultiDimKeyVarLengthEquiSplitGenerator.java
JihongMA: the whole purpose of this function is to unbox Integer array into integer array, not converting List[Integer] to Integer[]. 

 I looked closely at the caller of the function initialize(), which can be rewritten to avoid this back and forth set/list/array conversion.  and can be addressed in a separate PR.

jackylk: ok

jackylk:
LGTM

PR #562, opened 2017-01-20
[CARBONDATA-668] Dataloads  fail when no. of column in load query is greater than the no. of column in create table

manishgupta88:
Problem: Dataloads  fail when no. of column in load query is greater than the no. of column in create table

Analysis: During data load mapping of columns order is maintained between schema and columns provided in Fileheader/CSV file. If any duplicate column is provided in the fileheader, the length of array exceeds for the mapping order array as extra columns are getting matched and hence array index of bound exception is thrown.

Fix: Add a check for looping only till the mapping order array length is less than the number of schema columns.

Impact Area: Data load flow


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/717/

ravipesala:
LGTM

PR #565, opened 2017-01-20
[CARBONDATA-635]Fix Insert into query when used different names.

ravipesala:
When two tables have different column names then insert query uses alias, in strategy we are not handling alias so it fails.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/729/

gvramana:
LGTM

PR #580, opened 2017-01-30
[CARBONDATA-685]added vaidation for table name with spaces IN tableName option using carbon source

anubhav100:
when using carbon source i am able to create table with spaces
logs====
0: jdbc:hive2://localhost:10000> CREATE TABLE table (ID Int, date Timestamp, country String, name String, phonetype String, serialname String, salary Int) USING org.apache.spark.sql.CarbonSource OPTIONS("tableName"="t a b l e ");
---------+
Result
---------+
---------+
No rows select
here table with empty spaces is created in hdfs
it should not allow this
 
with this pr this check will be validated


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/780/

jackylk:
LGTM

PR #600, opened 2017-02-15
[CARBONDATA-705]Make partition distribution as configurable

ravipesala:
Make the partition distribution as configurable and keep spark distribution as default.

User can use `spark.carbon.custom.distribution=true` to use carbon's custom partition distribution.
By default this configuration is false, so it uses sparks distribution.

### DIFF
--- a/integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
+++ b/integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
@@ -143,14 +143,12 @@ class CarbonScanRDD(
         }
         noOfNodes = nodeBlockMapping.size
       } else {
-        var i = 0
-        splits.asScala.foreach { s =>
+        splits.asScala.zipWithIndex.foreach { splitWithIndex =>
           val multiBlockSplit =
             new CarbonMultiBlockSplit(identifier,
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/905/

integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -141,6 +142,17 @@ class CarbonScanRDD(
           }
         }
         noOfNodes = nodeBlockMapping.size
+      } else {
+        var i = 0
+        splits.asScala.foreach { s =>
```
jackylk: rename `s` to `split`, it is better to use a meaningful name

jackylk: And I think you can use `zipWithIndex` without using `i`

ravipesala: Ok, Updated

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/923/

jackylk:
LGTM

PR #619, opened 2017-03-01
[CARBONDATA-735]Dictionary performance issue with multiple task in same executor

kumarvishal09:
**Problem:**
Currently when more than 1 task is getting launched in one node for a query both the task is trying to load the dictionary data and its impacting dictionary loading performance.
**Solution:**
Need to add monitor for dictionary loading one task will load the dictionary and other task will be waiting and share the same dictionary data


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/986/

ravipesala:
LGTM

PR #621, opened 2017-03-02
[CARBONDATA-738] resolved bug for Able to load dataframe with boolean type in a carbon…

anubhav100:
i created a dataframe with boolean type as follows
case class People(name: String, occupation: String, consultant: Boolean)
val people = List(People("sangeeta", "engineer", true), People("pallavi", "consultant", true))
val peopleRDD: RDD[People] = cc.sc.parallelize(people)
val peopleDF: DataFrame = peopleRDD.toDF("name", "occupation", "id")
peopleDF.write
.format("carbondata")
.option("tableName", "carbon2")
.option("compress", "true")
.mode(SaveMode.Overwrite)
.save()
cc.sql("SELECT * FROM carbon2").show()
currently boolean type is not supported in carbon data but table gets created
but it shows me null values
------------------
name	occupation	id
------------------
pallavi	consultant	null
sangeeta	engineer	null
------------------
for boolean type it should throw unsupported type exception it is not supported in carbon data

there is problem with carbondataframe writer


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/992/

ravipesala:
LGTM

PR #627, opened 2017-03-07
CARBONDATA-748

simafengyun:
fix jira CARBONDATA-748, use binary search to replace linear search
the performance is much better no ,from more than 10 seconds to a few millisecond

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/ByteUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/ByteUtil.java
@@ -152,6 +152,159 @@ public final class ByteUtil {
       return (x1 + Long.MIN_VALUE) < (x2 + Long.MIN_VALUE);
     }
 
+
+/*		public static int binarySearch(byte[] a, int fromIndex, int toIndex, byte[] key) {
+			int keyLength = key.length;
+			rangeCheck(a.length, fromIndex, toIndex, keyLength);
+			return binarySearch0(a, fromIndex, toIndex / keyLength, key);
+		}
+
+		// Like public version, but without range checks.
+		private static int binarySearch0(byte[] a, int fromIndex, int toIndex, byte[] key) {
+			int low = fromIndex;
+			int high = toIndex - 1;
+			int keyLength = key.length;
+			// int high = toIndex/keyLength;
+
+			while (low <= high) {
+				int mid = (low + high) >>> 1;
+				// byte midVal = a[mid];
+
+				int result = ByteUtil.UnsafeComparer.INSTANCE.compareTo(a, mid * keyLength, keyLength, key, 0,
+						keyLength);
+
+				if (result < 0)
+					low = mid + keyLength;
+				else if (result > 0)
+					high = mid - keyLength;
+				else
+					return mid; // key found
+			}
+			return -(low + 1); // key not found.
+		}*/
+	/**
+	 * Checks that {@code fromIndex} and {@code toIndex} are in the range and toIndex % keyLength = 0
+	 * and throws an exception if they aren't.
+	 */
+	private static void rangeCheck(int arrayLength, int fromIndex, int toIndex, int keyWordLength) {
+		if (fromIndex > toIndex || toIndex % keyWordLength != 0) {
+			throw new IllegalArgumentException("fromIndex(" + fromIndex + ") > toIndex(" + toIndex + ")");
+		}
+		if (fromIndex < 0) {
+			throw new ArrayIndexOutOfBoundsException(fromIndex);
+		}
+		if (toIndex > arrayLength) {
+			throw new ArrayIndexOutOfBoundsException(toIndex);
+		}
+	}
+	
+	/**
+	 * search a specific key's range boundary in sorted byte array
+	 * 
+	 * @param dataChunk
+	 * @param fromIndex
+	 * @param toIndex
+	 *            it's max value should be word Number in dataChunk, equal to indataChunk.length/keyWord.length
+	 * @param keyWord
+	 * @return int[] contains a range's lower boundary and upper boundary
+	 */
+	public static int[] binaryRangeSearch(byte[] dataChunk, int fromIndex, int toIndex, byte[] keyWord) {
+
+
+		int keyLength = keyWord.length;
+		rangeCheck(dataChunk.length, fromIndex, toIndex, keyLength);
+		
+		// reset to index =  word total number in the dataChunk
+		toIndex = toIndex/keyWord.length;	
+		
+		int[] rangeIndex = new int[2];
+		int low = fromIndex;
+		int high = toIndex - 1;
+
+		while (low <= high) {
+			int mid = (low + high) >>> 1;
+
+			int result = ByteUtil.UnsafeComparer.INSTANCE.compareTo(dataChunk, mid * keyLength, keyLength, keyWord, 0,
+					keyLength);
+
+			if (result < 0)
+				low = mid + 1;
+			else if (result > 0)
+				high = mid - 1;
+			else {
+				// key found  then find the range bound
+				rangeIndex[0] = binaryRangeBoundarySearch(dataChunk, low, mid, keyWord, false);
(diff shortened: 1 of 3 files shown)


CarbonDataQA:

core/src/main/java/org/apache/carbondata/core/util/ByteUtil.java
```diff
@@ -152,6 +152,159 @@ static boolean lessThanUnsigned(long x1, long x2) {
       return (x1 + Long.MIN_VALUE) < (x2 + Long.MIN_VALUE);
     }
 
+
```
Hexiaoqiao: please delete this annotation if not use any more.

core/src/main/java/org/apache/carbondata/core/util/ByteUtil.java
```diff
@@ -152,6 +152,159 @@ static boolean lessThanUnsigned(long x1, long x2) {
...
+			}
+			return -(low + 1); // key not found.
+		}*/
+	/**
+	 * Checks that {@code fromIndex} and {@code toIndex} are in the range and toIndex % keyLength = 0
+	 * and throws an exception if they aren't.
+	 */
+	private static void rangeCheck(int arrayLength, int fromIndex, int toIndex, int keyWordLength) {
+		if (fromIndex > toIndex || toIndex % keyWordLength != 0) {
+			throw new IllegalArgumentException("fromIndex(" + fromIndex + ") > toIndex(" + toIndex + ")");
```
Hexiaoqiao: the exception msg may be misleading, please give some msg about `toIndex % keyWordLength != 0`.

core/src/main/java/org/apache/carbondata/scan/filter/executer/IncludeFilterExecuterImpl.java
```diff
@@ -154,24 +154,45 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+	      for (int k = 0; k < filterValues.length; k++) {
+	    	  
+	    	  int[] index = ByteUtil.UnsafeComparer.INSTANCE.binaryRangeSearch(dataChunk, 0, dataChunk.length, filterValues[k]);
+	    	  for(int i = index[0]; i<=index[1];i++){
+	    		  
+	    		  bitSet.set(i);
+	    	  }
+	      }
+	  
+	      //below comments code is used to confirm the binary range search is correct
```
Hexiaoqiao: please delete annotation if not used.

core/src/main/java/org/apache/carbondata/core/util/ByteUtil.java
```diff
@@ -152,6 +152,159 @@ static boolean lessThanUnsigned(long x1, long x2) {
...
+  /**
+   * use to search a specific keyword's lower boundary and upper boundary according to upFindFlg
+   * @param dataChunk
+   * @param fromIndex
+   * @param toIndex
+   * @param keyWord
+   * @param upFindFlg true:find upper boundary  false: find lower boundary
+   * @return boundary index 
+   */
+	public static int binaryRangeBoundarySearch(byte[] dataChunk, int fromIndex, int toIndex, byte[] keyWord, boolean upFindFlg) {
```
Hexiaoqiao: 1. there are some common codes between `binaryRangeSearch` and `binaryRangeBoundarySearch`, is it possible to merge them into one method?
2. are you interesting to impl. a complete function about Interval (of Mathematics).
FYI.

Hexiaoqiao:
@simafengyun 
thanks for your work, please rename title follow the format `[CARBONDATA-issue number>] Title of the pull request`.

zzcclp:
Is this pr just be fixed for branch-0.2?

ravipesala:
add to whitelist

CarbonDataQA:
Build Failed  with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1022/

CarbonDataQA:
Build Failed  with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1023/

simafengyun:
yes
I just fixed on branch-0.2.
If code is ok, I can continue to fix for version 1.0










At 2017-03-07 17:20:44, "Zhichao Zhang" <notifications@github.com> wrote:


Is this pr just be fixed for branch-0.2?

ravipesala:
@simafengyun Thanks for working on it.
Range binary search is impressive. we can use same search in other places as well.
But using this search in `setFilterdIndexToBitSet` is not always correct. Because here the data may not be sorted always, Only for the first column of dimensions is sorted naturally because of mdk order.
And if the data is sorted explicitly(it means it has inverted index) then it goes to another method `setFilterdIndexToBitSetWithColumnIndex`.  So here we need extra checks to do binary search in the method `setFilterdIndexToBitSet` , that is like whether the column is naturally sorted or not.This information we may need get from store.  
Please use this Range binary search in 'setFilterdIndexToBitSetWithColumnIndex' as well.

ravipesala:
@simafengyun Please raise this PR on master as we don't have any plans to provide patches on branch-0.2. And also please format the code.

simafengyun:
You mentioned the below,
>>>>>>>>>>>>>>>>>>>>>>>>>>
But using this search in setFilterdIndexToBitSet is not always correct. Because here the data may not be sorted always, Only for the first column of dimensions is sorted naturally because of mdk order.
>>>>>>>>>>>>>>>>>>>>>>>>>>
 but i don't think so. 
the order you mentioned is the logical level order(MDK).
As I know, for dimension column, it has physical order in chunk level.
for dimension data which has dictionary encode, the dictionary data will sorted in blocklet level and keep the order in chunk on the physical  disk.
So after the one chunk dimension data read, it will keep the order, so I think it is fit for the binary search.


if I was wrong, please feel free to tell me, thanks 










At 2017-03-07 17:44:28, "Ravindra Pesala" <notifications@github.com> wrote:


@simafengyun Thanks for working on it.
Range binary search is impressive. we can use same search in other places as well.
But using this search in setFilterdIndexToBitSet is not always correct. Because here the data may not be sorted always, Only for the first column of dimensions is sorted naturally because of mdk order.
And if the data is sorted explicitly(it means it has inverted index) then it goes to another method setFilterdIndexToBitSetWithColumnIndex. So here we need extra checks to do binary search in the method setFilterdIndexToBitSet , that is like whether the column is naturally sorted or not.This information we may need get from store.
Please use this Range binary search in 'setFilterdIndexToBitSetWithColumnIndex' as well.

ravipesala:
@simafengyun you are right, default we sort the data but we can avoid sorting using `no_inverted_index` property at tblproperties level. So if user uses this configuration then we cannot do binary search always.

simafengyun:
so binary search only fit the below cases，please comfirm thanks

1 the first dimension column ， since it's logicl order are the same with physical order in blocklet level，so it does not need inverted index

2 for dimension column which has inverted index

question 1

if a dimension column is not the first colmn and in a blocklet it just has the same order for both logical and physical，will carbondata create inverted index for it?

question 2

is there an easy way to identify whether a dimension column has no_inverted_index property config by user in source code ，so it will be helpful for me to add a condition to use binary search



thanks

在 2017-03-07 18:35:05，"Ravindra Pesala" <notifications@github.com> 写道：


@simafengyun you are right, default we sort the data but we can avoid sorting using no_inverted_index property at tblproperties level. So if user uses this configuration then we cannot do binary search always.

ravipesala:
Yes, those are the 2 cases we can do binary search on dimension.

Regarding question1 , no it does not create inverted index if the data is already sorted(naturally sorted).

Regarding question2, it is an issue pending from long time, it requires changes from data writer and as well as data reader to update the sort state.  
There are attributes like SortState and Encoders in DataChunk of format needs to updated properly while writing and get them back while reading to know the dimension has inverted_index or not, or it is naturally sorted or not. 

My suggestion is don't change the method of  `setFilterdIndexToBitSet` as the old logic is required for no_inverted index and non sorted columns. Update the the method `setFilterdIndexToBitSetWithColumnIndex` with your range search.  To know the dimension is naturally sorted or not we can add another PR to update the read/write logic.

PR #644, opened 2017-03-10
[CARBONDATA-757]Big decimal optimization

ravipesala:
Currently Decimal is converted to bytes and using LV (length + value) format to write to store. And while getting back read the bytes in LV format and convert back the bigdecimal.
We can do following operations to improve storage and processing.
1. if decimal precision is less than 9 then we can fit in int (4 bytes)
2. if decimal precision is less than 18 then we can fit in long (8 bytes)
3. if decimal precision is more than 18 then we can fit in fixed length bytes(the length bytes can vary depends on precision but it is always fixed length)
So in this approach we no need store bigdecimal in LV format, we can store in fixed format.It reduces the memory.

Carbondata format changes -> Added fixedLength in datachunk to know about the column length of big  decimal. This attribute can be used in case of char(fixedlength) or varchar(fixedlength) datatypes as well.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/chunk/reader/measure/v3/CompressedMeasureChunkFileBasedReaderV3.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/chunk/reader/measure/v3/CompressedMeasureChunkFileBasedReaderV3.java
@@ -226,10 +226,12 @@ public class CompressedMeasureChunkFileBasedReaderV3 extends AbstractMeasureChun
     ByteBuffer rawData = measureRawColumnChunk.getRawData();
     rawData.position(copyPoint);
     rawData.get(data);
+    values.setFixedLength(dataChunk3.fixedLength);
     values.uncompress(compressionModel.getConvertedDataType()[0], data, 0,
         measureColumnChunk.data_page_length, compressionModel.getMantissa()[0],
         compressionModel.getMaxValue()[0], measureRawColumnChunk.getRowCount()[pageNumber]);
     CarbonReadDataHolder measureDataHolder = new CarbonReadDataHolder(values);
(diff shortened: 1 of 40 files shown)


ravipesala:
Build will not compile as there are carbon-format changes.

CarbonDataQA:
Build Failed  with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1082/

ravipesala:
Test results witj 100 million data 
**DDL**
CREATE TABLE perftesta (c1 string,c2 string,c3 string,c4 string,c5 string,c6 bigint,c7 decimal(7,2),c8 int,c9 decimal(7,2),c10 decimal(15,2)) STORED BY 'carbondata'

**Queries**
Q1 -> SELECT count(c1),count(c2),count(c3),count(c4),count(c5),count(c6),count(c7),count(c8),count(c9),count(c10) FROM perftesta99;
Q2 -> SELECT sum(c7), sum(c8), sum(9), sum(c10) FROM perftesta99 WHERE c2="P2_75" and c7<5;
Q3 -> SELECT c2, c5, count(distinct c1), sum(c7) FROM perftesta99 WHERE c4="P4_4" and c5="P5_7" GROUP BY c2, c5;

**Master Code**
Load time  -> 576 seconds
Data size after load  -> 1800MB
Query(first_reading, second_reading)
Q1(25.27, 21.794)
Q2(27.296, 28.21)
Q3(7.383, 5.103)

**This PR Code**
Load time -> 431 seconds
Data size after load  -> 1720MB
Query(first_reading, second_reading)
Q1(18.507,14.427)
Q2(24.102, 23.322)
Q3(6.87,5.079)

ravipesala:
Closing it as another PR https://github.com/apache/carbondata/pull/1117 raised

PR #657, opened 2017-03-15
[CARBONDATA-770] Fixed null filter issue

kumarvishal09:
In can of exclude filter dictionary column DictionaryColumnVisitor is adding null value surrogate twice and because of this is not null query is giving wrong result


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1157/

ravipesala:
LGTM

PR #676, opened 2017-03-20
[CARBONDATA-797] Data loss for BigInt datatype if data contains long max and min values

manishgupta88:
When data contains long max and min values for a measure column with bigInt datatype, the delta compression selected is DATA_BYTE which is incorrect. For selecting the delta compression min value is decremented from max value and here the min value is negative, so it performs addition operation and goes out of long range. When a long value goes out of range it starts again from the long max negative value which results in wrong compression selection.

This leads to data loss and incorrect query results.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1241/

gvramana:
LGTM

PR #680, opened 2017-03-21
[CARBONDATA-800] ArrayIndexOfBound Exception thrown when block size is specified as 2048 MB

manishgupta88:
If table block size if specified for its max value 2048, then while converting it into bytes it goes out of integer range and hence a negative value is assigned for blocklet size. Because of this exception is thrown.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1253/

gvramana:
LGTM

PR #711, opened 2017-03-30
[CARBONDATA-834] Fixed Describe Table for Presto

bhavya411:
Changed the way the Metadata was getting created used column schema instead of dimensions and measure seperately.


ffpeng90:
LGTM

PR #718, opened 2017-03-31
[CARBONDATA-843]null pointer exception was thrown when floor operation is done on decimal column

akashrn5:
problem: null pointer exception is thrown when floor operation is done on decimal column.
analysis: when floor operation was done on decimal column, scale was greater than precision.During floor operation , we try to change the precision of the data.
solution: When sending to the spark layer for performing floor operation, we need to explicitly send the precision and scale to the spark layer which are taken from the CarbonMeasure


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1399/

PR #719, opened 2017-04-01
[WIP][CARBONDATA-844] Avoid to get useless splits

watermen:
In current implements of CarbonInputFormat.getDataBlocksOfSegment, 
1. Get all of the carbondata splits in segments directory.
2. Read the carbonindex and construct the B-tree.
3. Apply filter and get matching splits.

I think we get some useless splits and the operator of getSplits is expensive. So we'd better to do the getSplits after filter:
1. List the segment directory, and filter the path of carbonindex.
2. Read the carbonindex and construct the B-tree.
3. Apply filter and get matching blocks.
4. Get carbondata splits from filtered blocks.

TODOList:
- [ ] How to do something about IUD(isValidBlockBasedOnUpdateDetails)?
- [ ] Fix bug in suite

@jackylk @QiangCai Plz review the code when you have time.


CarbonDataQA:
Build Failed  with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1401/

PR #723, opened 2017-04-03
[CARBONDATA-845] Insert Select Into Same Table

sounakr:
Problem : Insert Select Implementation in current Spark-2.1 prohibits to read and write into same table. 

Analysis : This restriction is imposed by spark in case InsertableRelation is extended. Current insert select implementation in spark-2.1 CarbonDatasourceHadoopRelation extends InsertableRelation and override the insert method. Spark rules in case of InsertableRelation checks if source and target tables are same, then it fails and prompts  "Cannot insert overwrite into table that is also being read from."

Fix:: Spark 1.5 implementation of insert select in carbon allows reading and writing into same table, so as Hive and Parquet. Therfore as part of current fix reimplemented insert-select logic of Spark-2.1 similar to Spark-1.5. 

 a) CarbonDatasourceHadoopRelation doesn't extends InsertableRelation.
 b) New CarbonPreInsertionCasts Rule is written for spark-2.1 (For Spark-1.5 it was already present). 
 d) This new Rule is registered CarbonSessionState, 
 c) During Insert Select processing the Logical Plan wrapped by InsertIntoCarbonTable and later execute LoadTableByInsert to process insert select.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1406/

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonPreInsertionCasts.scala
```diff
@@ -0,0 +1,62 @@
+/*
```
QiangCai: InsertInto function is a part of IUD feature. No need to create a new file.

ravipesala:
Can't we have same restriction like Spark, is it really required to override this behavior?

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonPreInsertionCasts.scala
sounakr: Done, currently to keep 1.5 and 2.1 in sync, CarbonAnalysisRules.scala will keep the rules of IUD and Insert Select. Created CarbonAnalysisRules.scala in 2.1 where currenly on Insert-Select is there as IUD is still not ported to 2.1

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1419/

sounakr:
@ravipesala Insert Select from same table is allowed both in Hive and Parquet. Spark-5746 introduced this  behavior as they are deleting the source table data first before inserting when overwrite is turned ON. But for carbon insert select Overwrite is always false and as other data sources are allowing insert select into same table it is good to have. Also the below kind of queries will help to load data. 
insert into tab1 sel a *4, b*4 from tab1;

integration/spark2/src/main/scala/org/apache/spark/sql/execution/command/DDLStrategy.scala
```diff
@@ -54,6 +54,9 @@ class DDLStrategy(sparkSession: SparkSession) extends SparkStrategy {
             identifier.table.toLowerCase)) :: Nil
       case ShowLoadsCommand(databaseName, table, limit) =>
         ExecutedCommandExec(ShowLoads(databaseName, table.toLowerCase, limit, plan.output)) :: Nil
+      case InsertIntoCarbonTable(relation: CarbonDatasourceHadoopRelation,
+      _, child: LogicalPlan, _, _) =>
```
ravipesala: Please format it properly

ravipesala:
LGTM

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1453/

ravipesala:
LGTM

PR #736, opened 2017-04-05
[CARBONDATA-871] If store type is HDFS then set HDFS lock as default 

ManoharVanam:
Problem : if no locktype is configured and store type is HDFS , while acquiring lock getting exception like "path is  null".

Solution : if no locktype is configured and store type is HDFS set HDFS lock as default 

Query Example: DELETE SEGMENTS FROM TABLE DataRetentionTable where STARTTIME before '2099-01-01'


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1444/

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1470/

ravipesala:
LGTM

PR #753, opened 2017-04-06
[CARBONDATA-877] Fixed ArrayIndexOutOfBoundException for ColumnarVectorWrapper

kunal642:
Row id was being incremented with the column number and eventually the index on which we had to put the data would be greater than the array size.


ravipesala:
LGTM

PR #787, opened 2017-04-12
[CARBONDATA-911] Updated ddl-operation-on-carbondata.md

shivangi1015:
Corrected the example in bucketing section.


CarbonDataQA:

PR #811, opened 2017-04-18
[CARBONDATA-949] null value handled during compaction

rahulforallp:
After load if new column added to the table and minor compaction has been performed .
Then new column contains null value by default or if bad_record exists . And cast should not done with null value.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1679/

ravipesala:
LGTM

PR #816, opened 2017-04-19
[CARBONDATA-944] Fixed unwanted exception in case of Drop table through spark-sql-CLI

bhavya411:
This PR includes a fix for Table Already Exists exception in case of Drop Table command.

Build verification with spark-1.6 successful
Build with spark-2.1 successful
Verified The defect Manually

                 
---

### DIFF
--- a/integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
+++ b/integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
@@ -213,11 +213,12 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
       throw new MalformedCarbonCommandException("Table Name Should not have spaces ")
     }
     try {
-      CarbonEnv.get.carbonMetastore.lookupRelation(Option(dbName), tableName)(sparkSession)
-      CarbonEnv.get.carbonMetastore.storePath + s"/$dbName/$tableName"
+      CarbonEnv.getInstance(sparkSession).carbonMetastore
+        .lookupRelation(Option(dbName), tableName)(sparkSession)
+      CarbonEnv.getInstance(sparkSession).carbonMetastore.storePath + s"/$dbName/$tableName"
     } catch {
       case ex: Exception =>
-        throw new Exception(s"do not have $dbName and $tableName for carbon table", ex)
+        throw new Exception(s"Do not have $dbName and $tableName", ex)
     }
   }
 
--- a/integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
+++ b/integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
@@ -112,7 +112,17 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
       dataSchema: StructType): BaseRelation = {
     CarbonEnv.getInstance(sqlContext.sparkSession)
     addLateDecodeOptimization(sqlContext.sparkSession)
-    val path = createTableIfNotExists(sqlContext.sparkSession, parameters, dataSchema)
+    val dbName: String = parameters.getOrElse("dbName",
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1694/

integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
```diff
@@ -184,4 +194,29 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
...
+      tableName : String): String = {
+
+    if (StringUtils.isBlank(tableName)) {
+      throw new MalformedCarbonCommandException("The Specified Table Name is Blank")
+    }
+    if (tableName.contains(" ")) {
+      throw new MalformedCarbonCommandException("Table Name Should not have spaces ")
+    }
+    try {
+      CarbonEnv.get.carbonMetastore.lookupRelation(Option(dbName), tableName)(sparkSession)
```
chenliang613: Why not using the below code , keep consistent with createTableIfNotExists?

CarbonEnv.getInstance(sparkSession).carbonMetastore
        .lookupRelation(Option(dbName), tableName)(sparkSession)

integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
```diff
@@ -184,4 +194,29 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
...
+      CarbonEnv.get.carbonMetastore.storePath + s"/$dbName/$tableName"
```
chenliang613: CarbonEnv.getInstance(sparkSession).carbonMetastore.storePath + s"/$dbName/$tableName"

integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
```diff
@@ -184,4 +194,29 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
...
+    } catch {
+      case ex: Exception =>
+        throw new Exception(s"do not have $dbName and $tableName for carbon table", ex)
```
chenliang613: How about : 
throw new Exception(s"Do not have $dbName and $tableName", ex)

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1788/

integration/spark2/src/main/scala/org/apache/spark/sql/CarbonSource.scala
```diff
@@ -112,7 +112,17 @@ class CarbonSource extends CreatableRelationProvider with RelationProvider
       dataSchema: StructType): BaseRelation = {
     CarbonEnv.getInstance(sqlContext.sparkSession)
     addLateDecodeOptimization(sqlContext.sparkSession)
-    val path = createTableIfNotExists(sqlContext.sparkSession, parameters, dataSchema)
+    val dbName: String = parameters.getOrElse("dbName",
+      CarbonCommonConstants.DATABASE_DEFAULT_NAME).toLowerCase
+    val tableName: String = parameters.getOrElse("tableName", "default_table").toLowerCase
```
jackylk: can you do this check in caller function where you can use CarbonOption to get the dbName and tableName instead of doing it here?

bhavya411: @jackylk  I can not call it in the caller function as the createRelation method is a framework metthod so it can be called while Using clause is used , if I move the code to caller function the createTableIfnotExists will not be called in that scenario. Please let me know if we still need to do it.

chenliang613:
LGTM

PR #817, opened 2017-04-19
[CARBONDATA-958] update timestamp file when schema file is updated due to auto high cadinality detection and conversion to no-dictionary

sraghunandan:


gvramana:
LGTM

PR #830, opened 2017-04-21
[CARBONDATA-969] Don't persist rdd because it is only use once

watermen:
In GlobalDictionaryUtil.readAllDictionaryFiles, don't persist rdd because it is only use once.

@foryou2030 Can you check it because persist is added by you?
cc @QiangCai

### DIFF
--- a/integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
+++ b/integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
@@ -630,7 +630,7 @@ object GlobalDictionaryUtil {
     try {
       // read local dictionary file, and spilt (columnIndex, columnValue)
       val basicRdd = sqlContext.sparkContext.textFile(allDictionaryPath)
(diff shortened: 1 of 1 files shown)


integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -630,7 +630,7 @@ object GlobalDictionaryUtil {
     try {
       // read local dictionary file, and spilt (columnIndex, columnValue)
       val basicRdd = sqlContext.sparkContext.textFile(allDictionaryPath)
-        .map(x => parseRecord(x, accumulator, csvFileColumns)).persist()
+        .map(x => parseRecord(x, accumulator, csvFileColumns))
```
foryou2030: yes, it is no need. thank you

CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1724/

jackylk:
LGTM

PR #834, opened 2017-04-21
[CARBONDATA-970] getSplits() from carbonScanRDD not ignoring the blocks which are cleared using clean files DDL

sraghunandan:
problem: After clean files DDL is run, the blocks which are deleted are also getting in getSplits() of CarbonScanRDD.scala

Solution: when the segments index map is created in carbonInputFormat.scala , get only the valid tasks which contain the valid blocks


CarbonDataQA:
Build Failed  with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1735/

PR #768, opened 2017-04-08
[CARBONDATA-829] resolved bug for dictionary_exclude not working using carbondatasource

anubhav100:
.include stringtype in table creator  method isDataTypeSupportedForDictionary_Excluded method
.refactor the bad naming convention
.added test case for bug

PR #894, opened 2017-05-08
[CARBONDATA-1038] Fixed exception in dictionary_exclude for string datatype

kunal642:
Analysis: 
1. The data type that could be added to the dictionary_exclude list was listed as "string" instead of "stringtype" which caused the exception.
2. If no table name is provided in the options while creating table then the default table name of "defalut_table" is taken. Which is not the expected result.

Solution: 
1. changed the datatype from "string" to "stringtype".
2. Throw exception instead of taking "default_table" as the table name if no name is provided.


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1953/

ravipesala:
It is already handled in https://github.com/apache/carbondata/pull/768

PR #904, opened 2017-05-11
[CARBONDATA-1045] Mismatch in message display with insert and load operation on failure due to bad records in update operation

manishgupta88:
Problem: When bad records action is set to fail and any IUD operation… is executed and it fails due to bad records error message is not displayed correctly because of which user is not clear with the cause of update operation failure. Whereas in the same case in other operations like data load and insert into, if there is any failure due to bad record proper error message is displayed to the user for failure due to bad record.

Fix: Instead of forming own message get the executor failure message and send it to update operation as exception message.

### DIFF
--- a/integration/spark-common-test/src/test/resources/IUD/bad_record.csv
+++ b/integration/spark-common-test/src/test/resources/IUD/bad_record.csv
@@ -1,0 +1,2 @@
+item,name
+2,Apple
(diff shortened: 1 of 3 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/1991/

ravipesala:
LGTM

PR #919, opened 2017-05-17
[CARBONDATA-1056] Restrict starting DictionaryServer if no column has Dictionary encoding

kunal642:
Analysis: If all the dictionary columns are marked as dictionary exclude then the dictionary generator is not created for that table but there was no check in the data loading flow for the same. Thus the query fails with "cannot write dictionary" exception.

Solution: Added check in the loading flow for Dictionary and DirectDictionary encoding


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2050/

ravipesala:
LGTM

PR #922, opened 2017-05-17
[CARBONDATA-1062] Data load fails if a column specified as sort column is of numeric data type

manishgupta88:
Fixes include: 
1. If a numeric data type column is specified as sort column and if it contains non numeric value then data load fails.
2. Intermediate merger thread is not shutdown before starting final merge sort during alter table compaction. This can lead to file not found exception.
3. During compaction after restructure operation, number of sort columns and number of no dictionary sort columns are not getting population during sorting which will lead to unsorted data being written to file and sometimes failure of compaction after restructure operation.

Fix: 
1. In case when bad records action is force and given value is invalid for a datatype update the row with default null value array.
2. Call finish on intermediate merger thread before starting final merge sort merge
3. Populate the number of sort and no sort columns from carbon table during compaction after restructure operation.

### DIFF
--- a/integration/spark-common-test/src/test/resources/numeric_column_invalid_values.csv
+++ b/integration/spark-common-test/src/test/resources/numeric_column_invalid_values.csv
@@ -1,0 +1,6 @@
+1,Pallavi,25
+2,Rahul,24
+3,Prabhat,twenty six
+7,Neha,25
+2,Geetika,22
+3,Sangeeta,26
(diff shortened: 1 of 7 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2056/

jbonofre:
retest this please

gvramana:
LGTM

PR #926, opened 2017-05-18
[CARBONDATA-1065] Added set command in carbon to update properties dynamically

ravipesala:
The below sql command updates the carbon properties in driver and as well as in executor dynamically in memory.
```
set key1=value1
```

It does not update carbon.properties file, it just updates the running driver and executor properties in memory, so after restart of driver all set properties will be gone.

### DIFF
--- a/integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/NewCarbonDataLoadRDD.scala
+++ b/integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/NewCarbonDataLoadRDD.scala
@@ -20,6 +20,7 @@ package org.apache.carbondata.spark.rdd
 import java.io.{IOException, ObjectInputStream, ObjectOutputStream}
 import java.nio.ByteBuffer
 import java.text.SimpleDateFormat
+import java.util
 import java.util.{Date, UUID}
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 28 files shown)


CarbonDataQA:
Build Success with Spark 1.6.2, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2080/

integration/spark-common/src/main/scala/org/apache/carbondata/spark/rdd/NewCarbonDataLoadRDD.scala
```diff
@@ -125,12 +126,16 @@ class SparkPartitionLoader(model: CarbonLoadModel,
...
+  def initialize(addedProperies: util.Map[String, String]): Unit = {
     val carbonPropertiesFilePath = System.getProperty("carbon.properties.filepath", null)
     if (null == carbonPropertiesFilePath) {
       System.setProperty("carbon.properties.filepath",
         System.getProperty("user.dir") + '/' + "conf" + '/' + "carbon.properties")
     }
+    // Add the properties added in driver to executor.
+    CarbonProperties.getInstance().setProperties(addedProperies)
+    // Add the properties added in driver to executor.
+    CarbonProperties.getInstance().setProperties(addedProperies)
```
xuchuanyin: duplicated...

integration/spark/src/main/scala/org/apache/spark/sql/hive/execution/command/CarbonHiveCommands.scala
```diff
@@ -53,3 +55,15 @@ private[hive] case class DropDatabaseCascadeCommand(dbName: String,
     rows
   }
 }
+
+case class CarbonSetCommand(command: SetCommand)
+  extends RunnableCommand {
+
+  override val output = command.output
+
+  override def run(sparkSession: SQLContext): Seq[Row] = {
```
xuchuanyin: name it `sqlContext` would be better

xuchuanyin:
Hi, I have the following questions:

Will these properties ...

1. be in effect only in current session or until the node restarts ?  `(Effective Scope)`

2. be logged into log file like execution environment variables ?  `(Audit & Tuning)`

PR #932, opened 2017-05-21
[CARBONDATA-1074] Add TablePage and ConvertedRow, preparing for data load refactory

jackylk:
1. Add TablePage preparing for data load refactory.
2. Add a ConvertedRow to unify input to write step from different steps, includes:
- write step of normal sort table
- write step of no sort table
- compaction merging step
3. fix some typo


jackylk:
retest this please

processing/src/main/java/org/apache/carbondata/processing/store/TablePage.java
```diff
@@ -0,0 +1,203 @@
...
+ * all processing apply to TablePage can be done in vectorized fashion.
+ */
+class TablePage {
+
+  // For all dimension and measure columns, we store the column data directly in the page,
+  // the length of the page is the number of rows.
+
+  // TODO: we should have separate class for key columns so that keys are stored together in
+  // one vector to make it efficient for sorting
+  private KeyColumnPage keyColumnPage;
```
kumarvishal09: @jackylk I have one query related to this changes please correct me.
Can we use the ColumnVector interface for this?? the way we are storing the data while processing the vector batch during query, we can use the same interface while loading. There we will have flexibility for storing data in offheap/onheap both.

jackylk: You mean through ColumnarVectorWrapper? In fact, processing package should not depend on spark, so here I think it is better to use our own data structure. In future PR, we can change KeyPage to have offheap implementation.

integration/spark2/src/main/java/org/apache/carbondata/spark/readsupport/SparkRowReadSupportImpl.java
```diff
@@ -36,7 +36,7 @@
     isMeasure = new boolean[carbonColumns.length];
     dataTypes = new DataType[carbonColumns.length];
     for (int i = 0; i < carbonColumns.length; i++) {
-      isMeasure[i] = !carbonColumns[i].isDimesion();
+      isMeasure[i] = !carbonColumns[i].isDimension();
```
kumarvishal09: Now isMeasure() is added in CarbonColumn You can u use that method.

jackylk: actually, the initialization in this class is unnecessary, because `readRow` only return a `new GenericInternalRow(data)`, so I will move `isMeasure` variable

jackylk:
retest this please

chenliang613:
retest this please

chenliang613:
retest this please

chenliang613:
retest this please

processing/src/main/java/org/apache/carbondata/processing/newflow/row/ConvertedRow.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+package org.apache.carbondata.processing.newflow.row;
+
+import org.apache.carbondata.core.datastore.block.SegmentProperties;
+import org.apache.carbondata.core.keygenerator.KeyGenException;
+import org.apache.carbondata.core.keygenerator.KeyGenerator;
+import org.apache.carbondata.core.scan.wrappers.ByteArrayWrapper;
+
+public class ConvertedRow {
```
ravipesala: Why we should create new instance here for every row. Here all methods could be static. Just pass CarbonRow to the required method and get the converted data. So even `CarbonFactDataHandlerColumnar.addDataToStore` should only take `CarbonRow`.

processing/src/main/java/org/apache/carbondata/processing/newflow/row/WriteStepRow.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.apache.carbondata.processing.newflow.row;
+
+import org.apache.carbondata.core.datastore.block.SegmentProperties;
+import org.apache.carbondata.core.keygenerator.KeyGenException;
+import org.apache.carbondata.core.keygenerator.KeyGenerator;
+import org.apache.carbondata.core.scan.wrappers.ByteArrayWrapper;
+
+public class WriteStepRow {
```
ravipesala: Better name it as some utility as all methods are static here.

ravipesala:
It seems some check style issues are there, please fix it.

ravipesala:
LGTM

PR #933, opened 2017-05-21
[CARBONDATA-1075] Close Dictionary Server when application ends

kunal642:
Analysis: The dictionary server which is started when a load is done with "single_pass=true" is never closed which does not allow the carbon application to exit because it is running on the main thread.

Solution: Add a listener to the sparkContext which closes the server when application ends


kunal642:
retest this please

ravipesala:
LGTM, but it can work only if user explicitly says context.stop() otherwise won't work.

PR #937, opened 2017-05-23
[CARBONDATA-1076] Join Issue caused by dictionary and shuffle exchange

ravipesala:
When under join aggregate plan is there then there is a high chance of having same partitioning so it will generate single exchange instead of two. So in that case better push down the decoder to aggregate rather than handling at join phase.


ravipesala:
retest this please

jackylk:
LGTM. 
I verified locally. Currently CI has problem

PR #940, opened 2017-05-24
[CARBONDATA-1081] Use binarySearch to replace for clause to improve performance

mayunSaicmotor:
use binarySearch to replace for clause in  method setFilterdIndexToBitSet of ExcludeFilterExecuterImpl.java

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/scan/filter/executer/ExcludeFilterExecuterImpl.java
+++ b/core/src/main/java/org/apache/carbondata/core/scan/filter/executer/ExcludeFilterExecuterImpl.java
@@ -146,9 +146,17 @@ public class ExcludeFilterExecuterImpl implements FilterExecuter {
     BitSet bitSet = new BitSet(numerOfRows);
     bitSet.flip(0, numerOfRows);
     byte[][] filterValues = dimColumnExecuterInfo.getFilterKeys();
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/carbondata/core/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -146,9 +146,17 @@ private BitSet setFilterdIndexToBitSet(FixedLengthDimensionDataChunk dimColumnDa
     BitSet bitSet = new BitSet(numerOfRows);
     bitSet.flip(0, numerOfRows);
     byte[][] filterValues = dimColumnExecuterInfo.getFilterKeys();
-    for (int k = 0; k < filterValues.length; k++) {
+    if (filterValues.length > 1) {
```
kumarvishal09: This code will not work when No Inverted index is configured for that column while creating the column. In that case data wont be sorted and this code will five wrong result and whether filter value is one or more than one this if data is sorted we can add binary search

kumarvishal09:
retest this please

kumarvishal09:
retest this please

kumarvishal09:
retest this please

kumarvishal09:
retest this please

gvramana:
retest this please

PR #952, opened 2017-05-25
[CARBONDATA-1094] Wrong results returned by the query in case inverted index is not created on a column

manishgupta88:
Problem: Wrong results returned by the query in case inverted index is not created on a column

Fix: When inverted index does not exist for a column or column is not a sort column then
1. Block or blocklet cannot be pruned as data for that column is not sorted
2. While applying the filter linear search should be applied instead of binary search as binary search can be applied only on sorted data

Verified result
------------------
SELECT * FROM index1 WHERE city >= 'Shanghai'
+---+-----+----------+
| id| name|      city|
+---+-----+----------+
| 11|James|Washington|
| 20|Kevin| Singapore|
|  9| Mary|     Tokyo|
| 16| Paul|  Shanghai|
|  4| Sara|     Tokyo|
+---+-----+----------+

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RangeValueFilterExecuterImpl.java
+++ b/core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RangeValueFilterExecuterImpl.java
@@ -242,7 +242,7 @@ public class RangeValueFilterExecuterImpl extends ValueBasedFilterExecuterImpl {
     //                       Block Min <-----------------------> Block Max
     //         Filter Min <-----------------------------------------------> Filter Max
 
-    if (isDimensionPresentInCurrentBlock == true) {
+    if (isDimensionPresentInCurrentBlock) {
       if (((lessThanExp == true) && (
           ByteUtil.UnsafeComparer.INSTANCE.compareTo(blockMinValue, filterValues[1]) >= 0)) || (
           (lessThanEqualExp == true) && (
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RangeValueFilterExecuterImpl.java
```diff
@@ -242,7 +256,14 @@ public boolean isScanRequired(byte[] blockMinValue, byte[] blockMaxValue, byte[]
     //                       Block Min <-----------------------> Block Max
     //         Filter Min <-----------------------------------------------> Filter Max
 
-    if (isDimensionPresentInCurrentBlock == true) {
+    if (isDimensionPresentInCurrentBlock) {
+      // if a column is unsorted, min and max for that column will not be correct and hence
```
kumarvishal09: During data loading even if data is not sorted we are calculating the min max so not need to add this code, in all the filter classes only if inverted index is false we need to go iterative approach

core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RangeValueFilterExecuterImpl.java
```diff
@@ -474,80 +495,142 @@ private BitSet setFilterdIndexToBitSet(DimensionColumnDataChunk dimensionColumnD
...
+   * This method will evaluate the result for filter column based on the lower range value
+   *
+   * @param dimensionColumnDataChunk
+   * @param filterValue
+   * @param numberOfRows
+   * @return
+   */
+  private BitSet evaluateGreaterThanFilterForUnsortedColumn(
+      DimensionColumnDataChunk dimensionColumnDataChunk, byte[] filterValue, int numberOfRows) {
+    BitSet bitSet = new BitSet(numberOfRows);
```
kumarvishal09: As we are adding @NU#LL$! for No dictionary null values in this method after getting great values or greater than equal to values we need to check whether null values are present or not if it is present then we need to remove it from bitset

core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RangeValueFilterExecuterImpl.java
```diff
@@ -474,80 +495,142 @@ private BitSet setFilterdIndexToBitSet(DimensionColumnDataChunk dimensionColumnD
...
+  /**
+   * This method will evaluate the result for filter column based on the upper range value
+   *
+   * @param dimensionColumnDataChunk
+   * @param filterValue
+   * @param numberOfRows
+   * @return
+   */
+  private BitSet evaluateLessThanFilterForUnsortedColumn(
+      DimensionColumnDataChunk dimensionColumnDataChunk, byte[] filterValue, int numberOfRows) {
```
kumarvishal09: As we are adding @nu#LL$! for No dictionary null values in this method after getting less values or less than equal to values we need to check whether null values are present or not if it is present then we need to remove it from bitset

core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
```diff
@@ -88,6 +88,13 @@ private void ifDefaultValueMatchesFilter() {
   private boolean isScanRequired(byte[] blockMaxValue, byte[][] filterValues) {
     boolean isScanRequired = false;
     if (isDimensionPresentInCurrentBlock[0]) {
+      // if a column is unsorted, min and max for that column will not be correct and hence
```
kumarvishal09: During data loading even if data is not sorted we are calculating the min max so not need to add this code, in all the filter classes only if inverted index is false we need to go iterative approach. Please check in all the rowlevel range filter

core/src/main/java/org/apache/carbondata/core/scan/filter/executer/RowLevelFilterExecuterImpl.java
```diff
@@ -94,6 +94,14 @@
...
    */
   protected boolean[] isMeasurePresentInCurrentBlock;
+  /**
+   * flag to check whether inverted index exists for the filter column
+   */
+  protected boolean isInvertedIndexColumn;
+  /**
+   * flag to check whether the column is included for sorting during data load
+   */
+  protected boolean isSortColumn;
```
kumarvishal09: Please correct me isSortColumn property is for global sort, so filter does not care about whether column was part of global sort or not. Only we need to check whether column data is sorted or not so isSortColumn check and this variable is not required. Please check and handled in all the places if not required

core/src/main/java/org/apache/carbondata/core/scan/filter/executer/ExcludeFilterExecuterImpl.java
```diff
@@ -43,6 +52,11 @@ public ExcludeFilterExecuterImpl(DimColumnResolvedFilterInfo dimColEvaluatorInfo
     this.segmentProperties = segmentProperties;
     FilterUtil.prepareKeysFromSurrogates(dimColEvaluatorInfo.getFilterValues(), segmentProperties,
         dimColEvaluatorInfo.getDimension(), dimColumnExecuterInfo);
+    // check if inverted index exists for the filter column
+    isInvertedIndexColumn = CarbonUtil
+        .hasEncoding(dimColEvaluatorInfo.getDimension().getEncoder(),
+            Encoding.INVERTED_INDEX);
+    isSortColumn = dimColEvaluatorInfo.getDimension().getColumnSchema().isSortColumn();
```
manishgupta88: During data load, according to the implementation, And operator is applied between isUseInvertedIndex and isSortColumn property. If anyone is false it will not sort the column vertically and will not create inverted index for that column. But in master schema the boolean flag will still be according to the values specified by the user. Therefor we need to check false as if anyone is false inverted index will not exist for that column

kumarvishal09:
LGTM
@sounakr Please have a look on range filter changes.

kumarvishal09:
restest this please

PR #970, opened 2017-05-29
[CARBONDATA-1104] Query failure while using unsafe for query execution numeric data type column specified as sort column

manishgupta88:
Problem: Query failure while using unsafe for query execution numeric data type column specified as sort column

Analysis: When unsafe for query is enabled, while filling the data in column vector data is filled as bytes and not based on data type of the column because of which null pointer exception is thrown during result preparation in the query flow.

Fix: Convert the data based on its data type and then fill in the vector


core/src/main/java/org/apache/carbondata/core/datastore/chunk/store/impl/unsafe/UnsafeVariableLengthDimesionDataChunkStore.java
```diff
@@ -164,7 +175,26 @@ public UnsafeVariableLengthDimesionDataChunkStore(long totalSize, boolean isInve
         .equals(CarbonCommonConstants.MEMBER_DEFAULT_VAL_ARRAY, value)) {
       vector.putNull(vectorRow);
     } else {
-      vector.putBytes(vectorRow, value);
+      DataType dt = vector.getType();
```
kumarvishal09: Add switch case instead of instanceof check , please handle in other places

kumarvishal09:
LGTM

kumarvishal09:
retest this please

PR #978, opened 2017-05-31
[CARBONDATA-1109] Acquire semaphore before submit a producer in finish

watermen:
# Problem
1. We use Producer-Consumer model in the write step, have n(default value is 2 and can be configured) producers and one consumer. The task of generate last page(less than 32000) is added to thread pool at the end, but can't be guaranteed to be finished and put to BlockletDataHolder at the end. Because we have n tasks running concurrently.
2. We can't guarantee consumer get the last page from BlockletDataHolder at the end.
3. We have 2 ways to invoke `writeDataToFile`, one is the size of `DataWriterHolder` reach the size of blocklet and two is the page is the last page.

So if the last page is not be consumed at the end, we lost the page which be consumed after last page.
# Solution
Acquire semaphore before submit a producer in finish.

cc @jackylk @ravipesala @kumarvishal09

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
@@ -488,13 +488,19 @@ public class CarbonFactDataHandlerColumnar implements CarbonFactHandler {
   public void finish() throws CarbonDataWriterException {
     // still some data is present in stores if entryCount is more
     // than 0
-    producerExecutorServiceTaskList.add(producerExecutorService
-        .submit(new Producer(blockletDataHolder, dataRows, ++writerTaskSequenceCounter, true)));
-    blockletProcessingCount.incrementAndGet();
-    processedDataCount += entryCount;
-    closeWriterExecutionService(producerExecutorService);
-    processWriteTaskSubmitList(producerExecutorServiceTaskList);
-    processingComplete = true;
+    try {
+      semaphore.acquire();
+      producerExecutorServiceTaskList.add(producerExecutorService
+          .submit(new Producer(blockletDataHolder, dataRows, ++writerTaskSequenceCounter, true)));
+      blockletProcessingCount.incrementAndGet();
+      processedDataCount += entryCount;
+      closeWriterExecutionService(producerExecutorService);
+      processWriteTaskSubmitList(producerExecutorServiceTaskList);
+      processingComplete = true;
+    } catch (InterruptedException e) {
+      LOGGER.error(e, e.getMessage());
+      throw new CarbonDataWriterException(e.getMessage(), e);
+    }
   }
 
   /**


kumarvishal09:
@watermen Can u please add a testcase

watermen:
test it please

watermen:
retest this please

watermen:
retest this please

watermen:
@kumarvishal09 It is very hard to write the testcase because this case is happened occasionally.

QiangCai:
We need make sure all pages of a blocklet are in the right order to be written to carbon data file.

ravipesala:
@watermen 
This fix is contradictory to the design of producer and consumer. The producer process the data and puts in holder in sequence and consumer takes the data in the same sequence. So for the last set of data there is already a boolean flag `writeAll` inside `NodeHolder` enabled to write all the remaining data.
So this fix should not solve the issue, if it solves then there is an issue in producer/consumer design. Please provide the reproduce steps of this issue in the jira.

watermen:
@ravipesala You can reproduce this case by add some log and run loading sample.csv testcase(TestDataLoadWithFileName).
It's hard to reproducer, so I add sleep in Producer to simulate the time of processing data.
image
image
The log lost page like below:
```
00:15:42 [Thread-65]###addDataToStore
00:15:42 [Thread-65]###addDataToStore
00:15:43 [Thread-73]###Put ---> isWriteAll:false	index:1
00:15:44 [Thread-72]###Put ---> isWriteAll:false	index:0
00:15:44 [Thread-71]###Get ---> isWriteAll:false	index:0
00:15:44 [Thread-71]###Get ---> isWriteAll:false	index:1
00:15:44 [Thread-65]###addDataToStore
00:15:44 [Thread-65]###addDataToStore
00:15:44 [Thread-65]###finish
00:15:46 [Thread-72]###Put ---> isWriteAll:false	index:1
00:15:46 [Thread-72]###Put ---> isWriteAll:true		index:0
00:15:46 [Thread-71]###Get ---> isWriteAll:true		index:0// Last page is not be consumed at the end.
00:15:46 [Thread-71]###Get ---> isWriteAll:false	index:1
00:15:47 [Thread-73]###Put ---> isWriteAll:false	index:0
00:15:47 [Thread-71]###Get ---> isWriteAll:false	index:0
```

ravipesala:
@watermen Thank you for giving reproduce steps, I can reproduce it.
But the fix you have given is not right, This issue happens because of missing `semaphore.acquire()` in `finish` method. 
So solution is just adding 'semaphore.acquire()' in starting of `finish()` method solves the issue. Please update the PR.

watermen:
@ravipesala Thanks for your solution, PR updated.

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2140/

processing/src/main/java/org/apache/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -488,13 +488,19 @@ private NodeHolder processDataRows(List<CarbonRow> dataRows)
...
     // than 0
-    producerExecutorServiceTaskList.add(producerExecutorService
-        .submit(new Producer(blockletDataHolder, dataRows, ++writerTaskSequenceCounter, true)));
-    blockletProcessingCount.incrementAndGet();
-    processedDataCount += entryCount;
-    closeWriterExecutionService(producerExecutorService);
-    processWriteTaskSubmitList(producerExecutorServiceTaskList);
-    processingComplete = true;
+    try {
+      semaphore.acquire();
```
manishgupta88: @watermen .....here a check is required for entryCount > 0.....because we need to acquire a semaphore lock only if total number of rows in raw data are not exactly divisible by page size...in this case only we will have some extra rows to be process by finish method else addDataToStore method will handle all the rows....Please refer the below code snippet....

public void finish() throws CarbonDataWriterException {
    // still some data is present in stores if entryCount is more
    // than 0
    if (this.entryCount > 0) {
      try {
        semaphore.acquire();
        producerExecutorServiceTaskList.add(producerExecutorService
            .submit(new Producer(blockletDataHolder, dataRows, ++writerTaskSequenceCounter, true)));
        blockletProcessingCount.incrementAndGet();
        processedDataCount += entryCount;
      } catch (InterruptedException e) {
        LOGGER.error(e, e.getMessage());
        throw new CarbonDataWriterException(e.getMessage(), e);
      }
    }
    closeWriterExecutionService(producerExecutorService);
    processWriteTaskSubmitList(producerExecutorServiceTaskList);
    processingComplete = true;

ravipesala:
Not required any check, entry count check handled inside

ravipesala:
retest this please

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2146/

manishgupta88:
@ravipesala ....Please correct me if I am wrong....do we need to acquire a semaphore lock if we do not have record to be processed....I think if the rows are exactly divisible we will not have any extra rows to be processed and hence we do not require to acquire semaphore lock and create a new Producer object

watermen:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder/2147/

ravipesala:
@manishgupta88 It was the case in V1 and V2 formats, but in V3 format we should launch producer at finish to flush the older pages it was holding.

ravipesala:
LGTM

PR #1253, opened 2017-08-11
[CARBONDATA-1373] Enhance update performance by increasing parallelism

xuchuanyin:
# Scenario

Recently I have tested the update feature provided in Carbondata and found its poor performance.

I had a table containing about 14 million records with about 370 columns(no dictionary columns) and the data files are about 3.8 GB in total. All the data files were in one segment.

I performed an update SQL which update a column for all the records and the SQL looked like `UPDATE myTable SET (col1)=(col1+1000) WHERE TRUE`. In my environment, the update job failed with 'executor lost errors'. And I found 'spill data' related messages in the container logs.

# Analyze

I've read about the implementation of update-delete in Carbondata in ISSUE#440. The update consists a delete and an insert operation. And the error occurred during the insert operation.

After studying the code, I have found that while doing inserting, the updated records are grouped by the `segmentId`, which means all the recoreds in one segment will be processed in only one task, thus will cause task failure when the amount of input data is quite large.

# Solution

We should improve the parallelism when doing update for a segment.

I append a random key to the `segmentId` to increase the partition number before doing the insertion stage and then remove the suffix when doing the real insertion.

# Modification

+ Increase parallelism while processing one segment in update
+ Add a property to configure the parallelism
+ Clean up local files after update (previous bugs)

# Notes

I have tested in my example and the job finished in about 13 minutes successfully. The records were updated as expected.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -539,6 +539,10 @@ public final class CarbonCommonConstants {
    */
   public static final String UNDERSCORE = "_";
   /**
+   * DASH
+   */
+  public static final String DASH = "-";
+  /**
    * POINT
    */
   public static final String POINT = ".";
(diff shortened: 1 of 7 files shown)


ravipesala:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/191/

sraghunandan:
In update case, group by is not required.We can use partition by and this should solve the problem

xuchuanyin:
@sraghunandan Do you mean partitionBy a customized partitioner or  just partitionedBy segmentId？

partitionBy segmentId will forward all records of a segment to a single partition, which will still cause the problem if a screwed partition has too many data.

xuchuanyin:
@sraghunandan I've implemented by using partitionBy as you suggested in https://github.com/xuchuanyin/carbondata/pull/10/files, Is this OK?

I customize a partitioner and make the records of a segment available in different partition to avoid a 
 partition with single huge segment.

xuchuanyin:
A new PR #1261 has been raised to solve this problem in a better way.

PR #1276, opened 2017-08-21
[CARBONDATA-1401] fix duplicate issue in ListInfo

lionelcao:
add validation for listInfo

### DIFF
--- a/integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/CommonUtil.scala
+++ b/integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/CommonUtil.scala
@@ -167,7 +167,7 @@ object CommonUtil {
 
   /**
    * 1. If partitioned by clause exists, then partition_type should be defined
-   * 2. If partition_type is Hash, then number_of_partitions should be defined
+   * 2. If partition_type is Hash, then num_partitions should be defined
    * 3. If partition_type is List, then list_info should be defined
    * 4. If partition_type is Range, then range_info should be defined
    * 5. Only support single level partition for now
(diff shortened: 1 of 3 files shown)


CarbonDataQA:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/277/

integration/spark-common/src/main/scala/org/apache/carbondata/spark/util/CommonUtil.scala
```diff
@@ -369,6 +369,14 @@ object CommonUtil {
     }
   }
 
+  def validateListInfo(listInfo: List[List[String]]): Unit = {
```
xuchuanyin: simplify this by
```
val list = listInfo.flatted
if (list.length != list.distinct.size) {
  sys.error("xxx")
}
``

xuchuanyin: two more things:
1. this function is only used once, so I think there is no need to place it in `CommonUtil`.
2. is `distinct` case-sensitive or not?

lionelcao: Hi @xuchuanyin ,
Thank you for the comments. Have changed to use flatten function.
1. Put in CommonUtil not only for reusing but also keep code in one style and put similar function in one place.
2. distinct is case sensitive

ravipesala:
@lionelcao Please add a testcase for this scenario

lionelcao:
@ravipesala Test case added，please verify.

ravipesala:
LGTM

ravipesala:
I have merged but still not closed, Please close it manually

CarbonDataQA:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/325/

PR #1283, opened 2017-08-22
[WIP] Add carbon encoding example

chenerlu:
Background: This PR is related to encoding.


jackylk:
Encoding is a developer interface. If you want to add an example, I think it is better to add an example encoding like delta encoding and use it in the application.

PR #1300, opened 2017-08-30
[CARBONDATA-1422] Fixed compaction issue when decimal is present

ravipesala:


xuchuanyin:
It seems that `decimal` related modification is not shown in the changes. Do you mean `date`?

CarbonDataQA:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/331/

jackylk:
There is a table with decimal type in test("Compaction_Bug_JIRA_1422"), I think this PR is for that

ravipesala:
@xuchuanyin As @jackylk told it is related to datatype conversion in carbon, if the request is from spark then we should do datatype conversion as per spark so that we should use `SparkDataTypeConverterImpl`

jackylk:
LGTM

PR #1301, opened 2017-08-30
[CARBONDATA-1434] Remove useless class para for metastore

xuchuanyin:
The `RuntimeConfig` has not been used in Carbon*Metastore class, so remove it.


jackylk:
LGTM

PR #1306, opened 2017-08-31
[CARBONDATA-1436] optimize concurrent control for datamap

xuchuanyin:
# Scenario

`DataMapStoreManager` provides a synchronized interface `getDataMap` to retrieve a table's `TableDataMap`. It will cause performance problems in current query scenario when all the queries have to wait the former query finished.

# Analyze

We can make the concurrent control in table scope instead of global scope.

# Modification

+ Synchronized by table.

+ Use double-checked locking to reduce lock overhead

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
@@ -56,16 +56,24 @@ public final class DataMapStoreManager {
    * @param factoryClass
    * @return
    */
-  public synchronized TableDataMap getDataMap(AbsoluteTableIdentifier identifier,
+  public TableDataMap getDataMap(AbsoluteTableIdentifier identifier,
       String dataMapName, String factoryClass) {
     String table = identifier.uniqueName();
     List<TableDataMap> tableDataMaps = allDataMaps.get(table);
     TableDataMap dataMap;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -82,7 +90,6 @@ public TableDataMap createAndRegisterDataMap(AbsoluteTableIdentifier identifier,
     List<TableDataMap> tableDataMaps = allDataMaps.get(table);
     if (tableDataMaps == null) {
       tableDataMaps = new ArrayList<>();
-      allDataMaps.put(table, tableDataMaps);
```
xuchuanyin: FYI: 

should move this line to the end, otherwise in concurrent scenario, the other thread will get empty (not null) `tableMaps` in `getDataMap`, which will result in `Datamap does not exist` exception.

CarbonDataQA:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/351/

CarbonDataQA:
SDV Build Success with Spark 2.1, Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/355/

jackylk:
LGTM

PR #1325, opened 2017-09-05
[CARBONDATA-1437] Fixed bug for incorrect exception message when loading to table having property 'bucketnumber' set to 0. 

ksimar:
1. In this PR, Table creation would not be allowed, if the table property 'bucketnumber' is set to 0 in create command.
2. Added unit test case for it as well in class TableBucketingTestCase.scala


ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/533/

jackylk:
LGTM

PR #1329, opened 2017-09-06
[WIP] Min Max DataMap Example

sounakr:
Min Max DataMap Example

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/cache/CacheType.java
+++ b/core/src/main/java/org/apache/carbondata/core/cache/CacheType.java
@@ -61,6 +61,9 @@ public class CacheType<K, V> {
   public static final CacheType<TableSegmentUniqueIdentifier, SegmentTaskIndexWrapper>
       DRIVER_BLOCKLET_DATAMAP = new CacheType("driver_blocklet_datamap");
 
+  public static final CacheType<TableSegmentUniqueIdentifier, SegmentTaskIndexWrapper>
+      DRIVER_BLOCKLET_MINMAX_DATAMAP = new CacheType("driver_blocklet_minmax_datamap");
+
   /**
    * cacheName which is unique name for a cache
    */
--- a/core/src/main/java/org/apache/carbondata/core/indexstore/blockletindex/DataMapMinMax.java
+++ b/core/src/main/java/org/apache/carbondata/core/indexstore/blockletindex/DataMapMinMax.java
@@ -1,0 +1,454 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


jackylk:
Great, thanks for working on this.
My first comment is that can you move this example to example module, since "letting developer extend datamap in their own library" is one of the goal of datamap framework. We should verify the datamap framework toward this goal.

core/src/main/java/org/apache/carbondata/core/cache/CacheType.java
```diff
@@ -61,6 +61,9 @@
   public static final CacheType<TableSegmentUniqueIdentifier, SegmentTaskIndexWrapper>
       DRIVER_BLOCKLET_DATAMAP = new CacheType("driver_blocklet_datamap");
 
+  public static final CacheType<TableSegmentUniqueIdentifier, SegmentTaskIndexWrapper>
```
ravipesala: Don't add cache type, no need to handle cache in example

core/src/main/java/org/apache/carbondata/core/indexstore/blockletindex/DataMapMinMax.java
```diff
@@ -0,0 +1,454 @@
...
+import org.apache.carbondata.core.scan.filter.FilterUtil;
+import org.apache.carbondata.core.scan.filter.executer.FilterExecuter;
+import org.apache.carbondata.core.scan.filter.resolver.FilterResolverIntf;
+import org.apache.carbondata.core.util.CarbonUtil;
+import org.apache.carbondata.core.util.DataFileFooterConverter;
+
+/**
+ * Datamap implementation for blocklet.
+ */
+public class DataMapMinMax implements DataMap, Cacheable {
```
ravipesala: No need to use Unsafe in example, don't confuse users with unsafe in example. just make it as simple as possibe

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datamap/DataMapMinMaxWriter.scala
```diff
@@ -0,0 +1,163 @@
...
+
+import org.apache.carbondata.core.datamap.{DataMapStoreManager, TableDataMap}
+import org.apache.carbondata.core.datamap.dev.DataMapWriter
+import org.apache.carbondata.core.datastore.page.ColumnPage
+import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier
+import org.apache.carbondata.core.metadata.datatype.DataType
+import org.apache.carbondata.core.util.CarbonProperties
+
+
+class DataMapMinMaxWriter extends QueryTest with BeforeAndAfterAll {
```
ravipesala: Append testcase to the file name

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datamap/MinMaxDataMap.java
```diff
@@ -0,0 +1,193 @@
...
+
+  @Override public List<Blocklet> prune(FilterResolverIntf filterExp) {
+    return null;
+  }
+
+  @Override public void clear() {
+
+  }
+
+  public void updateMinMaxIndex(String filePath) throws IOException, MemoryException {
```
ravipesala: it shouldn't be read from index file. Use the Datamap writer to write example index file and read here

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/9/

PR #1341, opened 2017-09-08
[CARBONDATA-1463] CompareTest should validate result size

jackylk:
CompareTest for spark2.1 should only validate result size instead of result value, because some test case include aggregation on double column which will give different result since carbon records are sorted


ravipesala:
SDV Build Fail , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/617/

ravipesala:
LGTM

PR #1377, opened 2017-09-21
[CARBONDATA-1504] Fixed refresh of segments in datamap for update and partition

ravipesala:
Currently datamap scans complete segment every time for query execution to get all the carbonindex files. It will be slow when data/number of files are big. 
So this PR caches the content of segment and refreshes when any updation or store changes.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
@@ -93,6 +93,7 @@ public final class DataMapStoreManager {
   public TableDataMap createAndRegisterDataMap(AbsoluteTableIdentifier identifier,
       String factoryClassName, String dataMapName) {
     String table = identifier.uniqueName();
+    // Just update the segmentRefreshMap with the table if not added.
     getTableSegmentRefresher(identifier);
     List<TableDataMap> tableDataMaps = allDataMaps.get(table);
     if (tableDataMaps == null) {
@@ -131,8 +132,8 @@ public final class DataMapStoreManager {
   }
 
   /**
-   * Clear the datamap/datamaps of a mentioned datamap name and table from memory
-   * @param identifier
+   * Clear the datamap/datamaps of a table from memory
+   * @param identifier Table identifier
    */
   public void clearDataMap(AbsoluteTableIdentifier identifier) {
     List<TableDataMap> tableDataMaps = allDataMaps.get(identifier.uniqueName());
@@ -158,6 +159,9 @@ public final class DataMapStoreManager {
     return instance;
   }
 
+  /**
+   * Get the TableSegmentRefresher for the table. If not existed then add one and return.
+   */
   public TableSegmentRefresher getTableSegmentRefresher(AbsoluteTableIdentifier identifier) {
     String uniqueName = identifier.uniqueName();
     if (segmentRefreshMap.get(uniqueName) == null) {
@@ -171,8 +175,12 @@ public final class DataMapStoreManager {
    */
   public static class TableSegmentRefresher {
 
+    // This map stores the latest segment refresh time.So in case of update/delete we check the
+    // time against this map.
     private Map<String, Long> segmentRefreshTime = new HashMap<>();
 
+    // This map keeps the manual refresh entries from users. It is mainly used for partition
+    // altering.
     private Map<String, Boolean> manualSegmentRefresh = new HashMap<>();
 
     public TableSegmentRefresher(AbsoluteTableIdentifier identifier) {
@@ -195,7 +203,12 @@ public final class DataMapStoreManager {
         return true;
       }
       Long updateTimestamp = updateVO.getLatestUpdateTimestamp();
-      return updateTimestamp != null && (updateTimestamp > segmentRefreshTime.get(segmentId));
+      boolean isRefresh =
+          updateTimestamp != null && (updateTimestamp > segmentRefreshTime.get(segmentId));
+      if (isRefresh) {
+        segmentRefreshTime.remove(segmentId);
+      }
+      return isRefresh;
     }
 
     public void refreshSegments(List<String> segmentIds) {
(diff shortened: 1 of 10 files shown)


CarbonDataQA:
Build Success with Spark 1.6, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/139/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/263/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/894/

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datamap/DataMapWriterSuite.scala
```diff
@@ -41,9 +41,9 @@ class C2DataMapFactory() extends DataMapFactory {
 
   override def fireEvent(event: ChangeEvent[_]): Unit = ???
 
-  override def clear(segmentId: String): Unit = ???
+  override def clear(segmentId: String): Unit = {}
```
jackylk: why change this?

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -87,6 +93,7 @@ public TableDataMap getDataMap(AbsoluteTableIdentifier identifier,
   public TableDataMap createAndRegisterDataMap(AbsoluteTableIdentifier identifier,
       String factoryClassName, String dataMapName) {
     String table = identifier.uniqueName();
+    getTableSegmentRefresher(identifier);
```
jackylk: ignoring the return object?

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -126,20 +133,20 @@ private TableDataMap getTableDataMap(String dataMapName,
   /**
    * Clear the datamap/datamaps of a mentioned datamap name and table from memory
```
jackylk: please modify the comment also, datamap name is removed

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -151,4 +158,61 @@ public static DataMapStoreManager getInstance() {
...
+    }
+    return segmentRefreshMap.get(uniqueName);
+  }
+
+  /**
+   * Keep track of the segment refresh time.
+   */
+  public static class TableSegmentRefresher {
+
+    private Map<String, Long> segmentRefreshTime = new HashMap<>();
```
jackylk: add comment to describe the content, like map segmentId to last updated timestamp of the segment

core/src/main/java/org/apache/carbondata/core/indexstore/blockletindex/BlockletDataMapFactory.java
```diff
@@ -120,15 +121,17 @@ public void clear(String segmentId) {
...
+          dataMap.clear();
+        }
       }
     }
   }
 
   @Override
   public void clear() {
-    for (String segmentId: segmentMap.keySet()) {
+    for (String segmentId: segmentMap.keySet().toArray(new String[segmentMap.size()])) {
```
jackylk: Why this is needed?

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -292,6 +291,26 @@ private AbsoluteTableIdentifier getAbsoluteTableIdentifier(Configuration configu
       }
     }
 
+    // Clean the updated segments from memory if the update happens on segments
```
jackylk: Suggest to extract these logic as a method to refresh the datamap, and put this method in TableDataMap

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datamap/DataMapWriterSuite.scala
ravipesala: Earlier clear was not calling, but not it is called so implementation has to be provided otherwise testcase fails.

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
ravipesala: Yes, not required as we want only update segmentRefreshMap here at first time.

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
ravipesala: ok

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
ravipesala: ok

core/src/main/java/org/apache/carbondata/core/indexstore/blockletindex/BlockletDataMapFactory.java
ravipesala: It is needed as we cannot we cannot iterate on set and remove an entry at the same time. So here it is converted to array so that we can remove an entry from map in clear method

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
ravipesala: Do we need send the `SegmentUpdateDetails` detail information to `TableDataMap` ? I think it is better finding out the refreshed segments can be kept here.

CarbonDataQA:
Build Success with Spark 1.6, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/188/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/313/

jackylk:
LGTM

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/944/

PR #1522, opened 2017-11-17
[HOTFIX] change to use store path in property in testcase

jackylk:
Change to use store path in property in testcase

 - [X] Any interfaces changed?
 No
 - [X] Any backward compatibility impacted?
 No
 - [X] Document update required?
No
 - [X] Testing done
No test case is added
NA


QiangCai:
LGTM

PR #1538, opened 2017-11-20
[CARBONDATA-1779]  GenericVectorizedReader

bhavya411:

 - No  interfaces changed?
 
 - No backward compatibility impacted?
 
 - No Document update required?

 - [ Yes] Testing done
        - All Unit test cases are passing, no new unit test cases were needed as this PR implements a Generic Vectorized Reader.
        - Manual Testing completed for the same .


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/1312/

integration/presto/pom.xml
```diff
@@ -31,7 +31,7 @@
   <packaging>presto-plugin</packaging>
 
   <properties>
-    <presto.version>0.186</presto.version>
```
chenliang613: why changed presto version again ?

chenliang613:
@bhavya411  please add the detail description for this pull request.

integration/presto/pom.xml
bhavya411: There was an issue with multiple queries on Presto 0.186 so they have declare that release as unstable, they fixed that issue in 0.187 so have upgraded it to 0.187

bhavya411:
This PR removes the Spark Dependency from Presto Integration Module for using the CarbonVectorizedRecordreader, This PR consolidate  CarbonVectorizedRecordReader into one,to make it shared for all integration modules

In the earlier version of Presto Integration we were using ColumnarBatch of Spark, which is not a good practice, here we provided our own implementation of the ColumnVector and the VectorBatch to eliminate the Spark all together. This generic ColumnVector can now be used for all the integration module wherever we want to have a VectorizedReader to speed up the processing. 

There are some core module classes changed to ensure that we are using Java data types instead of Spark datatypes, Decimal being one of them.

This PR tries to limit the changes to Core module .

Newly Added Classes
1.CarbonColumnVectorImpl:This Class Implements the Interface CarbonColumnVector and provides the methods to store the data in a Vector and to retrieved the data from it as well

2.CarbonVectorBatch: This Class Creates A VectorizedRowBatch which is a set of rows, organized with each column as a CarbonVector. It is the unit of query execution, organized to minimize the cost per row and achieve high cycles-per-instruction. The major fields are public by design to allow fast and convenient access by the vectorized query execution code.

3.StructField:This class is used to pass the Schema Information to the Carbon Columnar Batch

core/src/main/java/org/apache/carbondata/core/datastore/chunk/store/impl/safe/SafeVariableLengthDimensionDataChunkStore.java
```diff
@@ -141,24 +135,25 @@ public SafeVariableLengthDimensionDataChunkStore(boolean isInvertedIndex, int nu
       // for last record
       length = (short) (this.data.length - currentDataOffset);
     }
-    DataType dt = vector.getType();
-    if ((!(dt instanceof StringType) && length == 0) || ByteUtil.UnsafeComparer.INSTANCE
+    org.apache.carbondata.core.metadata.datatype.DataType dt = vector.getType();
```
chenliang613: why do this change ?   remove import, add  the full import at here ?

PR #1660, opened 2017-12-14
[CARBONDATA-1731,CARBONDATA-1728] [BugFix] Update fails incorrectly with error

anubhav100:
This Pr Will Resolve the Following Jiras

**Jira link: https://issues.apache.org/jira/browse/CARBONDATA-1728**

**Jira link: https://issues.apache.org/jira/browse/CARBONDATA-1731**

**Problem:**

**Inside DeleteDeltaBlockDetails  class**

public boolean addBlockletDetails(DeleteDeltaBlockletDetails blocklet) {
    int index = blockletDetails.indexOf(blocklet);
    if (blockletDetails.isEmpty() || index == -1) {
      return blockletDetails.add(blocklet);
    } else {
      return blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows());
    }
  }
**In Case Of Huge Data this check
      return blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows());
is causing update to fail because deleted row is already there in deletedRows treeset and we are adding the same row again **

**Solution :  if IsRowAddedForDeletion is False don't need to add the deleted data ****

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/mutate/DeleteDeltaBlockDetails.java
+++ b/core/src/main/java/org/apache/carbondata/core/mutate/DeleteDeltaBlockDetails.java
@@ -30,15 +30,13 @@ import org.apache.carbondata.common.logging.LogServiceFactory;
 public class DeleteDeltaBlockDetails implements Serializable {
 
   private static final long serialVersionUID = 1206104914918495724L;
-
-  private List<DeleteDeltaBlockletDetails> blockletDetails;
-  private String blockName;
-
   /**
    * LOGGER
    */
   private static final LogService LOGGER =
       LogServiceFactory.getLogService(DeleteDeltaBlockDetails.class.getName());
+  private List<DeleteDeltaBlockletDetails> blockletDetails;
+  private String blockName;
 
   public DeleteDeltaBlockDetails(String blockName) {
     this.blockName = blockName;
@@ -82,9 +80,21 @@ public class DeleteDeltaBlockDetails implements Serializable {
 
   public boolean addBlocklet(String blockletId, String offset, Integer pageId) throws Exception {
     DeleteDeltaBlockletDetails blocklet = new DeleteDeltaBlockletDetails(blockletId, pageId);
+    int index = blockletDetails.indexOf(blocklet);
+
     try {
-      blocklet.addDeletedRow(CarbonUpdateUtil.getIntegerValue(offset));
-      return addBlockletDetails(blocklet);
+      boolean isRowAddedForDeletion =
+          blocklet.addDeletedRow(CarbonUpdateUtil.getIntegerValue(offset));
+      if (isRowAddedForDeletion) {
+        if (blockletDetails.isEmpty() || index == -1) {
+          return blockletDetails.add(blocklet);
+        } else {
+          blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows());
+          return true;
+        }
+      } else {
+        return false;
+      }
     } catch (Exception e) {
       if (LOGGER.isDebugEnabled()) {
         LOGGER.debug(e.getMessage());


mohammadshahidkhan:
@anubhav100  Thanks for working on this issue 
Please mention the route cause.
Is it not failing with non external database?
Please add test case for the same scenario.

anubhav100:
@mohammadshahidkhan 
1.it is failing for both external and non external database i used the description as provided in jira i am updating it now
2.i think it is not possible to add test case because it is failing for very huge data,it is failing for tpch orders table with 1gb data

mohammadshahidkhan:
LGTM +1

CarbonDataQA:
Build Success with Spark 2.2.0, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/739/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/1968/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/2286/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/2288/

anubhav100:
@mohammadshahidkhan if it looks good please merge it

chenliang613:
@anubhav100  please check this pr whether fix jira 1731 and 1728 both ?    and please update further description, why for large data, the issue happen.

anubhav100:
@chenliang613 i have updated the description you can check,yes pr has resolved both jira 1731,1728

core/src/main/java/org/apache/carbondata/core/mutate/DeleteDeltaBlockDetails.java
```diff
@@ -82,9 +80,21 @@ public boolean addBlockletDetails(DeleteDeltaBlockletDetails blocklet) {
...
-      blocklet.addDeletedRow(CarbonUpdateUtil.getIntegerValue(offset));
-      return addBlockletDetails(blocklet);
+      boolean isRowAddedForDeletion =
+          blocklet.addDeletedRow(CarbonUpdateUtil.getIntegerValue(offset));
+      if (isRowAddedForDeletion) {
+        if (blockletDetails.isEmpty() || index == -1) {
+          return blockletDetails.add(blocklet);
+        } else {
+          blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows());
+          return true;
```
chenliang613: why here should add "return true" again ?  blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows()) should already return true ?

sounakr:
@anubhav100. The issue of jira 1728 and jira 1731 is different from the proposed code. The real intention of addBlockletDetails is to add deleted rows to Delete Delta file and also to report error in case same row is deleted twice. In case addBlockletDetails finds duplicate record and return false then will get error like "Multiple input rows matched for same row.". 
In case we get these error the problem is somewhere much before like splits might have choosen duplicate blocks etc. 
Therefore we need to debug the prblem why we should be getting duplicate rows at initial stage. 
So this fix is not required as it is not going to sove the jira-1728 and 1731 problems. 
The problems specified in those jira has to be debuged more to find the exact root cause.

core/src/main/java/org/apache/carbondata/core/mutate/DeleteDeltaBlockDetails.java
anubhav100: @chenliang613 i am looking into the suggestion which is given by @sounakr that is splits might have choosen duplicate blocks

anubhav100:
@sounakr when i run  my pr with this solution both jiras doesn't get reproduced so i think it is the right way to not to add deleted rows again and again now i am looking for the solution as you provided

sounakr:
@anubhav100 . Also it seems jira 1728 and jira 1731 might have different causes. 1728 doesn't even through "Multiple input rows matched for same row.". So for this case addBlockletDetails may not be returning false for duplicate rows. As all rows are deleted please check in table update Status file also as the segment should be MARKED FOR DELETE.

anubhav100:
@sounakr no you are wrong for 1728 it is also returnning false chetan executed the query select count(*) if he executed select * it will show the same error you can check i already checked that please suggest?

anubhav100:
@sounakr when i try  to reproduce the CARBONDATA-1728 

using this script

spark.sql("DROP TABLE IF EXISTS ORDERS")

    spark.sql("DROP TABLE IF EXISTS H_ORDERS")

    spark.sql("create table if not exists ORDERS(O_ORDERDATE string,O_ORDERPRIORITY string,O_ORDERSTATUS string,O_ORDERKEY string,O_CUSTKEY string,O_TOTALPRICE double,O_CLERK string,O_SHIPPRIORITY int,O_COMMENT string) STORED BY 'org.apache.carbondata.format'")

    spark.sql("load data inpath \"hdfs://localhost:54311/orders.csv\" into table ORDERS options('DELIMITER'='|','FILEHEADER'='O_ORDERKEY,O_CUSTKEY,O_ORDERSTATUS,O_TOTALPRICE,O_ORDERDATE,O_ORDERPRIORITY,O_CLERK,O_SHIPPRIORITY,O_COMMENT')")

    spark.sql(" create table h_orders as select * from orders").show()

    spark.sql("Delete from orders a where exists (select 1 from h_orders b where b.o_ORDERKEY=a.O_ORDERKEY)")

spark.sql("SELECT O_COMMENT FROM ORDERS").show()

here are errors
17/12/15 15:47:19 ERROR DeleteExecution$: Executor task launch worker-2 Multiple input rows matched for same row.
17/12/15 15:47:20 AUDIT DeleteExecution$: [anubhav-Vostro-3559][anubhav][Thread-1]Delete data operation is failed for default.orders
17/12/15 15:47:20 ERROR DeleteExecution$: main Delete data operation is failed due to failure in creating delete delta file for segment : null block : null

sounakr:
@anubhav100  The above errors are from Delete not from Select query. Please debug jira 1728 and 1731. They might have similar or different root causes.

anubhav100:
@sounakr both are failing for same cause that is Multiple input rows matched for same row.

chenliang613:
@anubhav100 @sounakr  you guys also can use my example script to reproduce. This example simulate 7500000 data, can reproduce 1728, and this pr also can fix this issue.  please @sounakr  double check it again.

@anubhav100  i still have some queries, why need append "return true" after "blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows());" ?
---------------------------------------------------------------------------------------
package org.apache.carbondata.examples

import java.io.File
import java.text.SimpleDateFormat

import org.apache.spark.sql.SaveMode
import org.apache.spark.sql.SparkSession

import org.apache.carbondata.core.constants.CarbonCommonConstants
import org.apache.carbondata.core.util.CarbonProperties

object DataUpdateDeleteExample {

  def main(args: Array[String]) {

    // for local files
    val rootPath = new File(this.getClass.getResource("/").getPath
      + "../../../..").getCanonicalPath
    // for hdfs files
    // var rootPath = "hdfs://hdfs-host/carbon"

    var storeLocation = s"$rootPath/examples/spark2/target/store"
    var warehouse = s"$rootPath/examples/spark2/target/warehouse"
    var metastoredb = s"$rootPath/examples/spark2/target"

    import org.apache.spark.sql.CarbonSession._
    val spark = SparkSession
      .builder()
      .master("local")
      .appName("DataUpdateDeleteExample")
      .config("spark.sql.warehouse.dir", warehouse)
      .config("spark.driver.host", "localhost")
      .config("spark.sql.crossJoin.enabled", "true")
      .getOrCreateCarbonSession(storeLocation)
    spark.sparkContext.setLogLevel("WARN")

    // Specify date format based on raw data
    CarbonProperties.getInstance()
      .addProperty(CarbonCommonConstants.CARBON_DATE_FORMAT, "yyyy-MM-dd")

    import spark.implicits._
    // Drop table
    spark.sql("DROP TABLE IF EXISTS t3")


     // Simulate data and write to table t3
    var sdf = new SimpleDateFormat("yyyy-MM-dd")
    var df = spark.sparkContext.parallelize(1 to 7500000)
      .map(x => (x, new java.sql.Date(sdf.parse("2015-07-" + (x % 10 + 10)).getTime),
        "china", "aaa" + x, "phone" + 555 * x, "ASD" + (60000 + x), 14999 + x))
      .toDF("t3_id", "t3_date", "t3_country", "t3_name",
          "t3_phonetype", "t3_serialname", "t3_salary")
    df.write
      .format("carbondata")
      .option("tableName", "t3")
      .option("tempCSV", "true")
      .option("compress", "true")
      .mode(SaveMode.Overwrite)
      .save()

    // Query data again after the above update
    spark.sql("""
           SELECT * FROM t3 ORDER BY t3_id
           """).show()


    spark.sql("delete from t3 where exists (select 1 from t3)").show()
    spark.sql("""
           SELECT count(*) FROM t3
           """).show()

    // Drop table
    spark.sql("DROP TABLE IF EXISTS t3")

    spark.stop()
  }
}

anubhav100:
@chenliang reason that i return true is that when blockletDetails.get(index).addDeletedRows(blocklet.getDeletedRows())

it was adding the same row again to deletedrows treeset in   DeleteDeltaBlockletDetail class and we are validating the result of whether treeset can add the rows in it or not if these rows are duplicated it will not add them but will return me false,so what i analyze that is it is not required to check whether this treeset can add row again or not

because this check can also validate whether row is added to deleted rows treeset or not 
blocklet.addDeletedRow(CarbonUpdateUtil.getIntegerValue(offset));

so what i do is  IsRowAddedForDeletion is true that means deletion is succesfull and if that blocklet is  there  in blockletdetails simply add the deleted rows to  deletedrows treeset in DeleteDeltaBlockletDetails Class even if it is duplicated treeset will not add it 
but what
@sounakr said is correct if it is adding the same row again that means root cause is that somehow split has choosen duplicate blocks which is correct when i further debug the code if found out that in my table i have 15 lakh rows after 1408000 rows same block is picked up again by the splits which is not correct i am looking to debug it more and seeing why it happend

CarbonDataQA:
Build Failed with Spark 2.2.0, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/943/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/2172/

chenliang613:
retest this please

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/2203/

CarbonDataQA:
Build Failed with Spark 2.2.0, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/980/

chenliang613:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/2214/

CarbonDataQA:
Build Success with Spark 2.2.0, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/992/

anubhav100:
we have find the root cause and raised another pr for this issue https://github.com/apache/carbondata/pull/1719

PR #1695, opened 2017-12-20
[CARBONDATA-1920] [PrestoIntegration] Sparksql query result is not same as presto on same sql 

anubhav100:
**Sparksql query result is not same as presto on same sql** 

https://issues.apache.org/jira/browse/CARBONDATA-1920

**Problem:** In Stream Readers we are only decoding dictionary value in case our javatype is slice and it is not decimal type 

**Solution**: To solve this problem we need to do dictionary decoding in case of everydatatype if its dictionary exists because user can put any data type in dictionary_include when creating carbontable

**This Pr Include Following Changes**

1.In  stream readers functionality of dictionary decoding is provided 
2.Removed duplicate variables
3.In presto filter util case of smallint was missing in ConvertDataByType method,it is added 
4.Refactor the code of carbondatastorecreator to include dictionary encoding
5.In CarbonDictionaryDecodeReadSupport create sliceArrayBlock Only If data type is string in case of any other datatype do not create sliceArrayBlock we can decode the dictionary values using dictionary 
array the same way orc is doing on presto 
6.test cases are same but with dictionary encoding

**How Testing Is Done**

1.mvn -Pspark-2.1 clean install is passing
2.Manually checked for all the tpch presto queries
3.check all the possible queries with both dictionary_include and dictionary_exclude

### DIFF
--- a/integration/presto/src/main/java/org/apache/carbondata/presto/CarbondataPageSource.java
+++ b/integration/presto/src/main/java/org/apache/carbondata/presto/CarbondataPageSource.java
@@ -65,7 +65,8 @@ class CarbondataPageSource implements ConnectorPageSource {
   private long nanoEnd;
 
   CarbondataPageSource(RecordSet recordSet) {
-    this(requireNonNull(recordSet, "recordSet is null").getColumnTypes(), recordSet.cursor());
+    this(requireNonNull(recordSet, "recordSet is null").getColumnTypes(),
+        recordSet.cursor());
   }
 
   private CarbondataPageSource(List<Type> types, RecordCursor cursor) {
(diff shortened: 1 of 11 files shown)


anubhav100:
@chenliang can you please review this pr

chenliang613:
sure, i will review it. thanks for your contribution.

chenliang613:
Have you used carbondata 1.3.0-master code and spark 2.1 to test ?
Same issues ?

anubhav100:
@chenliang613 yes same issues are getting replicated for both 1.2.0 and 1.3.0

chenliang613:
this pr verified , looks good to me.

PR #1674, opened 2017-12-16
[CARBONDATA-1859][CARBONDATA-1861][PARTITION] Support show and drop partitions

ravipesala:
This PR depends on https://github.com/apache/carbondata/pull/1642 and https://github.com/apache/carbondata/pull/1654 and https://github.com/apache/carbondata/pull/1672

It supports show and drop partitions.

Be sure to do all of the following checklist to help us incorporate 
your contribution quickly and easily:

 - [X] Any interfaces changed?
 - [X] Any backward compatibility impacted?
   NO
 - [X] Document update required?
   Yes
 - [X] Testing done
        Tests added
       
 - [ ] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #1672, opened 2017-12-15
[CARBONDATA-1858][PARTITION] Support querying data from partition table.

ravipesala:
This PR depends on https://github.com/apache/carbondata/pull/1642 and https://github.com/apache/carbondata/pull/1654
In case of partition table first, use sessioncatalog to prune the partitions. With the partition information, datamap should read partition.map file to get the index file and corresponding blocklets to prune

Be sure to do all of the following checklist to help us incorporate 
your contribution quickly and easily:

 - [ ] Any interfaces changed?
 
 - [ ] Any backward compatibility impacted?
 
 - [ ] Document update required?

 - [ ] Testing done
        Please provide details on 
        - Whether new unit test cases have been added or why no new tests are required?
        - How it is tested? Please attach test report.
        - Is it a performance related change? Please attach the performance test report.
        - Any additional information to help reviewers in testing this change.
       
 - [ ] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #1677, opened 2017-12-19
[CARBONDATA-1860][PARTITION] Support insertoverwrite for a specific partition.

ravipesala:
This PR depends on https://github.com/apache/carbondata/pull/1672 and https://github.com/apache/carbondata/pull/1674
User should able to overwrite partition for a specific partition. Like
INSERT OVERWRITE TABLE partitioned_user
      PARTITION (country = 'US')
      SELECT * FROM another_user au 
      WHERE au.country = 'US';
In the above example, the user can overwrite only the partition(country = 'US') data. So remaining partitions data would be intact.
While overwriting a specific partition carbon should first load data to the new segment and drop that partition from all remaining segments using partition.map file.

Be sure to do all of the following checklist to help us incorporate 
your contribution quickly and easily:

 - [X] Any interfaces changed? NO
 
 - [X] Any backward compatibility impacted? NO
 
 - [X] Document update required? YES

 - [X] Testing done
       Tests added
 - [X] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #1700, opened 2017-12-21
[CARBONDATA-1860][PARTITION] Support insertoverwrite for a specific partition.

ravipesala:
This PR depends on https://github.com/apache/carbondata/pull/1672 and https://github.com/apache/carbondata/pull/1674
User should able to overwrite partition for a specific partition. Like
INSERT OVERWRITE TABLE partitioned_user
      PARTITION (country = 'US')
      SELECT * FROM another_user au 
      WHERE au.country = 'US';
In the above example, the user can overwrite only the partition(country = 'US') data. So remaining partitions data would be intact.
While overwriting a specific partition carbon should first load data to the new segment and drop that partition from all remaining segments using partition.map file.

 - [X] Any interfaces changed? NO
 
 - [X] Any backward compatibility impacted? NO
 
 - [X] Document update required? YES

 - [X] Testing done
       Tests added

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/PartitionMapFileStore.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/PartitionMapFileStore.java
@@ -228,7 +228,7 @@ public class PartitionMapFileStore {
       }
       PartitionMapper mapper = new PartitionMapper();
       mapper.setPartitionMap(partitionMap);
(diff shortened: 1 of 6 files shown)


ravipesala:
This PR is just duplicate of https://github.com/apache/carbondata/pull/1677 . The other having problem in merging to master, so i reaised new pR

PR #2211, opened 2018-04-23
[CARBONDATA-2380][DataMap] Support visible/invisible datamap for performance tuning

xuchuanyin:
Support making datamap visible/invisible through session env.
Invisible datamap will only be ignored during query and user can still
see this datmap and its data will be updated as normal.
We can specify a datamap as invisible by
'set carbon.datamap.invisible.dbName.tableName.dataMapName=false' and it
is only effective in this session.

 - [x] Any interfaces changed?
 `Add internal interface for getting visible datamaps`
 `NO`
 `NO`
 - [x] Testing done
        Please provide details on 
`Test added`
`Tested in local`
`No`
       
`Not related`

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
@@ -19,6 +19,7 @@ package org.apache.carbondata.core.datamap;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -29,6 +30,7 @@ import org.apache.carbondata.common.exceptions.sql.MalformedDataMapCommandExcept
 import org.apache.carbondata.common.exceptions.sql.NoSuchDataMapException;
 import org.apache.carbondata.common.logging.LogService;
 import org.apache.carbondata.common.logging.LogServiceFactory;
+import org.apache.carbondata.core.constants.CarbonCommonConstants;
 import org.apache.carbondata.core.datamap.dev.DataMapFactory;
 import org.apache.carbondata.core.indexstore.BlockletDetailsFetcher;
 import org.apache.carbondata.core.indexstore.SegmentPropertiesFetcher;
@@ -43,6 +45,8 @@ import org.apache.carbondata.core.mutate.SegmentUpdateDetails;
 import org.apache.carbondata.core.mutate.UpdateVO;
 import org.apache.carbondata.core.statusmanager.SegmentUpdateStatusManager;
 import org.apache.carbondata.core.util.CarbonProperties;
+import org.apache.carbondata.core.util.CarbonSessionInfo;
+import org.apache.carbondata.core.util.ThreadLocalSessionInfo;
 
 /**
  * It maintains all the DataMaps in it.
@@ -75,13 +79,12 @@ public final class DataMapStoreManager {
   }
 
   /**
-   * It gives all datamaps of type @mapType except the default datamap.
-   *
+   * It gives all visible datamaps of type @mapType except the default datamap.
    */
-  public List<TableDataMap> getAllDataMap(CarbonTable carbonTable, DataMapLevel mapType)
+  public List<TableDataMap> getAllVisibleDataMap(CarbonTable carbonTable, DataMapLevel mapType)
       throws IOException {
     List<TableDataMap> dataMaps = new ArrayList<>();
-    List<TableDataMap> tableIndices = getAllDataMap(carbonTable);
+    List<TableDataMap> tableIndices = getAllVisibleDataMap(carbonTable);
     if (tableIndices != null) {
       for (TableDataMap dataMap : tableIndices) {
         if (mapType == dataMap.getDataMapFactory().getDataMapType()) {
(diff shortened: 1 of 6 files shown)


ravipesala:
SDV Build Fail , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/4474/

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -92,6 +95,30 @@ private DataMapStoreManager() {
...
+    CarbonSessionInfo sessionInfo = ThreadLocalSessionInfo.getCarbonSessionInfo();
+    List<TableDataMap> allDataMaps = getAllDataMap(carbonTable);
+    Iterator<TableDataMap> dataMapIterator = allDataMaps.iterator();
+    while (dataMapIterator.hasNext()) {
+      TableDataMap dataMap = dataMapIterator.next();
+      String dbName = carbonTable.getDatabaseName();
+      String tableName = carbonTable.getTableName();
+      String dmName = dataMap.getDataMapSchema().getDataMapName();
+      boolean isDmVisible = sessionInfo.getSessionParams().getProperty(
+          String.format("%s%s.%s.%s", CarbonCommonConstants.CARBON_DATAMAP_VISIBLE,
```
jackylk: Are you intend to make FG Datamap invisible? If yes, better to persistent this state by setting DataMapStatus in DataMapStatusDetail.

jackylk: And I think it is better to provide a statement for Disabling the datamap instead of a set command

xuchuanyin: It's not going to disable the datamap forever. It is only intend to see how this datamap will enhance performance during the query.
As described in PR description, "Invisible datamap will only be ignored during query and user can still see this datmap and its data will be updated as normal."

It is named after MySQL document, and have the similar function as it. Please refer "MySQL now supports invisible indexes" in this link: https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html

jackylk: ok, so you mean this is temporary only? Not needed to persist it, right?

xuchuanyin: Having a statement to do this means that we need to store the related information.

I also have considered to  store the visible/invisible information in DataMapSchema/DataMapStatus related, but found that currently there is no place other than file to store them.

Besides, this feature is introduced for tuning, which means the status may change frequently. Having a file stored information may need frequently refresh & read the DataMapSchemaFile.

By using a set command here, we don't have to make the information durable. It meets our needs and can be implemented easily.

@jackylk How do you think about it?

xuchuanyin:
retest this please

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/5328/

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
xuchuanyin: yes, it is temporary, no need to persist it.

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/4163/

xuchuanyin:
retest this please

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/4182/

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
jackylk: I think it is ok, please add a @InterfaceStability.Unstable to the property key you have introduced

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/5351/

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
kumarvishal09: @xuchuanyin It is applicable for all type of data maps like pre aggregate, timeseries, etc??

xuchuanyin: @kumarvishal09 It
1. won't work on carbondata default datamap, such as DefaultBlockletDataMap/PreAggregate/Timeseries, 
2. and will work for extened datamaps, such as LuceneDataMap/BloomDataMap

The reason is that the visibility of datamap is done during blocklet pruning, it seems  PreAggregate/Timeseries won't arrive there and DefaultBlockletDataMap will always come to effective.

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/4227/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/5394/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/4533/

jackylk:
LGTM

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/datamap/CGDataMapTestCase.scala
```diff
@@ -383,6 +384,39 @@ class CGDataMapTestCase extends QueryTest with BeforeAndAfterAll {
       sql("select * from normal_test where name='n502670' and city='c2670'"))
   }
 
+  test("test invisible datamap during query") {
```
xubo245: This test case shouldn't test the invisible/visible table directly, can you add some test case for checking invisible/visible table? Are there any performance report for this?

xuchuanyin: What's your proposal?
The invisible datamap is a temporary status for a datamap. It indicates that the query will not use the datamap for now.
See Line404 for usage.

xuchuanyin: There is no need to add a performance report here.
You can explain the query before/after change the visibility of the datamap, the output should be different.

xubo245: I think we should add some validate in test case, which ensure the function correct.

xuchuanyin: Yes. We can validate it by validating the output of explain.

xuchuanyin: @xubo245 Will you work on it or just raise a jira?

PR #2294, opened 2018-05-10
[CARBONDATA-2443][SDK] Multi level complex type support for AVRO based SDK

manishgupta88:
**Problem:**
Problem inferring the complex type schema with boolean array type from the store created using SDK writer

**Analysis:**
When we create an external table and infer the schema from store created using SDK writer, the operation fails because of complex type field with boolean array dataType. This is because during schema creation by SDK writer, for array type children a child with column name val is added.
While parsing the logic to append the parent name with child column name is missing for boolean type which is causing this problem.

**Solution:**
Handle the parsing for boolean type

 - [ ] Any interfaces changed?
 No
 No
No
 - [ ] Testing done
Manually verified
NA

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/schema/table/TableSchemaBuilder.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/schema/table/TableSchemaBuilder.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
 import org.apache.carbondata.core.constants.CarbonV3DataFormatConstants;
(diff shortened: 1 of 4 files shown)


CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/4649/

core/src/main/java/org/apache/carbondata/core/metadata/schema/table/TableSchemaBuilder.java
```diff
@@ -114,12 +115,12 @@ public void setSortColumns(List<ColumnSchema> sortColumns) {
     this.sortColumns = sortColumns;
   }
 
-  public ColumnSchema addColumn(StructField field, boolean isSortColumn) {
-    return addColumn(field, null, isSortColumn, false);
+  public ColumnSchema addColumn(StructField field, AtomicInteger valIndex, boolean isSortColumn) {
```
sounakr: Do we need Atomic Integer ? This is single threaded. Later phase also even if writer becomes multi threaded, building schema and CarbonTable should remain single threaded only.

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/5806/

core/src/main/java/org/apache/carbondata/core/metadata/schema/table/TableSchemaBuilder.java
manishgupta88: AtomicInteger is not used here for multi threading purpose. It is used for incrementing a value for complex boolean array type to assign child column name

ravipesala:
LGTM

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/4858/

PR #2314, opened 2018-05-17
[CARBONDATA-2309][DataLoad] Add strategy to generate bigger carbondata files in case of small amo…

ndwangsen:
In some scenario, the input amount of loading data is small, but carbondata still distribute them to each executors (nodes) to do local-sort, thus resulting to small carbondata files generated by each executor. 

In  some extreme conditions, if the cluster is big enough or if the amount of data is small enough, the carbondata file contains only one blocklet or page.

I  think a new strategy should be introduced to solve the above problem.

The new strategy should:

be able to control the minimum amount of input data for each node
ignore data locality otherwise it may always choose a small portion of particular nodes

 - [x] Any interfaces changed?
   NO
   NO
   YES
 - [x] Testing done
        Please provide details on 
           YES
          Tested in local
        - Is it a performance related change? Please attach the performance test report.mac
          After this PR, performance is as we expected
          NO
          NO


CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/4843/

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/6002/

kumarvishal09:
@ndwangsen This pr will help during concurrent query performance as number of task will be less,  but it will impact data loading performance as it will not use all the nodes and data locality is not considered. 

But adding a property is not the correct way. It's better to expose one data load strategy interface and add multiple implementation..like distribute the data loading across nodes or your pr scenario when user wants to distribute the data based on size. Please expose one data loading strategy interface and  concrete implementation for each type (existing+ size based distribution) and user can configure or pass in load parameter which strategy they want to opt based on their use case. If user is not passing any strategy implementation, it should take default implementation.  By this way user can add some custom strategy based on their use case

kumarvishal09:
@ndwangsen What about the existing strategy???
 Can u please explain how u are planing to handle this??

ndwangsen:
@kumarvishal09  If the user specifies or default that the minimum data load of the node is less than the average data amount of each node, the existing strategy is used to handle

processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
```diff
@@ -596,20 +599,50 @@ public static Dictionary getDictionary(AbsoluteTableIdentifier absoluteTableIden
...
+      }
+      // If the average expected size for each node greater than load min size,
+      // then fall back to default strategy
+      if (iLoadMinSize * 1024 * 1024 < sizePerNode) {
+        if (CarbonProperties.getInstance().isLoadSkewedDataOptimizationEnabled()) {
+          blockAssignmentStrategy = BlockAssignmentStrategy.BLOCK_SIZE_FIRST;
+        } else {
+          blockAssignmentStrategy = BlockAssignmentStrategy.BLOCK_NUM_FIRST;
+        }
+      } else {
```
xuchuanyin: Better to add log
```
LOG.info("Specified minimum data size to load is less than the average size for each node, fallback to default strategy" + blockAssignmentStrategy);
```

processing/src/main/java/org/apache/carbondata/processing/loading/model/LoadOption.java
```diff
@@ -196,6 +196,8 @@
     optionsFinal.put("single_pass", String.valueOf(singlePass));
     optionsFinal.put("sort_scope", "local_sort");
     optionsFinal.put("sort_column_bounds", Maps.getOrDefault(options, "sort_column_bounds", ""));
+    optionsFinal.put(CarbonCommonConstants.CARBON_LOAD_MIN_SIZE_INMB,Maps.getOrDefault(options,CarbonCommonConstants.CARBON_LOAD_MIN_SIZE_INMB, CarbonCommonConstants
```
xuchuanyin: add a space after `,`

processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
```diff
@@ -575,11 +577,12 @@ public static Dictionary getDictionary(AbsoluteTableIdentifier absoluteTableIden
    * @param noOfNodesInput -1 if number of nodes has to be decided
    *                       based on block location information
    * @param blockAssignmentStrategy strategy used to assign blocks
+   * @param loadMinSize the property load_min_size_inmb specified by the user
    * @return a map that maps node to blocks
    */
   public static Map<String, List<Distributable>> nodeBlockMapping(
       List<Distributable> blockInfos, int noOfNodesInput, List<String> activeNodes,
-      BlockAssignmentStrategy blockAssignmentStrategy) {
+      BlockAssignmentStrategy blockAssignmentStrategy, String loadMinSize ) {
```
xuchuanyin: How about changing the name `loadMinSize` to `expectedMinSizePerNode`?

xuchuanyin:
@ndwangsen please resolve the conflicts and review comments

processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
ndwangsen: ok，I modify it according to your reviewed message，thanks.

kumarvishal09:
@ndwangsen I think still load is based on property....and based on enum you are checking which type of strategy it will choose , as already mentioned in earlier comment 
Please expose one data loading strategy interface and add concrete implementation for each type (existing+ size based distribution) and user can configure or pass in load parameter which strategy they want to opt based on their use case. If user is not passing any strategy implementation, it should take default implementation. By this way user can add some custom strategy based on their use case**

jackylk:
LGTM

PR #2492, opened 2018-07-11
[CARBONDATA-2733]should support scalar sub query for partition tables

sandeep-katta:
1)Multiple Projection in plan are supported
2)Scalar sub queries are supported for Carbon Tables

In this PR carbon tables will support scalar sub queries for Carbon Tables.
If multiple projections are present in the plan,carbon should support it

Highlights:
In CarbonLateDecodeStrategy need to filter the queries which does not support sub-queries
In CarbonLateDecodeStrategy if one of the projection of the plan is failed,catch the exception and 
continue
## How was this patch tested?
UT Code


ravipesala:

CarbonDataQA:

CarbonDataQA:

PR #2385, opened 2018-06-20
[CARBONDATA-2617]invalid tuple-id and block id getting formed for Non partition table

rahulforallp:
**Problem**
Invalid tuple and block id getting formed for non partition table

**Analysis**
While creating a partition table a segment file was written in the Metadata folder under table structure. This was introduced during development of partition table feature. At that time segment file was written only for partition table and it was used to distinguish between parition and non partition table in the code. But later the code was modified to write the segment file for both parititon and non partition table and the code to distinguish partition and non partition table was not modified which is causing this incorrect formation of block and tuple id.

**Fix**
Modify the logic to distinguish partitioned and non partitioned table and the same has been handled in this PR.

 - [x] Any interfaces changed? No
 
 - [x] Any backward compatibility impacted? No
 
 - [x] Document update required? No

 - [x] Testing done UT added
 - [x] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA. NA

PR #2503, opened 2018-07-13
[CARBONDATA-2734] Update is not working on the table which has segmentfile present

ravipesala:
It reverts to the PR https://github.com/apache/carbondata/pull/2385 
It fixes the IUD on the flat folder.

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
@@ -2910,13 +2910,13 @@ public final class CarbonUtil {
    * @return
    */
   public static String getBlockId(AbsoluteTableIdentifier identifier, String filePath,
-      String segmentId, boolean isTransactionalTable, boolean isPartitionTable) {
+      String segmentId, boolean isTransactionalTable, boolean hasSegmentFile) {
     String blockId;
     String blockName = filePath.substring(filePath.lastIndexOf("/") + 1, filePath.length());
     String tablePath = identifier.getTablePath();
 
     if (filePath.startsWith(tablePath)) {
-      if (!isTransactionalTable || !isPartitionTable) {
+      if (!isTransactionalTable || !hasSegmentFile) {
         blockId = "Part0" + CarbonCommonConstants.FILE_SEPARATOR + "Segment_" + segmentId
             + CarbonCommonConstants.FILE_SEPARATOR + blockName;
       } else {
--- a/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
@@ -2906,17 +2906,17 @@ public final class CarbonUtil {
    * @param identifier
    * @param filePath
    * @param segmentId
-   * @param isTransactionalTable
+   * @param isStandardTable
    * @return
    */
   public static String getBlockId(AbsoluteTableIdentifier identifier, String filePath,
-      String segmentId, boolean isTransactionalTable, boolean hasSegmentFile) {
+      String segmentId, boolean isTransactionalTable, boolean isStandardTable) {
     String blockId;
     String blockName = filePath.substring(filePath.lastIndexOf("/") + 1, filePath.length());
     String tablePath = identifier.getTablePath();
 
     if (filePath.startsWith(tablePath)) {
-      if (!isTransactionalTable || !hasSegmentFile) {
+      if (!isTransactionalTable || isStandardTable) {
         blockId = "Part0" + CarbonCommonConstants.FILE_SEPARATOR + "Segment_" + segmentId
             + CarbonCommonConstants.FILE_SEPARATOR + blockName;
       } else {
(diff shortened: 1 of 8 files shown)


ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/5822/

integration/spark2/src/main/scala/org/apache/spark/sql/execution/command/mutation/DeleteExecution.scala
```diff
@@ -134,7 +134,7 @@ object DeleteExecution {
                        groupedRows.toIterator,
                        timestamp,
                        rowCountDetailsVO,
-                       carbonTable.isHivePartitionTable)
+                       segmentFile)
```
jackylk: Is it better to pass SegmentStatus instead of string, to have better encapsulation

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/7170/

integration/spark2/src/main/scala/org/apache/spark/sql/execution/command/mutation/DeleteExecution.scala
ravipesala: Ok, But passed Segment

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/6015/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/7284/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/6051/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/5913/

manishgupta88:
@ravipesala ...Please correct me if my understanding is not correct.
SegmentFileName != null  check was introduced at the time when partition feature was developed to distinguish between partition and non partition table. Now segment file is written even if is a normal table. So using this check is not correct.
I think better we can write one generic method and pass carbonTable instance as argument to it and check for hive partiton table/supportFlatFolder/non transactional table. On this condition we can use if and else block and call the respective methods

image

ravipesala:
@manishgupta88 `SegmentFileName != null ` check is added only to distinguish legacy or not, not for partition table.  if the segment is new then segmentfile will be present so that carbondata path can be taken relative to the table path and generate blockid.  I feel it is always better to take the carbondata file path relative to the tablepath to avoid these types of checks otherwise whenever we change the path we have to update the code.

manishgupta88:
@ravipesala ...I got your point that we should always take the path relative to tablePath.
But problem with using CarbonTablePath.getShortBlockIdForPartitionTable is even in case of normal carbonTable, if you try to get the tupledId using select query, it will give some 15-20 characters extra and if we want to store this tuple ID then space will increase. So we need to find some way to shorten this ID.
For normal carbon table, TupleId value using method
**getShortBlockIdForPartitionTable** --> Fact#Part0#Segment_0/0/0-0_batchno0-0-0-1532066017077/0/0/0
**getShortBlockId** and correcting **CarbonUtil.getBlockId** --> 0/0/0-0_batchno0-0-0-1532069720380/0/0/0

ravipesala:
@manishgupta88 Ok, done the necessary changes

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder1/7363/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/6124/

ravipesala:
SDV Build Success , Please check CI http://144.76.159.231:8080/job/ApacheSDVTests/5939/

core/src/main/java/org/apache/carbondata/core/util/CarbonUtil.java
```diff
@@ -3235,4 +3235,17 @@ public boolean accept(CarbonFile file) {
...
+   * Check whether it is standard table means tablepath has Fact/Part0/Segment_ tail present with
+   * all carbon files. In other cases carbon files present directly under tablepath or
+   * tablepath/partition folder
+   * TODO Read segment file and corresponding index file to get the correct carbondata file instead
+   * of using this way.
+   * @param table
+   * @return
+   */
+  public static boolean isStandardCarbonTable(CarbonTable table) {
+    return !(table.isSupportFlatFolder() || table.isHivePartitionTable());
```
manishgupta88: Please add a check for sdk case also
CarbonTable.isTransactionalTable()

ravipesala: This check is not valid here as some places they are explicitly checking it.

manishgupta88: ok

manishgupta88:
LGTM

PR #2545, opened 2018-07-24
[CARBONDATA-2774][BloomDataMap] Exception should be thrown if expression do not satisfy bloomFilter's requirement

kevinjmh:
If query on string column with bloom index  using number as filter value, we would get wrong result. Because DDL will wrap the filter with a datatype conversion, and DatamapChooser chooses datamap by searching column deep down to all children of filter expression. However, bloom filter required the expression is simply `column =/in filterValue`.  So bloom will fail to build any querymodel and no hit blocklet, such that we will get an empty result.

To avoid getting wrong answer silently, we thrown exception instead of only giving a error log currently.

For users who wants to get correct result, can fix the SQL using corresponding data type or disable bloom datamap.

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/datamap/bloom/src/main/java/org/apache/carbondata/datamap/bloom/BloomCoarseGrainDataMap.java
+++ b/datamap/bloom/src/main/java/org/apache/carbondata/datamap/bloom/BloomCoarseGrainDataMap.java
@@ -203,7 +203,9 @@ public class BloomCoarseGrainDataMap extends CoarseGrainDataMap {
         }
         return queryModels;
       } else {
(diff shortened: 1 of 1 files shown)


xuchuanyin:
As we discussed offline, the internal reason lies in DataMapChooser.chooseDataMap or BloomDataMapFactory.prune.

If the expression can not be handled by BloomFilter, the chooser should not forwarded it to BloomFilter

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://88.99.58.216:8080/job/ApacheCarbonPRBuilder/6194/

PR #2710, opened 2018-09-11
[2875]two different threads overwriting the same carbondatafile 

shardul-cr7:
Problem : Two different threads are overwriting the same carbondata file during creation of external table.

Solution: Chances of two threads concurrently loading same carbondatafile is reduced by changing the timestamp attached in .carbondata file from millisecond to nanosecond.So chances of collision of different threads having the same file name is reduced.

 - [ ] Any interfaces changed?
 
 - [x] Testing done
         Done Manually
       

CarbonDataQA:

CarbonDataQA:

PR #2715, opened 2018-09-12
[CARBONDATA-2930] Support customize column compressor

xuchuanyin:
Support customize column compressor so that user can add their own
implementation of compressor. For customize compressor, user can
directly use its full class name while creating table or setting it to
system env.

 - [x] Any interfaces changed?
 `No`
 `NO`
`Yes, will update it later`
 - [x] Testing done
        Please provide details on 
`Tests added`
`Tested in local machine`
`NA`
`NA`
       
### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
@@ -17,19 +17,24 @@
 
 package org.apache.carbondata.core.datastore.compression;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
 import org.apache.carbondata.core.util.CarbonProperties;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.log4j.Logger;
+
 public class CompressorFactory {
+  private static final Logger LOGGER = LogServiceFactory.getLogService(
+      CompressorFactory.class.getName());
   private static final CompressorFactory COMPRESSOR_FACTORY = new CompressorFactory();
 
-  private final Map<String, SupportedCompressor> compressors = new HashMap<>();
+  private final Map<String, Compressor> allSupportedCompressors = new HashMap<>();
 
-  public enum SupportedCompressor {
+  public enum NativeSupportedCompressor {
     SNAPPY("snappy", SnappyCompressor.class),
     ZSTD("zstd", ZstdCompressor.class);
 
@@ -37,7 +42,7 @@ public class CompressorFactory {
     private Class<Compressor> compressorClass;
     private transient Compressor compressor;
 
-    SupportedCompressor(String name, Class compressorCls) {
+    NativeSupportedCompressor(String name, Class compressorCls) {
       this.name = name;
       this.compressorClass = compressorCls;
     }
@@ -54,7 +59,7 @@ public class CompressorFactory {
         try {
           this.compressor = compressorClass.newInstance();
         } catch (InstantiationException | IllegalAccessException e) {
-          throw new RuntimeException("Exception occurs while getting compressor for " + name);
+          throw new RuntimeException("Exception occurs while getting compressor for " + name, e);
         }
       }
       return this.compressor;
@@ -62,8 +67,9 @@ public class CompressorFactory {
   }
 
   private CompressorFactory() {
-    for (SupportedCompressor supportedCompressor : SupportedCompressor.values()) {
-      compressors.put(supportedCompressor.getName(), supportedCompressor);
+    for (NativeSupportedCompressor nativeSupportedCompressor : NativeSupportedCompressor.values()) {
+      allSupportedCompressors.put(nativeSupportedCompressor.getName(),
+          nativeSupportedCompressor.getCompressor());
     }
   }
 
@@ -71,6 +77,46 @@ public class CompressorFactory {
     return COMPRESSOR_FACTORY;
   }
 
+  /**
+   * register the compressor using reflection.
+   * If the class name of the compressor has already been registered before, it will return false;
+   * If the reflection fails to work or the compressor name has problem, it will throw
+   * RunTimeException; If it is registered successfully, it will return true.
+   *
+   * @param compressorClassName full class name of the compressor
+   * @return true if register successfully, false if failed.
+   */
+  private Compressor registerColumnCompressor(String compressorClassName) {
+    if (allSupportedCompressors.containsKey(compressorClassName)) {
+      return allSupportedCompressors.get(compressorClassName);
+    }
+
+    Class clazz;
+    try {
+      clazz = Class.forName(compressorClassName);
+      Object instance = clazz.newInstance();
+      if (instance instanceof Compressor) {
+        if (!((Compressor) instance).getName().equals(compressorClassName)) {
+          throw new RuntimeException(String.format("For not carbondata native supported compressor,"
+              + " the result of method getName() should be the full class name. Expected '%s',"
+              + " found '%s'", compressorClassName, ((Compressor) instance).getName()));
+        }
+        allSupportedCompressors.put(compressorClassName, (Compressor) instance);
+        LOGGER.info(
+            String.format("sucessfully register compressor %s to carbondata", compressorClassName));
+        return (Compressor) instance;
+      } else {
+        throw new RuntimeException(
+            String.format("Compressor '%s' should be a subclass of '%s'",
+                compressorClassName, Compressor.class.getCanonicalName()));
+      }
+    } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+      LOGGER.error(String.format("Failed to register compressor '%s'", compressorClassName), e);
+      throw new RuntimeException(
+          String.format("Failed to load compressor '%s', currently carbondata supports %s",
+              compressorClassName, StringUtils.join(allSupportedCompressors.keySet(), ", ")), e);
+    }
+  }
   /**
    * get the default compressor.
    * This method can only be called in data load procedure to compress column page.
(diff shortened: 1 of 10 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/261/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/430/

core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
```diff
@@ -62,15 +67,54 @@ public Compressor getCompressor() {
...
+    try {
+      clazz = Class.forName(compressorClassName);
+      Object instance = clazz.newInstance();
+      if (instance instanceof Compressor) {
+        if (!((Compressor) instance).getName().equals(compressorClassName)) {
+          throw new RuntimeException(String.format("For not carbondata native supported compressor,"
+              + " the result of method getName() should be the full class name. Expected '%s',"
+              + " found '%s'", compressorClassName, ((Compressor) instance).getName()));
+        }
+        allSupportedCompressors.put(compressorClassName, (Compressor) instance);
```
KanakaKumar: Please add a info log for new compression registered.

ravipesala:
@xuchuanyin It is better if the compressor is name taken from tableproperties.

xuchuanyin:
@ravipesala yeah, we do have a test "test create table with customize compressor" for this 'User can specify the customize compressor while creating table'.

core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
xuchuanyin: OK~

xuchuanyin:
all review comments are resolved

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8548/

xuchuanyin:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/310/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8556/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/486/

integration/spark-common-test/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithCompression.scala
```diff
@@ -42,6 +44,112 @@ case class Rcd(booleanField: Boolean, shortField: Short, intField: Int, bigintFi
     dateField: String, charField: String, floatField: Float, stringDictField: String,
     stringSortField: String, stringLocalDictField: String, longStringField: String)
 
+/**
+ * This compressor actually will not compress or decompress anything.
+ * It is used for test case of specifying customized compressor.
+ */
+class CustomizeCompressor extends Compressor {
+  override def getName: String = "org.apache.carbondata.integration.spark.testsuite.dataload.CustomizeCompressor"
```
ravipesala: I think it is better if we take shortname from compressor and store only shortname in carbondata file. And also user can use only short name in tableproperties while creating table. It should be like how Spark's fileformat interfaces uses the shortName and resolve the interfaces using the java service.

xuchuanyin: yeah, I also though about this earlier, but the problem is that we need to store the relationship between 'shortName' and 'className', which means that user has to register the compressor explicitly OR we need a property (or even a file) to manage that relationship.
In current PR's implementation, there is no need to register. It's convenient for usage.

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8776/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/709/

xuchuanyin:
@ravipesala  please review this PR

integration/spark-common-test/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithCompression.scala
ravipesala: We no need to maintain the relationship between shortname and class name. user mentions short name in the implemented class and keeps the jar in classpath. Carbondata loads all classes which are implementing compressor class and loads the class and gets the shortname from that class.
So user only should give shortname in table properties and we store only shortname in thrift as well. This is just like spark datasource , please go through it once. It would be simple for user if we provide short name.

xuchuanyin: > "Carbondata loads all classes which are implementing compressor class"

How can we figure out this? Spark's implementation still requires the user to specify the whole class name of their customize CompressionCodec.

My problem is that we do not know the full class name of the codec when we can only get the short name from the file meta.

xuchuanyin: > "Carbondata loads all classes which are implementing compressor class"

How can we figure out this? Spark's implementation still requires the user to specify the whole class name of their customize CompressionCodec.

My problem is that we do not know the full class name of the codec when we can only get the short name from the file meta.

ravipesala: I am not sure, what you are referring, but I am referring spark datasource (fileformat) design. For example, while creating carbon table we just mention `using carbon` but never mention whole implementation class name. It also saves only the shortname `carbon` in its metadata, not the whole class name. It can get the whole class name from implementation classes of `FileFormat`.

xuchuanyin: I'm refering to the implementation of 'CompressionCodec' in spark.
Due to yesterday's github problem, some content of my previous was missing. 

The corresponding code in spark is here: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/io/CompressionCodec.scala#L57-L82
For a customize compressionCodec, the line#75 use the name as the class name.

I think in your example, spark file format can handle the short name of 'carbon' because we are using CarbonSession and in this session we have registered the short name of carbon.

xuchuanyin: @ravipesala 
Hi, I've communicated with @KanakaKumar and finally figured out the background of your proposal.

For carbon spark datasource (fileformat), we can use ‘using carbon’ in creating table instead of specifying the whole class name for SparkCarbonDataSource. So here you want the compressioncodec can be implemented the same way like this -- using the short name instead of whole class name. But there are something underlying that prevent us to do that -- the very thing that I'm concerning is that we cannot get the real class name only from short name.

1. In carbon spark-datasource module, we have a file named 'org.apache.spark.sql.sources.DataSourceRegister' with content 'org.apache.spark.sql.carbondata.execution.datasources.SparkCarbonFileFormat' in the path 'resources/META-INF.services'. Spark may discover this file and load  this class, then get the short name and the whole class name for carbon datasource.

If our compressioncodec want to be implemented like this, we will have to provide extra property files and carbon will register the codecs after loading and reading this file.

2. There is another way to do this : java provides a way to find all the implementations of a specific interface. We can use this to find all the compressioncodecs and get their shortName as well as whole class name. But the problem may lie in that this procedure may take too much long time, especially we spark now have so many jars in classpath.

In conclusion, I think it's OK to accept current implementation by using the whole class name of compressioncodec. And actually spark implement it in this way too.

xuchuanyin:
retest this please

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1402/

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1188/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9454/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1410/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1196/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9460/

ravipesala:
LGTM

xuchuanyin:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1335/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9594/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1546/

jackylk:
LGTM

PR #2748, opened 2018-09-21
[CARBONDATA-2959] Added validations for TABLE OPTIONS

NamanRastogi:
create table validations are added for table OPTIONS while table is created using:
```
CREATE TABLE <tableName> ( <columns> )
USING carbon
OPTIONS ( <options> )
```
As the validations cannot be done at the table creation, it has to be done while loading data.

Validations are done for LONG_STRING_COLUMNS, DICTIONARY_INCLUDE, DICTIONARY_EXCLUDE, and NO_INVERTED_INDEX

 - [ ] Any interfaces changed?
 
 - [x] Testing done
       
### DIFF
--- a/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/CarbonSparkDataSourceUtil.scala
+++ b/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/CarbonSparkDataSourceUtil.scala
@@ -276,11 +276,6 @@ object CarbonSparkDataSourceUtil {
 
   def validateTableOptions(options: Map[String, String], schema: Schema): Unit = {
 
-    if (options.contains(CarbonCommonConstants.DICTIONARY_EXCLUDE)) {
-      LOGGER.warn("DICTIONARY_EXCLUDE option is deprecated, " +
-                  "by default string column does not use global dictionary.")
-    }
-
     val longStringColumns: Set[String] = if (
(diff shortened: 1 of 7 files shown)


CarbonDataQA:

integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/CarbonSparkDataSourceUtil.scala
```diff
@@ -269,4 +273,99 @@ object CarbonSparkDataSourceUtil {
     }
     model
   }
+
+  def validateTableOptions(options: Map[String, String], schema: Schema): Unit = {
+
+    if (options.contains(CarbonCommonConstants.DICTIONARY_EXCLUDE)) {
```
ravipesala: NO need of this validation. it does not support here. Please remove.

integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/CarbonSparkDataSourceUtil.scala
```diff
@@ -269,4 +273,94 @@ object CarbonSparkDataSourceUtil {
     }
     model
   }
+
+  def validateTableOptions(options: Map[String, String], schema: Schema): Unit = {
```
ravipesala: I don't expect any validations here. We cannot validate all the properties during load. We can do validation only for the properties which it supports. And also those validations should be inside sdk interfaces not here.

PR #2778, opened 2018-09-27
[CARBONDATA-2980][BloomDataMap] Fix bug in clearing bloomindex cache when recreating table and datamap

xuchuanyin:
clear bloomindex cache when we drop bloomfilter datamap, otherwise
after dropping and recreating the table and datamap, query will give wrong result due to the stale
cache.

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/datamap/bloom/src/main/java/org/apache/carbondata/datamap/bloom/BloomCoarseGrainDataMapFactory.java
+++ b/datamap/bloom/src/main/java/org/apache/carbondata/datamap/bloom/BloomCoarseGrainDataMapFactory.java
@@ -235,13 +235,13 @@ public class BloomCoarseGrainDataMapFactory extends DataMapFactory<CoarseGrainDa
       } else if (carbonFile.getName().equals(BloomIndexFileStore.MERGE_INPROGRESS_FILE)) {
         mergeShardInprogress = true;
       } else if (carbonFile.isDirectory()) {
-        shardPaths.add(carbonFile.getAbsolutePath());
+        shardPaths.add(FileFactory.getPath(carbonFile.getAbsolutePath()).toString());
       }
     }
     if (mergeShardFile != null && !mergeShardInprogress) {
       // should only get one shard path if mergeShard is generated successfully
       shardPaths.clear();
-      shardPaths.add(mergeShardFile.getAbsolutePath());
+      shardPaths.add(FileFactory.getPath(mergeShardFile.getAbsolutePath()).toString());
     }
     return shardPaths;
   }
@@ -349,6 +349,7 @@ public class BloomCoarseGrainDataMapFactory extends DataMapFactory<CoarseGrainDa
             FileFactory.getFileType(datamapPath));
         CarbonUtil.deleteFoldersAndFilesSilent(file);
       }
+      clear(segment);
     } catch (InterruptedException ex) {
       throw new IOException("Failed to delete datamap for segment_" + segment.getSegmentNo());
     }


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/614/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8875/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/806/

manishgupta88:
@xuchuanyin ...I think this PR changes are not correct...you are calling `lruCache.clear()` which will clear the complete LRU cache map....that means all the entries for all the tables will be cleared because LRUCache instance is only 1 per JVM.....I believe through this PR you are only trying to clear the stale entry for table being recreated
You can check the flow of drop table to understand how the dataMap cache gets cleared on dropping the table and try to incorporate Bloom dataMap cache clearing changes as per that. You can use the existing API `void invalidate(K key)` and try not to introduce a new API `void invalidateAll()`

xuchuanyin:
@manishgupta88 Oh, if it is so, then I think the interfaces of cache are confusing. Currently we use only one cache instance for all purpose like mainDataMap, bloomDataMap, dictionary, but we didn't provide a way to separate them well.
Here my initial intention is to drop all the cache for one datamap. If I call `invalidate(K key)`, I need to provide all the keys, which is inconvenient.

manishgupta88:
@xuchuanyin .....`CarbonLRUCache` instance is one per JVM but cache implementation is different for different purpose like we have separate cache provider implementation for blockDataMap and dictionary. You can check the `CacheProvider` class to get some more details on it.
Cache is a standard interface which has very generic methods. Because we are storing all the entries in LRU cache to control the memory through LRU eviction all the keys are being stored in LRU cache map. BloomDataMap should also be aware about its keys and then use the invalidate API to clean all the required keys. Check the `clear` method implementation of BlockletDataMpaFactory

xuchuanyin:
@manishgupta88 yeah, I go through the code again and finally find the root cause: the key used to load cache and the key used to clear cache is different due to the difference in path separator between linux and windows.

Please check the modification again.

datamap/bloom/src/main/java/org/apache/carbondata/datamap/bloom/BloomCoarseGrainDataMapFactory.java
```diff
@@ -235,13 +235,13 @@ public DataMapBuilder createBuilder(Segment segment, String shardName,
       } else if (carbonFile.getName().equals(BloomIndexFileStore.MERGE_INPROGRESS_FILE)) {
         mergeShardInprogress = true;
       } else if (carbonFile.isDirectory()) {
-        shardPaths.add(carbonFile.getAbsolutePath());
+        shardPaths.add(FileFactory.getPath(carbonFile.getAbsolutePath()).toString());
```
xuchuanyin: FYI：In windows env, the `carbonFile.getAbsolutePath` returns a path with path separator '\' while when we loading the cache, the path separator is '/'.

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8912/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/844/

manishgupta88:
LGTM

PR #2783, opened 2018-09-28
[CARBONDATA-2988] use unsafe for query model based on system property

xuchuanyin:
use unsafe for query model based on sytem property

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/scan/model/QueryModel.java
+++ b/core/src/main/java/org/apache/carbondata/core/scan/model/QueryModel.java
@@ -37,6 +37,7 @@ import org.apache.carbondata.core.scan.expression.UnknownExpression;
 import org.apache.carbondata.core.scan.expression.conditional.ConditionalExpression;
 import org.apache.carbondata.core.scan.filter.resolver.FilterResolverIntf;
 import org.apache.carbondata.core.stats.QueryStatisticsRecorder;
+import org.apache.carbondata.core.util.CarbonProperties;
 import org.apache.carbondata.core.util.CarbonUtil;
 import org.apache.carbondata.core.util.DataTypeConverter;
 
@@ -120,7 +121,7 @@ public class QueryModel {
   private boolean isFG;
 
   // whether to clear/free unsafe memory or not
-  private boolean freeUnsafeMemory = true;
+  private boolean freeUnsafeMemory;
 
   private boolean preFetchData = true;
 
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/629/

core/src/main/java/org/apache/carbondata/core/scan/model/QueryModel.java
```diff
@@ -120,7 +121,7 @@
   private boolean isFG;
 
   // whether to clear/free unsafe memory or not
-  private boolean freeUnsafeMemory = true;
+  private boolean freeUnsafeMemory;
```
jackylk: suggest to remove this variable, use system property whenever it is needed

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8890/

core/src/main/java/org/apache/carbondata/core/scan/model/QueryModel.java
xuchuanyin: oh, that will need many changes.
I think it's OK to get this value from system property at the beginning and pass the value to the place where it is needed, cause the system property can be modified during processing. And in this scenario, directly using system property will cause problem.

xuchuanyin:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/635/

ravipesala:
@xuchuanyin I don't what is the purpose of doing this? Please elaborate problem in PR description.
This variable is added to avoid JVM crash when freed multiple times. I am not getting why this should be configurable.

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/830/

core/src/main/java/org/apache/carbondata/core/scan/model/QueryModel.java
kumarvishal09: @xuchuanyin I think you are not clear on the purpose of this variable. 

After each task completion we are freeing all memory allocated for task to avoid memory leak based on the task id. 
In case of 'insert into' scenario, data load and query both will have same task id as they run on same thread.
so as query will finish first, it was clearing all memory (load memory also). so as data loading was is in progress it was causing jvm crash. For this issue we added this variable to avoid freeing memory from query flow in case of insert into.

Also end-user should not have this level control. What was the problem faced ? Why decided to make it a system property ?

CarbonDataQA:
Build Failed  with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8900/

xuchuanyin:
@ravipesala @kumarvishal09 We have encountered JVM crash problem.
The dump file complains
jvmcrash

Clearly it is unsafe memory related problem. Since in the previous PR, we have disabled the unsafe for query, I think in this place we should disable the unsafe accordingly.

ravipesala:
@xuchuanyin This problem is fixed in PR https://github.com/apache/carbondata/pull/2793

xuchuanyin:
OK, I'll close this PR instead

PR #2791, opened 2018-09-29
[WIP][HOTFIX]correct the exception handling in lookup relation

akashrn5:
### Problem:
 In lookup relation, we call super.lookup relation, and that can throw any exception like NoSuchTableException, Permission exception, NoSuchObjectException, but we do not care about the exception and we return false after catching exception, saying table does not exists. If there is any other exception thrown by Hive like permission exception, we should not say the table exists, we should throw the proper exception. So these change it required for any hive exception.

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
+++ b/integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
@@ -23,6 +23,7 @@ import java.net.URI
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.hadoop.fs.permission.{FsAction, FsPermission}
+import org.apache.hadoop.hive.ql.metadata.HiveException
 import org.apache.spark.sql.{CarbonDatasourceHadoopRelation, CarbonEnv, SparkSession}
 import org.apache.spark.sql.CarbonExpressions.{CarbonSubqueryAlias => SubqueryAlias}
 import org.apache.spark.sql.catalyst.TableIdentifier
@@ -208,7 +209,10 @@ class CarbonFileMetastore extends CarbonMetaStore {
     try {
       lookupRelation(tableIdentifier)(sparkSession)
     } catch {
-      case _: Exception =>
+      case ex: Exception =>
+        if (ex.getCause.isInstanceOf[HiveException]) {
+          throw ex
+        }
         return false
     }
     true


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/653/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/848/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8916/

integration/spark2/src/main/commonTo2.2And2.3/org/apache/spark/sql/hive/CarbonInMemorySessionState.scala
```diff
@@ -157,7 +158,13 @@ class InMemorySessionCatalog(
   }
 
   override def lookupRelation(name: TableIdentifier): LogicalPlan = {
-    val rtnRelation = super.lookupRelation(name)
+    val rtnRelation: LogicalPlan =
+      try {
```
sujith71955: why we need this try/catch, its super class doesn't throw any exception right, moreover we are throwing out the same exception object, can you please define the intention of this statement?

integration/spark2/src/main/commonTo2.2And2.3/org/apache/spark/sql/hive/CarbonInMemorySessionState.scala
```diff
@@ -18,6 +18,7 @@ package org.apache.spark.sql.hive
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.hive.ql.metadata.HiveException
```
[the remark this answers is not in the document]
akashrn5: scalastyle is passing for me in local too, initially also there was no space

integration/spark2/src/main/commonTo2.2And2.3/org/apache/spark/sql/hive/CarbonInMemorySessionState.scala
akashrn5: i have described in description section, may be that will explain, can you please have a look?

integration/spark2/src/main/commonTo2.2And2.3/org/apache/spark/sql/hive/CarbonInMemorySessionState.scala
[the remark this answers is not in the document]
sujith71955: I think there should be space between different system imports, like there shall be space between carbon and spark system, same shall be followed in case of hadoop also, spark is following the same styling format. please have a look, i think you can revisit the scala style guideline once again and update if required.
image

integration/spark2/src/main/commonTo2.2And2.3/org/apache/spark/sql/hive/CarbonInMemorySessionState.scala
```diff
@@ -157,7 +158,13 @@ class InMemorySessionCatalog(
   }
 
   override def lookupRelation(name: TableIdentifier): LogicalPlan = {
-    val rtnRelation = super.lookupRelation(name)
+    val rtnRelation: LogicalPlan =
+      try {
+        super.lookupRelation(name)
+      } catch {
+        case ex: Exception =>
+          throw ex
```
sujith71955: but here you are returning same type of exception object right? or am i missing anything?what about wraping this exception object with custom exception object ? or else i cannot see anything in this statement you are doing different

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
```diff
@@ -208,7 +209,10 @@ class CarbonFileMetastore extends CarbonMetaStore {
     try {
       lookupRelation(tableIdentifier)(sparkSession)
     } catch {
-      case _: Exception =>
+      case ex: Exception =>
+        if (ex.getCause.isInstanceOf[HiveException]) {
+          throw ex
+        }
```
sujith71955: i suggest if you are  skipping the exception propagation, better you log the reason why you are returning false.
this will be better for a secondary developer to locate the exact problem from a log  in some unwanted failure scenarios.

sujith71955: I think it will better to have a testcases here? i can see from your code in CarbonFileMetastore.scala  that you are only expecting a HiveException type to be thrown out, other exceptions you are ignoring.

integration/spark2/src/main/spark2.1/org/apache/spark/sql/hive/CarbonSessionState.scala
```diff
@@ -141,7 +141,13 @@ class CarbonHiveSessionCatalog(
    */
   override def lookupRelation(name: TableIdentifier,
       alias: Option[String]): LogicalPlan = {
-    val rtnRelation = super.lookupRelation(name, alias)
+    val rtnRelation =
+      try {
+        super.lookupRelation(name, alias)
+      } catch {
+        case ex: Exception =>
```
KanakaKumar: This change is not required. Catch and throwing same exception.

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
akashrn5: i checked the flow and permission errors will be thrown by Hive as HiveException and wrapped as analysis exception, and all the other exceptions will be same as i have described in description, during that time we cannot just catch exception and return false as table not exists. so i m checking as HiveException

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/691/

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
sujith71955: Didn't got your context of explanation is  for which review comment?

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
```diff
@@ -23,6 +23,7 @@ import java.net.URI
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.hadoop.fs.permission.{FsAction, FsPermission}
+import org.apache.hadoop.hive.ql.metadata.HiveException
```
sujith71955: nit: space is required, please check the rule

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/888/

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
akashrn5: here we arereturning false directly after catching exception, telling the table does not exists, this is wrong, i have check the flow , we might get hiveException(regarding permission) also if the user is not allowed to access the table, that time also we return false and we will not get the proper error.

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
akashrn5: scalastyle fails if i give the space, i have checked it

integration/spark2/src/main/scala/org/apache/spark/sql/hive/CarbonFileMetastore.scala
KanakaKumar: There are exceptions from different hierarchy incase of table not found, not all are HiveException.
Ex:- NoSuchTableException extends  AnalysisException, InvalidTableException extends HiveException.
So, I think we can refer the exceptions from Spark Code related to table not exists, not found and return false for only those exception cases. All other exceptions we can throw back to caller.

akashrn5:
i am marking it as WIP, i will check and whitelost all the other exception which can be thrown from spark

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1055/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/933/

PR #2796, opened 2018-10-03
[CARBONDATA-2991]NegativeArraySizeException during query execution

BJangir:
Issue :- During Query Execution sometime NegativeArraySizeException  Exception in Some Tasks . And sometime Executor is lost (JVM crash)

Root Cause :- It is because existing memoryblock is removed while it was in-use. This happened  because duplicate taskid generated. Sometime freed same memory addresses are assigned to another task which will initialize memory block to0 and this cause NegativeSizeArrayException whereas sometime freed memory will not be used any task of executor process but running task will try to access it and as that address is not part of process so JVM crash will happen.

Solution :-  Change taskID generation to UUID based instead of System.nanoTime()

 - [ ] Any interfaces changed?
 NA
 NA
NA
 - [ ] Testing done
        Please provide details on 
     
Manual Testing Done  
NA

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/ThreadLocalTaskInfo.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/ThreadLocalTaskInfo.java
@@ -16,6 +16,7 @@
  */
 package org.apache.carbondata.core.util;
 
+
 /**
  * Class to keep all the thread local variable for task
  */
@@ -30,7 +31,7 @@ public class ThreadLocalTaskInfo {
   public static CarbonTaskInfo getCarbonTaskInfo() {
     if (null == threadLocal.get()) {
       CarbonTaskInfo carbonTaskInfo = new CarbonTaskInfo();
-      carbonTaskInfo.setTaskId(System.nanoTime());
+      carbonTaskInfo.setTaskId(CarbonUtil.generateUUID());
       ThreadLocalTaskInfo.setCarbonTaskInfo(carbonTaskInfo);
     }
     return threadLocal.get();
--- a/core/src/main/java/org/apache/carbondata/core/datastore/chunk/store/impl/unsafe/UnsafeAbstractDimensionDataChunkStore.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/chunk/store/impl/unsafe/UnsafeAbstractDimensionDataChunkStore.java
@@ -62,7 +62,7 @@ public abstract class UnsafeAbstractDimensionDataChunkStore implements Dimension
    */
   protected boolean isMemoryOccupied;
 
-  private final long taskId = ThreadLocalTaskInfo.getCarbonTaskInfo().getTaskId();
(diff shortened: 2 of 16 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/690/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8955/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/887/

sujith71955:
@BJangir @jackylk @chenliang613 @ravipesala 
I think UUID.randomUUID().toString() has some performance bottle necks even though its negligible, as it internally uses SecureRandom API which is a synchronized method. 
if possible explore more on this as our system is more performance driven.
What about using JUG?
There is also a time bound uuid generation which seems to work lot faster compare to normal UUID generation supported by java.
Following link will give you a fair idea regarding implementation and performance comparison.
http://www.dcalabresi.com/blog/java/generate-java-uuid-performance/

core/src/main/java/org/apache/carbondata/core/util/ThreadLocalTaskInfo.java
```diff
@@ -30,7 +32,7 @@ public static void setCarbonTaskInfo(CarbonTaskInfo carbonTaskInfo) {
   public static CarbonTaskInfo getCarbonTaskInfo() {
     if (null == threadLocal.get()) {
       CarbonTaskInfo carbonTaskInfo = new CarbonTaskInfo();
-      carbonTaskInfo.setTaskId(System.nanoTime());
+      carbonTaskInfo.setTaskId(UUID.randomUUID().toString());
```
ravipesala: Please move the generation of UUID to some UUID to some common utility so all methods can use same. In future if we need to change we can change at one place

ravipesala:
@sujith71955 Carbondata usage of UUID is very minimal and also it is not safe to depend on another third party dependency (non tested) without much value add.

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/712/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/713/

ravipesala:
LGTM

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/8977/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/909/

PR #2816, opened 2018-10-16
 [CARBONDATA-3003] Support read batch row in CSDK

xubo245:
[CARBONDATA-3003] Support read batch row in CSDK
    1. Support read batch rows in SDK
    2. Support read batch rows in CSDK
    3. improve CSDK read performance

For SDK batch read:
readNextBatchRow:
```
total lines is 200100000, build time is 10.434133262 s, 	total read time is 167.567157044 s, 	average speed is 1194148.0868321797records/s.
```
readNextRow:
```
total lines is 200100000, build time is 15.775965656 s, 	total read time is 183.312544655 s, 	average speed is 1091578.322567037records/s.
```
read batch row is faster 9.4% than readCarbonRow( one by one)

For CSDK:

Test next Row Performance: in local

```

Test next Row Performance:
build time is: 6.119777 s

100000: time is 0.278082 s, speed is 359606.159334 records/s  from_email10144phillip.allen@enron.com	to_email69149buck.buckner@honeywell.com	from_to	<5164240.1075855667637.JavaMail.evans@thyme>	1538015558000000	971703720000000	
200000: time is 0.301743 s, speed is 331407.853703 records/s  from_email10324veronica.espinoza@enron.com	to_email44094cynthia.franklin@enron.com	from_to	<14154714.1075858633174.JavaMail.evans@thyme>	1538015558000000	1003768608000000	
300000: time is 0.180397 s, speed is 554332.943452 records/s  from_email10545store-news@amazon.com	to_email59755pallen@enron.com	from_to	<5977904.1075858636257.JavaMail.evans@thyme>	1538015558000000	1004057196000000	
......
199800000: time is 0.144240 s, speed is 693288.962840 records/s  from_email99414kathie.grabstald@enron.com	to_email97812cowan'.'beth@enron.com	from_cc	<25801832.1075858634648.JavaMail.evans@thyme>	1538015558000000	1003926153000000	
199900000: time is 0.139121 s, speed is 718798.743540 records/s  from_email99622kathie.grabstald@enron.com	to_email39726jean.mrha@enron.com	from_to	<30070113.1075858637581.JavaMail.evans@thyme>	1538015558000000	1004364487000000	
200000000: time is 0.139053 s, speed is 719150.252062 records/s  from_email99837kathie.grabstald@enron.com	to_email40090e.taylor@enron.com	from_bcc	<20754442.1075858633286.JavaMail.evans@thyme>	1538015558000000	1003755042000000	
[...]

PR #2850, opened 2018-10-24
[CARBONDATA-3056] Added concurrent reading through SDK

NamanRastogi:
Added an API  _CarbonReader.split_ to enable concurrent reading of carbondata files through SDK.
```java
List<CarbonReader> multipleReaders = CarbonReader.split(maxSplits)
```

For detailed information on how to use this API for concurrent reading, please refer **ConcurrentSdkReaderTest.java**

## Performance Metrics:

| | configured table block: 1 MB | configured table block size: 10 MB | configured table block: 100 MB |
| --- | --- | --- | --- |
| **# rows: 1e6**<br>**Store: 7.6 MB** | # files generated: 11<br><br>Sequential Read: 274 ms<br>Parallel Read: 123 ms | # files generated: 1<br><br>Sequential Read: 247 ms<br>Parallel Read: 248 ms | # files generated: 1<br><br>Sequential Read: 252 ms<br>Parallel Read: 254 ms |
| **# rows: 1e7**<br>**Store: 78 MB** | # files generated: 104<br><br>Sequential Read: 2685 ms<br>Parallel Read: 1230 ms | # files generated: 9<br><br>Sequential Read: 2499 ms<br>Parallel Read: 1357 ms | # files generated: 1<br><br>Sequential Read: 2527 ms<br>Parallel Read: 2597 ms |
| **# rows: 1e8**<br>**Store: 865 MB** | | # files generated: 95<br><br>Sequential Read: 27069 ms<br>Parallel Read: 16082 ms | # files generated: 15<br><br>Sequential Read: 25841 ms<br>Parallel Read: 13256 ms |

 - [ ] Any interfaces changed?
 - [x] Any backward compatibility impacted: No
 - [x] Testing done
        - New unit test case have been added.


CarbonDataQA:

xuchuanyin:
@NamanRastogi I think we can further optimize this function.

1. We can enable the parallel reading and set the parallelism while creating a CarbonReader;
2. Inside CarbonReader, we handle the concurrent processing;
3. The interfaces for CarbonReader should be kept the same as before, there is no need to add more interfaces. By calling hasNext or next, user can get the next record and will not care about which RecordReader does this record belong to.

The user interface looks like below:
```
CarbonReader reader = CarbonReader.builder(dataDir).parallelism(3).build();
while (reader.hasNext()) {
  reader.next();
}
reader.close();
```
To keep it simple, by default the parallelism can be 1 which means we will process each RecordReader one by one. Setting this parallelism to a higher value means that we will go process the RecordReaders in a thread pool with size 3.

NamanRastogi:
@xuchuanyin 

Since `CarbonReader` is iterator based, we can only read a line when user wants. So, even if we keep the parallelism internally before the build method, it will still read the file row by row (even though they are in different threads) and reading multiple rows will still be accessed sequentially. Please notice the test file **ConcurrentSdkReaderTest.java**, the reading is happening in the thread itself, and multiple threads are reading the files (line-by-line) concurrently.

The actual reading is happening inside the `CarbonReader.readNext()` method, as long as that is sequential, the actual reading performance is not going to get better, so we have to have `CarbonReader.readNext()` inside different threads,.

xuchuanyin:
emm, but in your implementation, most of the work has to be done by the user (multi-thread handling). CarbonData itself only split the input data and return multiple readers. If this is the solution, why not just tell the user to generate multiple CarbonReaders by passing only part of the input dir each time they create the reader?

Addition to my proposal, I think we can add a buffer for the records. When `CarbonReader.next` is called, we can retrieve the record from the buffer and fill the buffer asynchronously. When`CarbonReader.hasNext` is called, we can first detect this from the buffer, if it is empty, we will then detect this from the recordReader and fill the buffer asynchronously.

NamanRastogi:
@xuchuanyin 
Using this API the overhead of reading concurrently is on the consumer of CarbonReader. So yes, what you said what you said is right. But this API was made to consider this case only when a user wants to have more manual control over the concurrent reading. Consider the scenario when the user wants to read different files in different machines. If we take care of the concurrent reading internally (using a buffer as you suggested), the user cannot do that.

Another thing that you mentioned was user providing the location of _carbondata_ files in the SDK reader itself, and that does not work in the current implementation of SDK.

And for faster reading, there is one more pull request: [2816](https://github.com/apache/carbondata/pull/2816), it supports batch reading for a better reading performance. It is still not concurrent reading, but better than iterator based row-by-row reading.

store/sdk/src/test/java/org/apache/carbondata/sdk/file/ConcurrentSdkReaderTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.*;
+
+/**
+ * multi-thread Test suite for {@link CarbonReader}
+ */
+public class ConcurrentSdkReaderTest extends TestCase {
+
+  private static final String dataDir = "./testReadFiles";
+
+  public void cleanTestData() {
```
xubo245: you can add @Before or @After got it. especially there are many test case

store/sdk/src/test/java/org/apache/carbondata/sdk/file/ConcurrentSdkReaderTest.java
```diff
@@ -0,0 +1,155 @@
...
+      count = 0;
+      long start = System.currentTimeMillis();
+      for (CarbonReader reader_i : multipleReaders) {
+        results.add(executorService.submit(new ReadLogic(reader_i)));
+      }
+      for (Future result_i : results) {
+        count += (long) result_i.get();
+      }
+      long end = System.currentTimeMillis();
+      System.out.println("[Parallel read] Time:" + (end - start));
```
xubo245: Please add unit for it, such as ms

xubo245:
@NamanRastogi Hi, customer required carbon provide the same order between using one thread to read and use multiple threads to read data.

NamanRastogi:
Yes, data coming from one file will always be in order. Please check the `split` method, it splits the list of CarbonRecordReader into multiple CarbonReader s.

Suppose there are 10 carbondata files, and the user wants to get 3 splits, so he will get a list like this:

NamanRastogi:
Please check the split method, it splits the list of `CarbonRecordReader` into multiple `CarbonReader`s. It does not jumble the order of `CarbonRecordReader`, it still keeps them sequential.

Suppose there are 10 *carbondata* files and thus 10 `CarbonRecordReader` in `CarbonReader.readers` object and the user wants to get 3 splits, so he will get a list like this:
```java
CarbonReader reader = CarbonReader.builder(dataDir).build();
List<CarbonReader> multipleReaders = reader.split(3);
```
And the indices of `CarbonRecordReader`s in `multipleReaders` will be like:
`multipleReaders.get(0).readers` points to {0,1,2,3} indices of *carbondata* files
`multipleReaders.get(1).readers` points to {4,5,6} indices of *carbondata* files
`multipleReaders.get(2).readers` points to {7,8,9} indices of *carbondata* files

Now, if you read the rows like following code, the rows will still be in order.
```java
for (CarbonReader reader_i : multipleReaders) {
    reader_i.readNextRow();
}
```

Earlier, you were getting data from 5th `CarbonRecordReader` only after you have exhausted the 4th. But now, you are getting it earlier, maybe even before 0th. So the user has to make sure he consumes it after he has used up the 4th file if order is important for him/her, otherwise he/she can use it earlier also if order is not important. So, for example to count the total no. of rows, user does not need the original order.

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
```diff
@@ -114,6 +117,43 @@ public static CarbonReaderBuilder builder(String tablePath) {
     return builder(tablePath, tableName);
   }
 
+  /**
+   * Return a new list of {@link CarbonReader} objects
+   *
```
ajantha-bhat: Add a clear description, mention what happens if splits greater than the number of files and what happens if splits are lesser than the number of files

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
```diff
@@ -114,6 +117,43 @@ public static CarbonReaderBuilder builder(String tablePath) {
...
+
+    List<CarbonReader> carbonReaders = new ArrayList<>();
+
+    // If maxSplits < readers.size
+    // Split the reader into maxSplits splits with each
+    // element contains >= 1 CarbonRecordReader objects
+    if (maxSplits < this.readers.size()) {
+      for (int i = 0; i < maxSplits; ++i) {
+        carbonReaders.add(new CarbonReader<>(this.readers
+            .subList((int) Math.ceil((float) (i * this.readers.size()) / maxSplits),
```
ajantha-bhat: this is constant, do this outside loop and use it each time

store/sdk/src/test/java/org/apache/carbondata/sdk/file/ConcurrentSdkReaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+      }
+      writer.close();
+    } catch (Exception e) {
+      e.printStackTrace();
+      Assert.fail(e.getMessage());
+    }
+  }
+
+  @Test public void testReadParallely() throws IOException, InterruptedException {
+    long numRows = 10000000;
```
ajantha-bhat: We must not add huge record test case in UT. PR builder time for all the PR will affect by this. Locally test with huge data but update UT with fewer rows. say 10 rows.

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
NamanRastogi: Done!

shardul-cr7:
retest this please

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
```diff
@@ -114,6 +115,57 @@ public static CarbonReaderBuilder builder(String tablePath) {
...
+   * through >= 1 file.
+   *
+   * If the no. of files is less than maxSplits, then return list of
+   * CarbonReader with size as the no. of files, with each CarbonReader
+   * iterating through exactly one file
+   *
+   * @param maxSplits: Int
+   * @return list of {@link CarbonReader} objects
+   */
+  public List<CarbonReader> split(int maxSplits) throws IOException {
```
ravipesala: I feel this method should be moved to builder. Add another method in builder `build(int splits)` and return List of readers.

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
```diff
@@ -114,6 +115,57 @@ public static CarbonReaderBuilder builder(String tablePath) {
...
+    validateReader();
+    if (maxSplits < 1) {
+      throw new RuntimeException(
+          this.getClass().getSimpleName() + ".split: maxSplits must be positive");
+    }
+
+    List<CarbonReader> carbonReaders = new ArrayList<>();
+
+    if (maxSplits < this.readers.size()) {
```
ravipesala: Add UT only to this method to make sure splits happen correctly with multiple splits combinations and readers size,

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
ajantha-bhat: @ravipesala : Adding to builder will break builder pattern, recently we removed arguments from build() and make it as separate API for SDK writer. Reader also followed same.

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReader.java
kunal642: @ravipesala Let us add test cases in a separate PR. would it be okay?

ravipesala:
retest this please

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1244/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9509/

kunal642:
LGTM

PR #2900, opened 2018-11-05
[CARBONDATA-3078] Disable explain collector for count star query without filter

kevinjmh:
An issue is found about count star query without filter in explain command. It is a special case. It uses different plan. 
Considering
no useful information about block/blocklet pruning for count star query without filter, so disable explain collector and avoid the exception in https://issues.apache.org/jira/browse/CARBONDATA-3078

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
@@ -43,6 +43,7 @@ import org.apache.carbondata.core.mutate.CarbonUpdateUtil;
 import org.apache.carbondata.core.mutate.SegmentUpdateDetails;
 import org.apache.carbondata.core.mutate.UpdateVO;
 import org.apache.carbondata.core.mutate.data.BlockMappingVO;
+import org.apache.carbondata.core.profiler.ExplainCollector;
 import org.apache.carbondata.core.readcommitter.LatestFilesReadCommittedScope;
 import org.apache.carbondata.core.readcommitter.ReadCommittedScope;
 import org.apache.carbondata.core.readcommitter.TableStatusReadCommittedScope;
@@ -575,6 +576,14 @@ public class CarbonTableInputFormat<T> extends CarbonInputFormat<T> {
    */
   public BlockMappingVO getBlockRowCount(Job job, CarbonTable table,
       List<PartitionSpec> partitions) throws IOException {
+    // Normal query flow goes to CarbonInputFormat#getPrunedBlocklets and initialize the
+    // pruning info for table we queried. But here count star query without filter uses a different
+    // query plan, and no pruning info is initialized. When it calls default data map to
+    // prune(with a null filter), exception will occur during setting pruning info.
+    // Considering no useful information about block/blocklet pruning for such query
+    // (actually no pruning), so we disable explain collector here
+    ExplainCollector.remove();
+
     AbsoluteTableIdentifier identifier = table.getAbsoluteTableIdentifier();
     TableDataMap blockletMap = DataMapStoreManager.getInstance().getDefaultDataMap(table);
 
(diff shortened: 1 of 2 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1276/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1495/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9545/

kevinjmh:
retest this please

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -575,6 +576,8 @@ private BitSet setMatchedPartitions(String partitionIds, Expression filter,
    */
   public BlockMappingVO getBlockRowCount(Job job, CarbonTable table,
       List<PartitionSpec> partitions) throws IOException {
+    // no useful information for count star query without filter, so disable explain collector
+    ExplainCollector.remove();
```
xuchuanyin: I think this modification just try to avoid the problem but don't actually solve the problem.
Can you explain what is the root cause of that problem?

kevinjmh: You are right. 
Normal query flow goes to `CarbonInputFormat#getPrunedBlocklets` and initialize the pruning info for table we queried.  But count star query without filter use a different query plan, it does not go into that method, so no pruning info does not init. When it calls default data map to prune(with a null filter), exception will occur during settingg pruning info.

One solution is to init the pruning info for this type of query in mrthod `getBlockRowCount`. But considering
no useful information about block/blocklet pruning for such query(actually no pruning), I choose to disable the expalin collector instead.

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1294/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1507/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9555/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
xuchuanyin: That's the weird part -- We are trying to remove the pruning collector even the pruning info is not initialized.
I think you can add a flag for the collector to identify whether it is initialized. And this flag is used where carbon what to record the info. If you are planing to work like this, please add a comment for the scenario of this variable.

kevinjmh: No remove. Its implementation is disable.

xuchuanyin: oh... I understand.
The current implementation of pruning collector may has bugs. Based on the current implementation, your modification is OK...

xuchuanyin: please add comments for your modification in the code for better understanding

kevinjmh: OK

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1304/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1519/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9568/

xuchuanyin:
LGTM

PR #2863, opened 2018-10-27
[CARBONDATA-3112] Optimise decompressing while filling the vector during conversion of primitive types

ravipesala:
Following optimizations done in the PR.
1. Optimise decompressing while filling the vector during conversion of primitive types. It avoids creating an intermediate buffer while decompression.
2. Refactor the global dictionary decoder codegen to minimise the amount of code generated to reduce the time.
3. Disable lazy load for full scan queries as it is unnecessary.
4. Refactor the compressor interface and created Abstract class. All primitive datatype conversions now happen in little_endian as snappy does that conversion while compressing. So it might break the compatibility for ZSTD for the last version.

Be sure to do all of the following checklist to help us incorporate 
your contribution quickly and easily:

 - [ ] Any interfaces changed?
 
 - [ ] Any backward compatibility impacted?
 
 - [ ] Document update required?

 - [ ] Testing done
        Please provide details on 
        - Whether new unit test cases have been added or why no new tests are required?
        - How it is tested? Please attach test report.
        - Is it a performance related change? Please attach the performance test report.
        - Any additional information to help reviewers in testing this change.
       
 - [ ] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #2902, opened 2018-11-06
[CARBONDATA-3083] Fixed data mismatch issue after update

kunal642:
**Problem:** When filling a columnPage directly to vector, we are skipping the deleted rows based on BitSet value. Now consider a situation where the 6th row is Null i.e BitSet(6) and the 3rd row is marked as deleted i.e BitSet(3). 
During filling of the vector we will skip the 3rd row and the final vector will have 1 less row(total 5 rows) than the columnPage. 
While reading we will only read 5 rows and when trying to set the 6th row as null we will end up making the wrong row as null.

**Solution:** Check if the vector has inverted index or deleted rows. If it has then dont blindly copy the array using System.arrayCopy instead iterated over the values check for null and insert the appropriate values.

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1298/

core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -1734,7 +1734,7 @@ private CarbonCommonConstants() {
   public static final String CARBON_PUSH_ROW_FILTERS_FOR_VECTOR =
       "carbon.push.rowfilters.for.vector";
 
-  public static final String CARBON_PUSH_ROW_FILTERS_FOR_VECTOR_DEFAULT = "false";
+  public static final String CARBON_PUSH_ROW_FILTERS_FOR_VECTOR_DEFAULT = "true";
```
manishgupta88: Any specific reason for changing the default value?

integration/spark-common-test/src/test/scala/org/apache/carbondata/spark/testsuite/iud/UpdateCarbonTableTestCase.scala
```diff
@@ -772,12 +772,33 @@ class UpdateCarbonTableTestCase extends QueryTest with BeforeAndAfterAll {
...
+    sql(
+      s"""LOAD DATA LOCAL INPATH '$resourcesPath/IUD/negativevalue.csv' INTO table iud
+         |.dest33_part options('header'='false')""".stripMargin)
+    sql(
+      """update iud.dest33_part d set (c1) = (5) where d.c1 = 0""".stripMargin).show()
+    checkAnswer(sql("select c3 from iud.dest33_part"), Seq(Row(-300), Row(0), Row(-200), Row(700)
+      , Row(100), Row(-100), Row(null)))
+    sql("""drop table if exists iud.dest33_part""")
+    CarbonProperties.getInstance().addProperty(CarbonCommonConstants
+      .CARBON_PUSH_ROW_FILTERS_FOR_VECTOR, "true")
```
manishgupta88: After test case completion we should set the default value for `CARBON_PUSH_ROW_FILTERS_FOR_VECTOR`?...default property is false so I think at the start of test case no need to modify the property value

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1512/

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9559/

core/src/main/java/org/apache/carbondata/core/datastore/page/encoding/compress/DirectCompressCodec.java
```diff
@@ -257,7 +265,13 @@ private void fillVector(ColumnPage columnPage, CarbonColumnVector vector,
       } else if (pageDataType == DataTypes.SHORT) {
         short[] shortData = columnPage.getShortPage();
         if (vectorDataType == DataTypes.SHORT) {
-          vector.putShorts(0, pageSize, shortData, 0);
+          if (isUnderlyingVectorPresent) {
+            for (int i = 0; i < pageSize; i++) {
+              vector.putShort(i, shortData[i]);
+            }
+          } else {
+            vector.putShorts(0, pageSize, shortData, 0);
```
manishgupta88: I think using `putShorts/putFloats` is common and unavoidable. In future also any new encoding class can make use of these method and then again the same problem can occur. Is it feasible to modify the vector classes implementation methods itself just like an example below
`public void putShorts(int rowId, int count, short[] src, int srcIndex) {
    for (int i = srcIndex; i < count; i++) {
      putShort(rowId++, src[i]);
    }
  }`
This way it will be better

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/1303/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/1516/

ravipesala:
@kunal642 Please check PR https://github.com/apache/carbondata/pull/2863 . This issue should not happen there. Please verify once

CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9564/

PR #2958, opened 2018-11-27
[CARBONDATA-3136] Fix JVM crash with preaggregate datamap when average of decimal column is taken with orderby.  

ajantha-bhat:
problem:  JVM crash with preaggregate datamap when average of decimal column is taken with orderby.  

cause: When preparing plan with preaggregate datamap, decimal is cast to double in average expression. This was leading to JVM crash in spark as we were filling with wrong precision (callstack mentioned in JIRA) 

solution: division result of average, should be casted to decimal instead of double for decimal datatype.

 - [ ] Testing done.   
       yes, added UT       


CarbonDataQA:
Build Success with Spark 2.3.1, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/9817/

ravipesala:
LGTM

PR #3028, opened 2018-12-27
[CARBONDATA-3205]Fix Get Local Dictionary for empty Array of String

Indhumathi27:
Problem:
In case of Array data type, if the data is empty then the Local dictionary entry for the data will not be created.
So, while querying the data, dictionary value read from the local dictionary will go wrong.
In case we have 3 rows of data with the first row as empty data, then the number of local dictionary entries will be 2. Reading back the dictionary values from the local dictionary while query throws ArrayIndexOutOfBoundException.

Solution:
Check for the dictionary value presence, fill default value only if dictionary values exists.

 - [ ] Any interfaces changed?
 
 - [x] Testing done
      Added a test case
       

core/src/main/java/org/apache/carbondata/core/datastore/chunk/impl/DimensionRawColumnChunk.java
```diff
@@ -197,6 +197,10 @@ public static CarbonDictionary getDictionary(LocalDictionaryChunk localDictionar
           localDictionaryChunk.getDictionary_data().length);
       BitSet usedDictionary = BitSet.valueOf(compressor.unCompressByte(
           localDictionaryChunk.getDictionary_values()));
+      // if dictionary is empty, return null
+      if (usedDictionary.isEmpty()) {
```
qiuchenjian: DataFile class call this method，but it use the return value directly, it will throw NullException. i think it need be protected
Please see DataFile line 360,
CarbonDictionary dictionary = DimensionRawColumnChunk.getDictionary(dictionaryChunk, comp);
blockletDictionaryEntries = dictionary.getDictionaryActualSize();

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2255/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2048/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10300/

ravipesala:
LGTM

PR #3039, opened 2018-12-31
[CARBONDATA-3217] Optimize implicit filter expression performance by removing extra serialization

manishgupta88:
Fixed performance issue for Implicit filter column
1. Removed serialization all the implicit filter values in each task. Instead serialized values only for the blocks going to particular task
2. Removed 2 times deserialization of implicit filter values in executor for each task. 1 time is sufficient

 - [ ] Any interfaces changed?
 No
 No
No
 - [ ] Testing done
Added UT
NA

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/scan/expression/conditional/ImplicitExpression.java
+++ b/core/src/main/java/org/apache/carbondata/core/scan/expression/conditional/ImplicitExpression.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.carbondata.core.scan.expression.conditional;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.carbondata.core.constants.CarbonCommonConstants;
+import org.apache.carbondata.core.scan.expression.Expression;
+import org.apache.carbondata.core.scan.expression.ExpressionResult;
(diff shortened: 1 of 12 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2094/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10348/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2299/

core/src/main/java/org/apache/carbondata/core/scan/expression/conditional/ImplicitExpression.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private void addBlockEntry(String blockletPath) {
+    String blockId =
+        blockletPath.substring(0, blockletPath.lastIndexOf(CarbonCommonConstants.FILE_SEPARATOR));
+    Set<Integer> blockletIds = blockIdToBlockletIdMapping.get(blockId);
+    if (null == blockletIds) {
+      blockletIds = new HashSet<>();
+      blockIdToBlockletIdMapping.put(blockId, blockletIds);
+    }
+    blockletIds.add(Integer.parseInt(blockletPath.substring(blockId.length() + 1)));
```
qiuchenjian: Better to catch the NumberFormatException for Integer.parseInt

core/src/main/java/org/apache/carbondata/core/scan/expression/conditional/ImplicitExpression.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public Map<String, Set<Integer>> getBlockIdToBlockletIdMapping() {
+    return blockIdToBlockletIdMapping;
+  }
+
+  @Override public ExpressionType getFilterExpressionType() {
+    return ExpressionType.IMPLICIT;
+  }
+
+  @Override public void findAndSetChild(Expression oldExpr, Expression newExpr) {
```
qiuchenjian: Why is this method empty?

manishgupta88: This method is not required to be implemented in this class so its implementation is empty

core/src/main/java/org/apache/carbondata/core/scan/expression/conditional/ImplicitExpression.java
manishgupta88: Not required to catch the NumberFormatException. Blocklet Id is always expected to be an integer number and in the complete flow it is used as an integer. So if there is any exception in the conversion then the code problem is not here but from the other part of code. So if we handle the exception here the actual cause will get suppressed here

ravipesala:
LGTM

kumarvishal09:
LGTM

PR #2417, opened 2018-06-27
[CARBONDATA-2607][Complex Column Enhancements]Complex Primitive DataType Adaptive Encoding

sounakr:
In this PR the improvement was done to save the complex type more effectively so that reading becomes more efficient. 

The changes are. 

- Primitive types inside complex types are separate pages. Previously it was a single byte array column page for a complex column. Now all sub-levels inside the complex data types are stored as separate pages with their respective datatypes. 

- No Dictionary Primitive DataTypes inside Complex Columns will be processed through Adaptive Encoding. Previously only snappy compression was applied.  

- All Primitive datatypes inside complex if it is now dictionary, only value will be saved except String, Varchar which is saved as ByteArray. Previously all sub-levels are saved as Length And Value Format inside a single Byte Array. Currently only Struct And Array type column pages are saved in ByteArray. All other primitive except String and varchar are saved in respective fixed datatype length. 

- Support for the Safe and Unsafe Fixed length Column Page to support growing dynamic array implementation.  This is done to support Array datatype.

 - [x] Any interfaces changed?
 
 - [x] Any backward compatibility impacted?
 
 - [x] Document update required?

 - [x] Testing done
        Please provide details on 
        - Whether new unit test cases have been added or why no new tests are required?
        - How it is tested? Please attach test report.
        - Is it a performance related change? Please attach the performance test report.
        - Any additional information to help reviewers in testing this change.
       
 - [x] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #3053, opened 2019-01-07
[CARBONDATA-3233]Fix JVM crash issue in snappy compressor and update the pagesize correctly

akashrn5:
### Problem:
1. During dataload sometimes the JVM is crashed during offheap snappy compression. We get the maximun compress size from compressor and allocate that much memory and then call rawCompress with the base offset of page and then base offset of newly created memory block. During this call sometimes JVM crashes from Snappy. This issue is random one and fails only sometimes.
2. pageSize is getting updated wrongly, actaul pageSize is number of rows in page, we were updaing the rowId not rowSize

### Solution:
Remove the method implementation and let super class handle the compression based on the datatype. which will fix this random JVM crash issue

 - [x] Any interfaces changed?
 NA
 NA
NA
 - [x] Testing done
Tested in three node cluster for 60GB data load
        Please provide details on 
       
NA

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/page/UnsafeFixLengthColumnPage.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/page/UnsafeFixLengthColumnPage.java
@@ -17,10 +17,8 @@
 
 package org.apache.carbondata.core.datastore.page;
 
-import java.io.IOException;
 import java.math.BigDecimal;
 
-import org.apache.carbondata.core.datastore.compression.Compressor;
 import org.apache.carbondata.core.datastore.page.encoding.ColumnPageEncoderMeta;
 import org.apache.carbondata.core.memory.CarbonUnsafe;
 import org.apache.carbondata.core.memory.MemoryBlock;
@@ -112,7 +110,8 @@ public class UnsafeFixLengthColumnPage extends ColumnPage {
 
   private void updatePageSize(int rowId) {
     if (pageSize < rowId) {
-      pageSize = rowId;
+      // update the actual number of rows
+      pageSize = rowId + 1;
     }
   }
 
@@ -359,7 +358,7 @@ public class UnsafeFixLengthColumnPage extends ColumnPage {
 
   @Override
   public float[] getFloatPage() {
-    float[] data = new float[getPageSize()];
+    float[] data = new float[getEndLoop()];
     for (long i = 0; i < data.length; i++) {
       long offset = i << floatBits;
       data[(int)i] = CarbonUnsafe.getUnsafe().getFloat(baseAddress, baseOffset + offset);
@@ -369,7 +368,7 @@ public class UnsafeFixLengthColumnPage extends ColumnPage {
 
   @Override
   public double[] getDoublePage() {
-    double[] data = new double[getPageSize()];
+    double[] data = new double[getEndLoop()];
     for (long i = 0; i < data.length; i++) {
       long offset = i << doubleBits;
       data[(int)i] = CarbonUnsafe.getUnsafe().getDouble(baseAddress, baseOffset + offset);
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Failed  with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10449/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2409/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2198/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2415/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10455/

core/src/main/java/org/apache/carbondata/core/datastore/page/UnsafeFixLengthColumnPage.java
```diff
@@ -369,7 +367,7 @@ public BigDecimal getDecimal(int rowId) {
 
   @Override
   public double[] getDoublePage() {
-    double[] data = new double[getPageSize()];
+    double[] data = new double[getEndLoop()];
```
qiuchenjian: the return values of getPageSize() and getEndLoop seem to be same, when they are diff?

qiuchenjian:
i think the performance of rawCompress is better than compressLong,compressInt, can we find the root cause  of JVM crash？

core/src/main/java/org/apache/carbondata/core/datastore/page/UnsafeFixLengthColumnPage.java
akashrn5: during complex type enhancement, to convert value and to get the value, pageSize was changed and new method called getEndLoop added, this was missed for double, so when double datatype is there in complex type data miss match happens, this is handled here. you can refer #2417

akashrn5:
> i think the performance of rawCompress is better than compressLong,compressInt, can we find the root cause of JVM crash？

i dont think there is much difference we get with timing, but problem is JVM crash happens randomly, since we get maxsizefor compression from snapy itself, we even allocated that memory sucessfully and passed the address to snappy, after that JVM crashed. This is very random. So better to remove that.

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2436/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10472/

core/src/main/java/org/apache/carbondata/core/datastore/page/UnsafeFixLengthColumnPage.java
qiuchenjian: get it， thank you

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2440/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10477/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2220/

xuchuanyin:
Does this PR fix two problems?
If it is yes, better to separate it into two. And for the first problem, I'm also concerning about the performance decrease. The rawCompress can save some memory copy operations, that's why we add a check there and try to use that feature if the compressor supports that. It may needs more observations about the performance decreasement OR we can just add a switch there to control the behavior and it will be helpful for comparison.

manishgupta88:
@akashrn5 .....I agree with @xuchuanyin ....before merging the PR it is better to get the PR tested for performance. We can observe 2 things during the benchmark test - performance and compression ratio of rawcompress Vs compressDouble and then take the final decision

kumarvishal09:
@manishgupta88 @xuchuanyin I think if it's really a problem with snappy then whether any performance impact is there or not we have to merge as its a functional issue. :)
@akashrn5 May be this issue is coming because of offheap to onheap fallback in UnsafeMemoryManager can u please verify once. Please try discuss with snappy community also.

akashrn5:
@kumarvishal09 i have tested the fallback scenario by changing code, it is even failing with that also and i have raised discussion in snappy community also [https://groups.google.com/forum/#!topic/snappy-compression/4noNVKCMBqM](url)

manishgupta88:
@kumarvishal09 ...I agree with you that it is a functional issue and we need to merge it. My point was before merging we can do one load performance test to see if there is any performance degrade and if there is any then we can update the benchmark results

akashrn5:
> Does this PR fix two problems?
> If it is yes, better to separate it into two.
> 

the one line change of rowId to rowId + 1 is coupled with this, when i removed the compress method in unSafeFixLengthColumnPage, i got this issue and fixed in this, so this is required in this PR only

akashrn5:
@xuchuanyin @manishgupta88 @kumarvishal09 i have tested the data loading of 4.2TB raw data , with 3.5 billion rows with and without my changes, so the data loading time is same for both which is 1hr 10minutes. So these changes can be considerd as there is no major impact for performance.

manishgupta88:
LGTM

PR #2591, opened 2018-07-31
[CARBONDATA-2808] Insert into select is crashing as both are sharing the same task context

dhatchayani:
**Problem:**
Insert into select is failing as both are running as single task, both are sharing the same taskcontext and resources are cleared once if any one of the RDD(Select query's ScanRDD) is completed, so the other RDD(LoadRDD) running is crashing as it is trying to access the cleared memory. 

**Solution**:
Check if any other RDD is sharing the same task context. If so, don't the clear the resource at that time, the other RDD which shared the context should clear the memory once after the task is finished.

 - [ ] Any interfaces changed?
 
 - [ ] Any backward compatibility impacted?
 
 - [ ] Document update required?

 - [x] Testing done
        Manual Testing
 - [ ] For large changes, please consider breaking it into sub-tasks under an umbrella JIRA.

PR #3056, opened 2019-01-08
[CARBONDATA-3236] Fix for JVM Crash for insert into new table from old table

manishnalla1994:
Problem: Insert into new table from old table fails with JVM crash for file format(Using carbondata). This happened because both the query and load flow were assigned the same taskId and once query finished it freed the unsafe memory while the insert still in progress.

Solution: As the flow for file format is direct flow and uses on-heap(safe) so no need to free the unsafe memory in query.

 - [ ] Any interfaces changed?
 
 - [x] Testing done
        Please provide details on 
       
### DIFF
--- a/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
+++ b/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
@@ -410,15 +410,8 @@ class SparkCarbonFileFormat extends FileFormat
         val model = format.createQueryModel(split, hadoopAttemptContext)
         model.setConverter(new SparkDataTypeConverterImpl)
         model.setPreFetchData(false)
-        var isAdded = false
-        Option(TaskContext.get()).foreach { context =>
-          val onCompleteCallbacksField = context.getClass.getDeclaredField("onCompleteCallbacks")
-          onCompleteCallbacksField.setAccessible(true)
-          val listeners = onCompleteCallbacksField.get(context)
-            .asInstanceOf[ArrayBuffer[TaskCompletionListener]]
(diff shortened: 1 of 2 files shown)


CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2437/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2217/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10475/

xuchuanyin:
> because both the query and load flow were assigned the same taskId and once query finished it freed the unsafe memory while the insert still in progress.

How do you handle the scenario for stored-by-carbondata carbontable? In that scenario, both of the query flow and load flow use offheap and encounter the same problem just as you described above.
But I remembered we handle that differently from the current PR, which I think their modification can be similar. Please check this again.

manishnalla1994:
@xuchuanyin Datasource table uses direct filling flow. As in direct flow there is no intermediate buffer so we are not using off-heap to store the page data in memory(filling all the records of a page to vector instead of filling batch wise). So in this case we can remove freeing of unsafe memory for Query as its not required.

In case of stored by table, handling will be different as we support both batch wise filling and direct filling and for batch filling we are using unsafe, so we have to clear unsafe memory in this case.
Here same handling is not required for data source table.
Please refer https://github.com/apache/carbondata/pull/2591 for stored by handling of this issue.

ravipesala:
LGTM

xuchuanyin:
@manishnalla1994 

> Solution:
Check if any other RDD is sharing the same task context. If so, don't the clear the resource at that time, the other RDD which shared the context should clear the memory once after the task is finished.

It seems in #2591, for data source table scenario, if the query and insert procedures also share the same context, it can also benefit from the implementation in #2591 without any changes. Right?

PR #3068, opened 2019-01-11
[HOTFIX] Fixed NPE during query with Local Dictionary

kumarvishal09:
**Problem:**
Query is failing with NPE when some blocklet encoded with local dictionary and some without local dictionary.
**Root Cause:** 
This is coming because in carbonvectorProxy setDictionary with null it is not setting the dictionary to null because of this it is treated like a local dictionary column but column is not encoded with dictionary.
**Solution:**
Set dictionary to null
 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/integration/spark-datasource/src/main/spark2.1andspark2.2/org/apache/spark/sql/CarbonVectorProxy.java
+++ b/integration/spark-datasource/src/main/spark2.1andspark2.2/org/apache/spark/sql/CarbonVectorProxy.java
@@ -311,6 +311,7 @@ public class CarbonVectorProxy {
         vector.setDictionary(dictionaryWrapper);
         this.dictionary = dictionaryWrapper;
       } else {
+        this.dictionary = null;
         vector.setDictionary(null);
       }
     }


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2269/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2488/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10527/

qiuchenjian:
why does one segment have some blocklet encoded with local dictionary and some without local dictionary ?

ravipesala:
> why does one segment have some blocklet encoded with local dictionary and some without local dictionary ?

It is because carbon generates a dictionary based on the column value count threshold, so once it reaches that threshold it stops generating the dictionary. There are scenarios where some blocks/blocklets are with in threshold and some are not, thats why some blocks has local dictionary and some don't have

ravipesala:
LGTM

qiuchenjian:
@ravipesala I remember if the local dictionary value count reach threshold, it will go back to original value， right？
Does some nodes's local dictionary reach the threshold, bug other's not, so this scene appeared ？（some shards have local dictionary, some shards don't have  local dictionary）

integration/spark-datasource/src/main/spark2.1andspark2.2/org/apache/spark/sql/CarbonVectorProxy.java
```diff
@@ -311,6 +311,7 @@ public void setDictionary(CarbonDictionary dictionary) {
         vector.setDictionary(dictionaryWrapper);
         this.dictionary = dictionaryWrapper;
       } else {
+        this.dictionary = null;
```
xuchuanyin: This change lies in spark2.1andspark2.2, so spark2.3 does not have this problem？

ravipesala:
> @ravipesala I remember if the local dictionary value count reach threshold, it will fallback to original value， right？
> what level does it fallback? blocklet, page or block ?

The fallback is at page level.

qiuchenjian:
> > @ravipesala I remember if the local dictionary value count reach threshold, it will fallback to original value， right？
> > what level does it fallback? blocklet, page or block ?
> 
> The fallback is at page level.

get it, thanks

integration/spark-datasource/src/main/spark2.1andspark2.2/org/apache/spark/sql/CarbonVectorProxy.java
kumarvishal09: In 2.3 proper interface are exposed from spark and same has been implemented, so its not required

ravipesala:
LGTM

PR #3120, opened 2019-02-02
[CARBONDATA-3286] MV datamap doesn't take effect when query SQL has coalesce with all projections and no filter condition

qiuchenjian:
[Problem]
MV datamap doesn't take effect when query SQL has oalesce with all projections and no filter condition
test case:
`create table test_main(id int,name string,height int,weight int) using carbondata`
`create datamap test_main_mv using 'mv' as select sum(id) as sum_id, sum(height) as sum_height,name as myname from test_main group by name`
`select coalesce(sum(id),12) as sumid, sum(height) as sum_height  from test_main group by name`

[Solution]
Fix this bug to support the scene. 
`[== Physical Plan ==
*HashAggregate(keys=[myname#269], functions=[sum(sum_id#267L), sum(sum_height#268L)])
+- Exchange hashpartitioning(myname#269, 200)
   +- *HashAggregate(keys=[myname#269], functions=[partial_sum(sum_id#267L), partial_sum(sum_height#268L)])
      +- *FileScan carbondata default.test_main_mv_table[sum_id#267L,sum_height#268L,myname#269]]`

 - [ ] Any interfaces changed?
 
 - [ ] Testing done
        Please provide details on 
       
### DIFF
--- a/datamap/mv/core/src/main/scala/org/apache/carbondata/mv/rewrite/DefaultMatchMaker.scala
+++ b/datamap/mv/core/src/main/scala/org/apache/carbondata/mv/rewrite/DefaultMatchMaker.scala
@@ -354,6 +354,30 @@ object SelectSelectNoChildDelta extends DefaultMatchPattern with PredicateHelper
 }
 
 object GroupbyGroupbyNoChildDelta extends DefaultMatchPattern {
+  def  tryMatch(gb_2a: GroupBy,
+               gb_2q: GroupBy,
+               isGroupingEmR: Boolean,
+                isInheritTableRelation: Boolean) : Seq[ModularPlan] = {
+    val aliasMap = AttributeMap(gb_2a.outputList.collect { case a: Alias =>
+      (a.toAttribute, a)})
+    if (isGroupingEmR) {
+      Utils.tryMatch(
+        gb_2a, gb_2q, aliasMap).flatMap {
+        case g: GroupBy =>
+          Some(g.copy(child = g.child.withNewChildren(
+            g.child.children.map {
+              case modular.Select(_, _, _, _, _, _, _, _, _, _) => gb_2a;
+              case other => other
+            }), dataMapTableRelation =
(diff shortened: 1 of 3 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2543/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2772/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10802/

datamap/mv/core/src/main/scala/org/apache/carbondata/mv/rewrite/DefaultMatchMaker.scala
```diff
@@ -354,6 +354,30 @@ object SelectSelectNoChildDelta extends DefaultMatchPattern with PredicateHelper
...
+    if (isGroupingEmR) {
+      Utils.tryMatch(
+        gb_2a, gb_2q, aliasMap).flatMap {
+        case g: GroupBy =>
+          Some(g.copy(child = g.child.withNewChildren(
+            g.child.children.map {
+              case modular.Select(_, _, _, _, _, _, _, _, _, _) => gb_2a;
+              case other => other
+            }), dataMapTableRelation =
+            if (isInheritTableRelation) gb_2a.dataMapTableRelation
```
ravipesala: It is sort of hack, we should not copy the `dataMapTableRelation` like this. Actual matching should happen and the whole plan should be replaced not just `dataMapTableRelation`.

kevinjmh:
we may block `coalesce` when creating MV, because this will affect incremental build of MV. 
 
According to the usage by now, we should keep MV in a simple and flexible way, so that mv can fit more query. Further, MV should provide a guideline

PR #3131, opened 2019-02-20
[CARBONDATA-3298]Remove Log Message for Already Deleted Segments in old stores

Indhumathi27:
Problem:
In old store, Create table and perform one Insert operation. Now, update and delete that record, which marks that segment as  "MARKED FOR DELETE". Now run "clean files command" to delete the segment.
Note: In this case, Metadata folder doesn't  contain segment file.
In new store, Refresh the table and again perform IUD operation. Now, when "clean files" command is executed,  We will check if physically segment file exists or not, if not present, then we will log a warning message, as file not present. If old store contains more segments, then for each segment, this log message will be getting printed, which is not required.

Solution:
Removed log message for already deleted segments

 - [ ] Any interfaces changed?
 
 - [x] Testing done
       Testing done manually

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
@@ -156,11 +156,7 @@ public final class DeleteLoadFolders {
                 }
               }
 
-            } else {
-              LOGGER.warn("Files are not found in segment " + path
-                  + " it seems, files are already being deleted");
             }
-
           }
           List<Segment> segments = new ArrayList<>(1);
           for (TableDataMap dataMap : indexDataMaps) {


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/2587/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/10847/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/2818/

qiuchenjian:
In new store, do we need this log ?

Indhumathi27:
@qiuchenjian In new store also, this log is not required.

xuchuanyin:
@Indhumathi27 
Sorry that I didn't understand the problem clearly. Do you mean that 'clean files' did clean the files but didn't clean the segment folder, so in the new store, carbondata will complain that the are no files in that segment folder?
If it is so, I'd suggest that we should delete the segment folder in the old store, this will solve the problem at the root.

Indhumathi27:
@xuchuanyin No,  Clean files will delete the segment folder and files. In new store, when  clean files command is executed, we will take LoadMetaDetails from TableStatus file and delete Segment folder if Physically exist or if Segment file in MetaData folder exist. In this case, LoadMetaDetails will contain Details for already deleted segments in old store, where Segment folder physicaly does not exist, hence we will log the message for those Segments.

xuchuanyin:
@Indhumathi27 So the root cause is that while marking the segment as ‘marked as deleted’, the tablestatus file does not get updated, right?
I‘m wondering whether 'delete segment' command update the tablestatus file?

Indhumathi27:
@xuchuanyin we cannot decide on SegmentStatus "MARKED_FOR_DELETE" to check if physically segment folder exist or not. Both already deleted segments and segments ready for clean using "clean files" will be marked as "MARKED_FOR_DELETE" only. And i dont think this log is needed, if any error occurs during clean files, then there are other logs to display the error message

kunal642:
LGTM

PR #3197, opened 2019-04-30
[CARBONDATA-3367][CARBONDATA-3368] Fix multiple issues in SDK reader

ajantha-bhat:
Problem:
[CARBONDATA-3367] OOM when huge number of carbondata files are read from SDK reader

Cause:
Currently, for each carbondata file, one CarbonRecordReader will be created. And list of CarbonRecordReader will be maintained in carbonReader. so even when CarbonRecordReader is closed, the GC will not happen for that reader as list is still referring that object. 
so, each CarbonRecordReader needs separate memory , instead of reusing the previous memory. 

Solution : Once CarbonRecordReader.close is done, remove it from the list

problem:
[CARBONDATA-3368]InferSchema from datafile instead of index file

cause:
problem : In SDK, when multiple readers were created with same folder location with different file list, for inferschema all the readers refers same index file, which was causing bottle neck and JVM crash in case of JNI call.

solution: Inferschema from the data file mentioned while building the reader.

problem :
Support list interface for projection, when SDK is called from other languages, JNI interface supports only list from other languages. so need to add list interface for projections.
 
 - [ ] Testing done. Done
       
### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/schema/table/CarbonTable.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/schema/table/CarbonTable.java
@@ -38,7 +38,6 @@ import org.apache.carbondata.core.datamap.TableDataMap;
 import org.apache.carbondata.core.datamap.dev.DataMapFactory;
 import org.apache.carbondata.core.datastore.block.SegmentProperties;
 import org.apache.carbondata.core.datastore.filesystem.CarbonFile;
-import org.apache.carbondata.core.datastore.impl.FileFactory;
 import org.apache.carbondata.core.features.TableOperation;
 import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier;
 import org.apache.carbondata.core.metadata.CarbonTableIdentifier;
@@ -252,12 +251,13 @@ public class CarbonTable implements Serializable {
       String tableName,
       Configuration configuration) throws IOException {
     TableInfo tableInfoInfer = CarbonUtil.buildDummyTableInfo(tablePath, "null", "null");
-    CarbonFile carbonFile = getLatestIndexFile(FileFactory.getCarbonFile(tablePath, configuration));
-    if (carbonFile == null) {
-      throw new RuntimeException("Carbon index file not exists.");
-    }
-    org.apache.carbondata.format.TableInfo tableInfo = CarbonUtil
-        .inferSchemaFromIndexFile(carbonFile.getPath(), tableName);
+    // CarbonFile carbonFile =
+    //                      getLatestIndexFile(FileFactory.getCarbonFile(tablePath, configuration));
+    // if (carbonFile == null) {
+    //  throw new RuntimeException("Carbon index file not exists.");
+    // }
+    org.apache.carbondata.format.TableInfo tableInfo =
+        CarbonUtil.inferSchema(tablePath, tableName, false, configuration);
     List<ColumnSchema> columnSchemaList = new ArrayList<ColumnSchema>();
     for (org.apache.carbondata.format.ColumnSchema thriftColumnSchema : tableInfo
         .getFact_table().getTable_columns()) {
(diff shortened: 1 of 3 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/3038/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/11302/

ajantha-bhat:
@kumarvishal09 , @ravipesala : please review

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReaderBuilder.java
```diff
@@ -83,6 +83,21 @@ public CarbonReaderBuilder projection(String[] projectionColumnNames) {
     return this;
   }
 
+  /**
+   * Accepts projection list
+   *
+   * @param projectionColumnNames
+   * @return
+   */
+  public CarbonReaderBuilder projection(List<String> projectionColumnNames) {
```
xubo245: Why  create new interface in this PR?

core/src/main/java/org/apache/carbondata/core/metadata/schema/table/CarbonTable.java
```diff
@@ -252,12 +250,9 @@ public static CarbonTable buildTable(
...
     TableInfo tableInfoInfer = CarbonUtil.buildDummyTableInfo(tablePath, "null", "null");
-    CarbonFile carbonFile = getLatestIndexFile(FileFactory.getCarbonFile(tablePath, configuration));
-    if (carbonFile == null) {
-      throw new RuntimeException("Carbon index file not exists.");
-    }
-    org.apache.carbondata.format.TableInfo tableInfo = CarbonUtil
-        .inferSchemaFromIndexFile(carbonFile.getPath(), tableName);
+    // InferSchema from data file
+    org.apache.carbondata.format.TableInfo tableInfo =
+        CarbonUtil.inferSchema(tablePath, tableName, false, configuration);
```
xubo245: Did you consider the performance for reading data from S3? it will increase the cost if inferSchema from S3 data file, right?

ajantha-bhat: only footer is read. so not much difference and only one IO. so, not much difference.
Index file was introduced to avoid multiple data file footer reading, but here for carbon reader to infer schema one file footer reading is enough.

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReaderBuilder.java
ajantha-bhat: This was also one of the issue. we need interface that accepts List. Interfacing with other languages is difficult if we use just array interface.

xubo245: Please describe this issue in PR content， and add test cases for it。

core/src/main/java/org/apache/carbondata/core/metadata/schema/table/CarbonTable.java
xubo245: Carbon already support download little part data of whole file from S3?

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonReaderBuilder.java
ajantha-bhat: ok done

core/src/main/java/org/apache/carbondata/core/metadata/schema/table/CarbonTable.java
ajantha-bhat: I think it's supported by default. Let me test once in S3

ajantha-bhat: @xubo245 : I have tested with huge data file (800 MB) and small index fie (500 KB), infer schema time is same for both these files.
reason is only header is read from s3 and header is in top itself, so no seek operations also.

ajantha-bhat:
@xubo245 @kumarvishal09 : please check

xubo245:
LGTM

ajantha-bhat:
@ravipesala : PR is ready please check.

ravipesala:
LGTM

PR #3243, opened 2019-05-28
[CARBONDATA-3401] fix the java sdk create wrong carbondata filename

lamber-ken:
### Description
When use sdk to write carbondata files, use `System.nanoTime()` assign to the start timestamp and use `System.nanoTime()` assign to taskNo currently.

There are two bug here, first `System.nanoTime()` is different from `System.currentTimeMillis()`, it can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. Second, the carbondata file name written by sdk was different from written by spark.

For example, the carbondata file name are below
```
110438274127988_batchno0-0-null-110437247548158.carbonindex
part-0-110438274127988_batchno0-0-null-110437247548158.carbondata
```

 - [x] Any interfaces changed?
 
 - [x] Testing done
        Please provide details on 
       
### DIFF
--- a/store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonWriterBuilder.java
+++ b/store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonWriterBuilder.java
@@ -580,7 +580,7 @@ public class CarbonWriterBuilder {
 
   public CarbonLoadModel buildLoadModel(Schema carbonSchema)
       throws IOException, InvalidLoadOptionException {
-    timestamp = System.nanoTime();
+    timestamp = System.currentTimeMillis();
     // validate long_string_column
     Set<String> longStringColumns = new HashSet<>();
     if (options != null && options.get(CarbonCommonConstants.LONG_STRING_COLUMNS) != null) {
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableOutputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableOutputFormat.java
@@ -254,7 +254,7 @@ public class CarbonTableOutputFormat extends FileOutputFormat<NullWritable, Obje
     }
     if (null == loadModel.getTaskNo() || loadModel.getTaskNo().isEmpty()) {
       loadModel.setTaskNo(taskAttemptContext.getConfiguration()
-          .get("carbon.outputformat.taskno", String.valueOf(System.nanoTime())));
+          .get("carbon.outputformat.taskno", String.valueOf(System.currentTimeMillis())));
     }
     loadModel.setDataWritePath(
         taskAttemptContext.getConfiguration().get("carbon.outputformat.writepath"));


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://136.243.101.176:8080/job/ApacheCarbonPRBuilder2.1/3353/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://136.243.101.176:8080/job/carbondataprbuilder2.3/11620/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://95.216.28.178:8080/job/ApacheCarbonPRBuilder1/3554/

store/sdk/src/main/java/org/apache/carbondata/sdk/file/CarbonWriterBuilder.java
```diff
@@ -580,7 +580,7 @@ private void setCsvHeader(CarbonLoadModel model) {
 
   public CarbonLoadModel buildLoadModel(Schema carbonSchema)
       throws IOException, InvalidLoadOptionException {
-    timestamp = System.nanoTime();
+    timestamp = System.currentTimeMillis();
```
BJangir: in Concurrency (in same JVM multiple threads) System.currentTimeMillis() can be same , which can overwrite existing carbonfile .

lamber-ken: > in Concurrency (in same JVM multiple threads) System.currentTimeMillis() can be same , which can overwrite existing carbonfile .

Hi @BJangir, thanks for your mention about the concurrent problem. I have some points from my side.
**First**, the filename is consisted of (taskNo, batch_prefix, batchNo, bucketNumber, segmentNo, factUpdateTimestamp), I think the filename may nerver same.

**Second**, it's a wrong way to use `System.nanoTime()` to assign to the timestamp, the value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative), so the filename may like this
```
53353670335116_batchno0-0-null--12341312341.carbonindex
part-0--100010000_batchno0-0-null--12341312341.carbondata
```

**Thrid**, from the java doc, it says the values returned by this method become meaningful only when the difference between two such values, obtained within the same instance of a Java virtual machine, is computed. 

[Java Platform SE doc](https://docs.oracle.com/javase/8/docs/api/index.html)

Please correct me if I am wrong, thanks.

BJangir: :+1:

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/22/

ajantha-bhat:
@lamber-ken : In SDK writer, both `taskNo` and `timestamp` are configurable. If user wants milli, user can configure the value.

ajantha-bhat:
@lamber-ken : In SDK writer, both taskNo and timestamp are configurable. If user wants milli, user can configure the value.

Also SDK uses nano Second time so that in concurrency the chances that it will be unique than milli seconds. 

You might saw that name will have `taskNo, batch_prefix, batchNo, bucketNumber, segmentNo, factUpdateTimestamp`. But in most of the cases `batchNo, bucketNumber` will be zero and segmentNo will be null for SDK. 

so only differentiating parameters is task number and timestamp. In case of multi-thread SDK writing. task number also same. same task will write multiple files concurrently. That time to avoid same name we use nanoSeconds. 

In carbon session this problem will not come as segment will not be null. 

So I guess, that current design solves the problem. If we change to milli second, we get issues

lamber-ken:
> @lamber-ken : In SDK writer, both `taskNo` and `timestamp` are configurable. If user wants milli, user can configure the value.
> 
> Also SDK uses nano Second time so that in concurrency the chances that it will be unique than milli seconds.
> 
> You might saw that name will have `taskNo, batch_prefix, batchNo, bucketNumber, segmentNo, factUpdateTimestamp`. But in most of the cases `batchNo, bucketNumber` will be zero and `segmentNo` will be null for SDK.
> 
> so only differentiating parameters is `taskNo` and `timestamp`. In case of multi-thread SDK writing, `taskNo` also same. same task will write multiple files concurrently. To avoid same name we use nanoSeconds.
> 
> In carbon session, this problem will not come as segment will not be null.
> 
> So I guess, that current design solves the problem. If we change to milli second, we get issues

Thanks for your comment

ajantha-bhat:
@lamber-ken : please close this PR as discussed above.

PR #3372, opened 2019-09-04
[CARBONDATA-3511] Query time improvement by reducing the number of NameNode calls while having carbonindex files in the store

dhatchayani:
**Problem:**
If the store contains only carbonindex files, while querying that table, the index file to block details mapping needs to be formed. This formation is currently segment level. So the map formed will be of SegmentPath to all its indexFilePath to its block details. This is not reused and the same map is formed for each and every indexFile. This increases the name node calls as for each indexFile it needs to form the map segment level(which includes all the indexFiles).

**Solution:**
Make the map global and reuse it in needed times and clear the same in needed scenarios.

 - [ ] Any interfaces changed?
 
 - [x] Testing done
        No.of Index Files = 110
        Query Time in seconds

| **Before Changes** |  **After changes**|
|--|--|
|16.665  | 2.077 |
       
### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
+++ b/core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
@@ -65,6 +65,13 @@ public class BlockletDataMapIndexStore
   private Map<String, Object> segmentLockMap;
 
   /**
+   * map of segmentPath to the map of index/ merge index file path to its
+   * corresponding block info. While querying the table this map will be useful while creating the
+   * block info for the index or merge index file.
+   */
+  private Map<String, Map<String, BlockMetaInfo>> segInfoCache;
+
+  /**
    * constructor to initialize the SegmentTaskIndexStore
    *
    * @param lruCache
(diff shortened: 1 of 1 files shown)


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/403/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/403/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/402/

core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
```diff
@@ -234,6 +231,10 @@ private BlockletDataMapIndexWrapper get(TableBlockIndexUniqueIdentifierWrapper i
                 tableSegmentUniqueIdentifierWrapper.isAddTableBlockToUnsafeAndLRUCache());
       }
     }
+    String segmentPath =
+        tableSegmentUniqueIdentifierWrapper.getTableBlockIndexUniqueIdentifier().getIndexFilePath();
```
jackylk: `getIndexFilePath()` returns `segmentPath`?

core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
```diff
@@ -64,6 +64,13 @@
    */
   private Map<String, Object> segmentLockMap;
 
+  /**
+   * map of segmentPath to the map of index/ merge index file path to its
+   * corresponding block info. While querying the table this map will be useful while creating the
+   * block info for the index or merge index file.
+   */
+  private Map<String, Map<String, BlockMetaInfo>> segInfoCache;
```
ravipesala: Global cache will be risky, how to clean it? Better have a method level cache.

core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
dhatchayani: yes. getIndexFilePath() will return the indexFile parent path, which is segmentPath

core/src/main/java/org/apache/carbondata/core/indexstore/BlockletDataMapIndexStore.java
dhatchayani: ok.. we can reuse the getAll method to reuse the method level cache.
Closing this PR

dhatchayani:
As Global level cache is risky and need to be cleaned at needed scenarios, we can reuse the method level cache from the getAll method. Closing this PR

PR #3433, opened 2019-11-04
[CARBONDATA-3570] Change task number to jobid+taskid for FileFormat

ajantha-bhat:
problem :  Incase of File format, task retry is having different task number, so file name is becoming different

cause : Everytime in executor, UUID is used to generate task number for carbondata file name. so if previous task has copied few files insted of all files and it got crashed. New task will create new file instead of overwriting existing file

solution: make task number in file name same for task retry by using taskid + job number

before:
`edff9ef0cd66484ebbaff6ad683f0f62_batchno0-0-null-88487116477262.carbonindex`
After:
`2019110420310511x0_batchno0-0-null-14565827943299.carbonindex`

 - [ ] Testing done. UT is executed.

       
### DIFF
--- a/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
+++ b/integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
@@ -154,8 +154,11 @@ class SparkCarbonFileFormat extends FileFormat
           path
         }
         context.getConfiguration.set("carbon.outputformat.writepath", updatedPath)
+        // "jobid"+"x"+"taskid", task retry should have same task number
         context.getConfiguration.set("carbon.outputformat.taskno",
-          UUID.randomUUID().toString.replace("-", ""))
+          context.getTaskAttemptID.getJobID.getJtIdentifier +
+          context.getTaskAttemptID.getJobID.getId
+          + 'x' + context.getTaskAttemptID.getTaskID.getId)
         new CarbonOutputWriter(path, context, dataSchema.fields)
       }
 


CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/737/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/744/

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/746/

ajantha-bhat:
@ravipesala@caiqing please check

kumarvishal09:
@ajantha-bhat pls change the wip

ajantha-bhat:
@kumarvishal09 : done

integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
```diff
@@ -154,8 +154,11 @@ class SparkCarbonFileFormat extends FileFormat
           path
         }
         context.getConfiguration.set("carbon.outputformat.writepath", updatedPath)
+        // "jobid"+"x"+"taskid", task retry should have same task number
         context.getConfiguration.set("carbon.outputformat.taskno",
-          UUID.randomUUID().toString.replace("-", ""))
+          context.getTaskAttemptID.getJobID.getJtIdentifier +
+          context.getTaskAttemptID.getJobID.getId
+          + 'x' + context.getTaskAttemptID.getTaskID.getId)
```
jackylk: Why using `x` in between? Can we follow the same naming convention as carbondata table?

ajantha-bhat: @jackylk 
a. 'x' is used to separate jobid and task id.
b. we cannot use logic from carbondata table as taskid, in this case it will be just split index. And this id will be serialized and send to executor. So, during task retry it will use same id. However in case of file format, it was using UUID for each task.

ajantha-bhat: @ravipesala : please confirm and share your opinion.

kumarvishal09: @ajantha-bhat can we use underscore as separator??

ajantha-bhat: @kumarvishal09 : current file name has many underscore, If some parsing logic depends on underscore, it will have impact. Hence using 'x'

brijoobopanna:
@ajantha-bhat If this issue fix is specific to Carbon file format or applicable to Carbon session too

ajantha-bhat:
@brijoobopanna : This is only specific to file format. 

in carbonsession case, task id will be just split index. [concurrent scenario , no issue because it has segment id] And this id will be serialized and send to executor. So, during task retry it will use same id

integration/spark-datasource/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala
kumarvishal09: @ajantha-bhat then it’s better to use some other characters
Like #,-

QiangCai: old file name already use "-", "_"  to separate each parts. 
taskNo is a single part, no need to use "-", "_"  to separate it. 
So we use 'x' to contact  jobId and taskId to a single part of file name.

QiangCai:
look good for me

jackylk:
LGTM

PR #3449, opened 2019-11-13
[CARBONDATA-3578] Make table status file smaller

jackylk:
Currently, each segment entry in the table status file occupies 347 Bytes, if one has 10000 segments, the file becomes 3.47MB.
Since carbondata relies on this file heavily, it is better to reduce its size to improve IO, especially in data lake scenario.

Each entry in table status file is one LoadMetadataDetails object.
In this PR, following changes are made in LoadMetadataDetails to reduce its size:
1. Do not write fields that has default value, like "visibility", "fileFormat", etc
2. Use shorter key, for example, "loadStatus" is changed to "ls"

In this PR, table status file size is reduced to 1/3. 
Before change: 347Bytes
```json
     {
         "timestamp": "1573635015982",
         "loadStatus": "Success",
         "loadName": "0",
         "partitionCount": "0",
         "isDeleted": "FALSE",
         "dataSize": "2977",
         "indexSize": "1469",
         "updateDeltaEndTimestamp": "",
         "updateDeltaStartTimestamp": "",
         "updateStatusFileName": "",
         "loadStartTime": "1573635014638",
         "visibility": "true",
         "fileFormat": "columnar_v3",
         "segmentFile": "0_1573635014638.segment"
     }
```

After change: 118Bytes ( reduced to 1/3 size)
```json
     {
         "ts": "1573635284677",
         "ls": "S",         -- stands for Success
         "ln": "0",
         "ds": "2977",
         "is": "1469",
         "lt": "1573635284045",
         "sf": "0_1573635284045.segment"
     }
```

About the backward compatibility, this PR still can read the old table status file, by using GSON's @SerializedName(alternate), so it does not break backward compatibility.

 - [X] Any interfaces changed?
 No
 - [X] Any backward compatibility impacted?
 No
 - [X] Document update required?
No
 - [X] Testing done
        Please provide details on 
    Run existing testcase   
No

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -166,8 +166,8 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
     </plugins>
--- a/core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
+++ b/core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
@@ -26,6 +26,7 @@ import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
 
 import com.google.gson.annotations.SerializedName;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 
 /*
@@ -488,19 +489,19 @@ public class LoadMetadataDetails implements Serializable {
    * It checks if fields are default value, then make it null so GSON does not write it
    */
   void removeUnnecessaryField() {
-    if (updateDeltaEndTimestamp == null || updateDeltaStartTimestamp.isEmpty()) {
+    if (StringUtils.isEmpty(updateDeltaEndTimestamp)) {
       updateDeltaEndTimestamp = null;
     }
-    if (updateDeltaStartTimestamp == null || updateDeltaStartTimestamp.isEmpty()) {
+    if (StringUtils.isEmpty(updateDeltaStartTimestamp)) {
       updateDeltaStartTimestamp = null;
     }
-    if (updateStatusFileName == null || updateStatusFileName.isEmpty()) {
+    if (StringUtils.isEmpty(updateStatusFileName)) {
       updateStatusFileName = null;
     }
-    if (visibility == null || visibility.equals("true")) {
+    if (StringUtils.isEmpty(visibility) || visibility.equals("true")) {
       visibility = null;
     }
-    if (fileFormat == null || fileFormat.equals(FileFormat.COLUMNAR_V3.toString())) {
+    if (StringUtils.isEmpty(fileFormat) || fileFormat.equals(FileFormat.COLUMNAR_V3.toString())) {
       fileFormat = null;
     }
   }
--- a/core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
+++ b/core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
@@ -25,6 +25,7 @@ import java.util.Date;
 import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
 
+import com.google.gson.annotations.SerializedName;
 import org.apache.log4j.Logger;
 
 /*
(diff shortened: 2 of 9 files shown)


CarbonDataQA:
Build Failed  with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/832/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/829/

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/826/

core/pom.xml
```diff
@@ -166,8 +166,8 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
```
zzcclp: why change this in this pr? all modules are using 1.7 as default version, it's better raise another pr to handle this ?

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
```diff
@@ -468,4 +482,26 @@ public boolean isCarbonFormat() {
     return getFileFormat().equals(FileFormat.COLUMNAR_V3)
         || getFileFormat().equals(FileFormat.ROW_V1);
   }
+
+  /**
+   * Before writing table status file, call this to make the metadata smaller.
+   * It checks if fields are default value, then make it null so GSON does not write it
+   */
+  void removeUnnecessaryField() {
+    if (updateDeltaEndTimestamp == null || updateDeltaStartTimestamp.isEmpty()) {
```
zzcclp: It's better use StringUtils.isEmpty()

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
```diff
@@ -468,4 +482,26 @@ public boolean isCarbonFormat() {
...
+      updateDeltaEndTimestamp = null;
+    }
+    if (updateDeltaStartTimestamp == null || updateDeltaStartTimestamp.isEmpty()) {
```
zzcclp: It's better use StringUtils.isEmpty()

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
```diff
@@ -468,4 +482,26 @@ public boolean isCarbonFormat() {
...
+      updateDeltaStartTimestamp = null;
+    }
+    if (updateStatusFileName == null || updateStatusFileName.isEmpty()) {
```
zzcclp: It's better use StringUtils.isEmpty()

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
```diff
@@ -468,4 +482,26 @@ public boolean isCarbonFormat() {
...
+      updateStatusFileName = null;
+    }
+    if (visibility == null || visibility.equals("true")) {
```
zzcclp: It's better use (StringUtils.isEmpty(visibility) || visibility.equals("true"))

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
```diff
@@ -468,4 +482,26 @@ public boolean isCarbonFormat() {
...
+      visibility = null;
+    }
+    if (fileFormat == null || fileFormat.equals(FileFormat.COLUMNAR_V3.toString())) {
```
zzcclp: It's better use (StringUtils.isEmpty(fileFormat) || fileFormat.equals(FileFormat.COLUMNAR_V3.toString()))

core/pom.xml
zzcclp: OK, I know why change this here.
Will Carbondata use 1.8 as default version?

core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatus.java
```diff
@@ -58,7 +58,7 @@
   /**
    * Segment is compacted
    */
-  @SerializedName("Compacted")
+  @SerializedName(value = "C", alternate = "Compacted")
```
akashrn5: why only for Success and Compacted? what about the Marked for delete, Partial Success, Insert In Progress and other status

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
jackylk: fixed

core/pom.xml
jackylk: I think it is better to change it as java 1.8 provides some good features that makes code cleaner

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
jackylk: fixed

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
jackylk: fixed

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
jackylk: fixed

core/src/main/java/org/apache/carbondata/core/statusmanager/LoadMetadataDetails.java
jackylk: fixed

core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatus.java
jackylk: In table status file, most of the entries are SUCCESS and COMPACTED, so I gave alias for these two only, for others I prefer to keep it the same for better readibility

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/858/

CarbonDataQA:
Build Failed  with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/868/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/868/

CarbonDataQA:
Build Failed  with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/877/

CarbonDataQA:
Build Failed with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/875/

CarbonDataQA:
Build Failed  with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/869/

CarbonDataQA:
Build Success with Spark 2.1.0, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.1/876/

core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatus.java
akashrn5: yes you are right, but i think since goal is to reduce size, better to change all the status, because when we have both status, like one with short form, one with full name do not look good, what do you think?

CarbonDataQA:
Build Success with Spark 2.3.2, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/884/

CarbonDataQA:
Build Success with Spark 2.2.1, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.2/882/

xubo245:
@zzcclp @akashrn5 This PR is already ok or not?

xubo245:
@jackylk Have you tested the Compatibility ？   Whether new code can read old data that was written by old version? like 1.5

jackylk:
@xubo245 It is ok to read table status file of old version, if new field name is not found, it will find old field name. (supported by GSON's @SerializedName(alternate) annotation)

xubo245:
LGTM

xubo245:
Thanks for your contribution!

PR #3577, opened 2020-01-14
[HOTFIX] Fix After alter table add column schemaOrdinal is not continuous

ajantha-bhat:
### Why is this PR needed?
 Consider a scenario with complex columns[c1 int, c2 array<int>, c3 array<int>] here schema ordinal of child columns are -1. so c1, c2, c3 has ordinal 0, 1, 2. After alter table add column "c4 int", its ordinal should be 3. But currently it is 6 (due to flat column schema).
This will give problem when this ordinal is  referred for some other functionality like rearrange projection columns.
 
make schema ordinal continuous by collecting previous schema ordinal max. 
    
 - No

 - No


CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/1627/

PR #3611, opened 2020-02-11
[CARBONDATA-3692] Support NoneCompression during loading data.

Pickupolddriver:
### Why is this PR needed?

In some cases, the data need to be uncompressed after loading into Carbondata file.
In the current version, the project do not support loading data without compression.

Provide a new Compressor as NoneCompressor implement the AbstractCompressor.
This compressor can be set by calling
CarbonProperties.getInstance().addProperty(CarbonCommonConstants.COMPRESSOR,"none");

Yes

Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
@@ -35,6 +35,7 @@ public class CompressorFactory {
   private final Map<String, Compressor> allSupportedCompressors = new HashMap<>();
 
   public enum NativeSupportedCompressor {
+    NONE("none",NoneCompressor.class),
     SNAPPY("snappy", SnappyCompressor.class),
     ZSTD("zstd", ZstdCompressor.class),
     GZIP("gzip", GzipCompressor.class);
--- a/core/src/main/java/org/apache/carbondata/core/datastore/compression/NoneCompressor.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/compression/NoneCompressor.java
@@ -1,0 +1,51 @@
+package org.apache.carbondata.core.datastore.compression;
+
+import java.io.IOException;
+
+public class NoneCompressor extends AbstractCompressor {
+
+    @Override
+    public String getName() {
+        return "none";
+    }
+
+    @Override
+    public byte[] compressByte(byte[] unCompInput) {
+        return unCompInput;
+    }
+
+    @Override
+    public byte[] compressByte(byte[] unCompInput, int byteSize) {
+        return unCompInput;
+    }
+
+    @Override
+    public byte[] unCompressByte(byte[] compInput) {
+        return compInput;
+    }
+
+    @Override
+    public byte[] unCompressByte(byte[] compInput, int offset, int length) {
+        return compInput;
+    }
+
+    @Override
+    public long rawUncompress(byte[] input, byte[] output) throws IOException {
+        throw new RuntimeException("Not implemented rawCompress for noneCompressor yet");
+    }
+
+    @Override
+    public long maxCompressedLength(long inputSize) {
+        return inputSize;
+    }
+
+    @Override
+    public int unCompressedLength(byte[] data, int offset, int length) {
+        throw new RuntimeException("Unsupported operation Exception");
+    }
+
+    @Override
+    public int rawUncompress(byte[] data, int offset, int length, byte[] output) {
+        throw new RuntimeException("Not implemented rawCompress for noneCompressor yet");
+    }
+}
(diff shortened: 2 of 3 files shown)


CarbonDataQA1:

kunal642:
Can you please explain the scenario where no-compression would be beneficial?

Pickupolddriver:
> Can you please explain the scenario where no-compression would be beneficial?

This NoneCompress Compressor will improve the speed of loading data from Flink to OBS File by trade-off space and IO in some cases. 

For example: when loading data from Flink to OBS, data needs to be compressed by Flink to temporary files and then decompressed by OBS. 
After adding the NoneCompressor, users can use the NoneCompressor load data without compress first and then decompress the temporary files.

ajantha-bhat:
@Pickupolddriver : Agree that it can improve the loading speed. But data will be 3x bigger. So, storage cost on OBS will be 3x more!

integration/spark-common-test/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithCompression.scala
```diff
@@ -272,6 +272,79 @@ class TestLoadDataWithCompression extends QueryTest with BeforeAndAfterEach with
...
+  test("test current none compressor on legacy store with snappy") {
+    CarbonProperties.getInstance().addProperty(CarbonCommonConstants.ENABLE_OFFHEAP_SORT, "true")
+    CarbonProperties.getInstance().addProperty(CarbonCommonConstants.COMPRESSOR, "snappy")
+    createTable()
+    loadData()
+
+    CarbonProperties.getInstance().addProperty(CarbonCommonConstants.ENABLE_OFFHEAP_SORT, "true")
+    CarbonProperties.getInstance().addProperty(CarbonCommonConstants.COMPRESSOR, "none")
+    loadData()
+    checkAnswer(sql(s"SELECT count(*) FROM $tableName"), Seq(Row(16)))
```
kunal642: Can you change select count(*) to select * so that actual data is validated

kunal642: Can you change select count(*) to select * so that actual data is validated

Pickupolddriver: Sure

Pickupolddriver:
> @Pickupolddriver : Agree that it can improve the loading speed. But data will be 3x bigger. So, storage cost on OBS will be 3x more!

Data would be processed after loaded to OBS. So if we could provide a NonCompressor, it could avoid the data being compressed and then uncompressed. And the uncompressed data would be deleted after processed in OBS.

core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
```diff
@@ -35,6 +35,7 @@
   private final Map<String, Compressor> allSupportedCompressors = new HashMap<>();
 
   public enum NativeSupportedCompressor {
+    NONE("none",NoneCompressor.class),
```
jackylk: add space after `,`

core/src/main/java/org/apache/carbondata/core/datastore/compression/NoneCompressor.java
```diff
@@ -0,0 +1,51 @@
+package org.apache.carbondata.core.datastore.compression;
```
jackylk: please add license header as other source file

core/src/main/java/org/apache/carbondata/core/datastore/compression/NoneCompressor.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.IOException;
+
+public class NoneCompressor extends AbstractCompressor {
```
jackylk: You can add description to this class that it does not perform any compression

integration/spark-common-test/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithCompression.scala
Pickupolddriver: So you want to change all the test cases in this class from select count(*) to *?

core/src/main/java/org/apache/carbondata/core/datastore/compression/CompressorFactory.java
jackylk: suggest change to NONE("nocompress", NoneCompressor.class)
`nocompress` will be appended to the data file name

ajantha-bhat:
add to whitelist

integration/spark-common-test/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataWithCompression.scala
kunal642: No, just change the newly added test cases to select *

QiangCai:
please rebase

PR #3637, opened 2020-02-24
[CARBONDATA-3721][CARBONDATA-3590] Optimize Bucket Table

Zhangshunyu:
### Why is this PR needed?
 Support Bucket Table consistent with spark to improve the join performance by avoid shuffle for bucket column. The same time, fix bugs about load/compact query of bucket.
 
Support Bucket Table and consistent with spark to improve the join performance by avoid shuffle for bucket column. Fix bugs also.
1. For create table, ddl support both tblproperties and clustered by like hive.
2. For loading, fix some problems in loading when bucket column specified, make it clusterd into diff files based on bucket column.
3. For query, the hash impl should either keep the same for a given value or keep same with parquet table, so that the join result of diff bucket tables give correct result.  By the way, the hash impl is configurable.
4. For compaction, group the block files based on bucket id, the data should hash into diff carbondata files also, otherwise the query flow will group the files based on bucket num, all data compacted will com into 1 partition and the join result will mismatch, the performace will very slow.
5. For tests, add 19 test cases in TableBucketingTestCase
    
 - No

 - Yes

 ### Performance

    3 million data each table with distinct 60000， 50 buckets join				

    select count(*) from
			(select t1.*, t2.*
			from t7 t1, t7_ t2
			where t1.name = t2.name) temp
  +---------+
  |count(1) |
  +---------+
  |190781696|
  +---------+
  		
1. When there are buckets in carbon, The first query is 30% higher and the second query is 50%-60% higher than no bucket table.
2. The carbon of no bucket is 30% faster than the parquet of no bucket
3. Carbon with bucket and parquet with bucket are slow for the first time and 10% faster for the second time, roughly equal.
4. Both buckets are faster than their own no bucket tables.

	format	    bucket 2nd cost	    bucket 1st cost	    no-bucket 2nd cost	no-bucket 1st cost
	carbon	        2126 ms	         3453 ms	    		        4952 ms	    			5092 ms
	parquet		2484 ms	    	 2770 ms 			6327 ms	    			8779 ms
	orc	        	2687 ms	   	  3765 ms	    		8678 ms	    			6567 ms


CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/427/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2127/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/432/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2132/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/447/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/449/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2149/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/456/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2157/

Zhangshunyu:
retest this please

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/459/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2160/

processing/src/main/java/org/apache/carbondata/processing/loading/partition/impl/BucketMurmur3HashPartitionerImpl.java
```diff
@@ -0,0 +1,181 @@
...
+    for (int i = 0; i < indexes.size(); i++) {
+      DataType dataType = columnSchemas.get(i).getDataType();
+      if (dataType == DataTypes.LONG || dataType == DataTypes.DOUBLE) {
+        hashes[i] = new LongHash(indexes.get(i));
+      } else if (dataType == DataTypes.SHORT || dataType == DataTypes.INT ||
+          dataType == DataTypes.FLOAT || dataType == DataTypes.BOOLEAN) {
+        hashes[i] = new IntegralHash(indexes.get(i));
+      } else if (DataTypes.isDecimal(dataType)) {
+        hashes[i] = new DecimalHash(indexes.get(i));
+      } else if (dataType == DataTypes.TIMESTAMP) {
```
Indhumathi27: What about Hash for Date Type?

Zhangshunyu: @Indhumathi27 if use hash for datatype the hash value will diff from spark, and join result will mismatch with parquet etc.

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/506/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2205/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/512/

ravipesala:
@Zhangshunyu Bucketing is already supported in Carbon.  I wonder why all this code is added again to support it.  If there are any issues if we are facing please put the testcases first which are not working or raise a jira.

Zhangshunyu:
@ravipesala pls check all the new testcases added in TableBucketingTestCase and the comment i added in the pr desc.
1. all data stored into 1 file
2. join with parquet return wrong result
3. after compaction it will store into file of bucket id 0
4. new insert flow not work in
5. the others pls check testcases added

CarbonDataQA1:
Build Failed  with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/530/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2230/

CarbonDataQA1:
Build Failed  with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/533/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2233/

CarbonDataQA1:
Build Failed  with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/539/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2238/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/543/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2243/

ravipesala:
@Zhangshunyu It was a supported feature earlier but it is bad that code got removed some time back. Anyway, spark changed the hashing technique on creating buckets so we cannot rely on our own hashing anymore. 
I see a lot of code got copied spark to just get the hashing. it is not recommended to do so as in the future if they change it will again break. Even they follow industry-standard murmur hash to do the hash. So please use the guava library and do the murmur hashing. Please don't copy the code unnecessarily from the spark.

ravipesala:
@Zhangshunyu other way is to let the spark do the bucketing like how the partitioner is implemented. In fact, we can add the bucketing directly into the partition flow. Not much changes needed in that case.

Zhangshunyu:
> @Zhangshunyu other way is to let the spark do the bucketing like how the partitioner is implemented. In fact, we can add the bucketing directly into the partition flow. Not much changes needed in that case.

@ravipesala is guava murmur hash the same as spark using?

> @Zhangshunyu It was a supported feature earlier but it is bad that code got removed some time back. Anyway, spark changed the hashing technique on creating buckets so we cannot rely on our own hashing anymore.
> I see a lot of code got copied spark to just get the hashing. it is not recommended to do so as in the future if they change it will again break. Even they follow industry-standard murmur hash to do the hash. So please use the guava library and do the murmur hashing. Please don't copy the code unnecessarily from the spark.

spark using guava hash but not all the same like guava's impl, as for the changes in future of spark, if we want to keep same hash code as spark, maybe we can depend on spark-unsafe jar directly base on spark-version just like carbon depend on diff spark version.

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/545/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2245/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/547/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2247/

core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -2379,4 +2379,18 @@ private CarbonCommonConstants() {
...
+   */
+  public static final String BUCKET_HASH_METHOD = "bucket_hash_method";
+  public static final String BUCKET_HASH_METHOD_DEFAULT = "spark_hash_expression";
+  public static final String BUCKET_HASH_METHOD_SPARK_EXPRESSION = "spark_hash_expression";
+  public static final String BUCKET_HASH_METHOD_NATIVE = "native";
+
+  /**
+   * bucket properties
+   */
+  public static final String BUCKET_COLUMNS = "bucketcolumns";
```
jackylk: Is these for table properties? suggest to change to "bucket_columns"

jackylk: Can we follow the bucket table syntax from hive?

integration/flink/src/test/scala/org/apache/carbon/flink/TestCarbonWriter.scala
```diff
@@ -20,20 +20,23 @@ package org.apache.carbon.flink
...
-
 import org.apache.carbondata.core.datastore.impl.FileFactory
 import org.apache.carbondata.core.util.CarbonProperties
 import org.apache.carbondata.core.util.path.CarbonTablePath
+import org.apache.spark.sql.execution.exchange.Exchange
 
 class TestCarbonWriter extends QueryTest {
 
   val tableName = "test_flink"
+  val tableName2 = "insert_bucket_table"
```
jackylk:
```suggestion
  val bucketTableName = "insert_bucket_table"
```

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -584,36 +591,53 @@ class CarbonMergerRDD[K, V](
...
-          val multiBlockSplit = if (null == rangeColumn || singleRange) {
-            new CarbonMultiBlockSplit(
-              taskInfo.getCarbonInputSplitList,
-              Array(nodeName))
-          } else {
-            var splitListForRange = new util.ArrayList[CarbonInputSplit]()
-            new CarbonMultiBlockSplit(
-              splitListForRange,
-              Array(nodeName))
+    if (bucketInfo != null) {
```
jackylk: extract these new logic into a private function, so not to make internalCompute too big

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -91,6 +91,7 @@ class CarbonMergerRDD[K, V](
   var singleRange = false
   var expressionMapForRangeCol: util.Map[Integer, Expression] = null
   var broadCastSplits: Broadcast[CarbonInputSplitWrapper] = null
+  val bucketInfo = carbonLoadModel.getCarbonDataLoadSchema.getCarbonTable.getBucketingInfo
```
jackylk: Seems executor side does not need this variable, so move it into internalCompute to make it as a local variable

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonScanRDD.scala
```diff
@@ -96,7 +96,7 @@ class CarbonScanRDD[T: ClassTag](
 
   private var directFill = false
 
-  private val bucketedTable = tableInfo.getFactTable.getBucketingInfo
+  private val bucketInfo = tableInfo.getFactTable.getBucketingInfo
```
jackylk: Does executor side need this? Can we make it as a local variable in internalCompute

processing/pom.xml
```diff
@@ -45,6 +45,11 @@
       <artifactId>spark-sql_${scala.binary.version}</artifactId>
       <version>${spark.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
```
jackylk: Should try to avoid depending on spark in core modules including: carbondata-core, carbondata-hadoop, carbondata-processing
Can we avoid adding this?

core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
Zhangshunyu: @jackylk ok, will change the property name, support both property and syntax from hive already.

processing/pom.xml
Zhangshunyu: @jackylk if want to keep correct join result with parquet bucket tables, need to use same methods to hash the data of each datatype, so the code is needed.
1. copy the code from spark, but there are about 2,000 lines and if we copy the code, once spark change them we need to change together, its not a good choice, more details pls check the conversations above.
2. depend on spark-unsafe jar, we just depend 1 jar of spark and the changes of diff spark version don't have effect on us since we use it by version control in pom.

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/570/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2274/

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/600/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2307/

PR #3642, opened 2020-02-28
[CARBONDATA-3725]fix concurrent creation of Materialized Views issue

akashrn5:
### Why is this PR needed?
When two create MVs are running concurrently in two different sessions with two different databases, then in any create fails in any session, other sessions's current database is changed and corresponding queries fails with table not found exception.
 
When the session is changed we need to clear the datamap catalog map which may lead to this problem. Initially it was handled only for query and other flows, it wasnt handled in create MV flow. Handled same, make proper synchronisations to avoid these problems. 
    
 - No
 - Yes. (added a isSchemaRegistered in DatamapCatalog Interface, which is required to avoid multiple registration of same datamap schemas)

 - No, existing test cases will take care

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapCatalog.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapCatalog.java
@@ -48,4 +48,9 @@ public interface DataMapCatalog<T> {
    */
   void refresh();
 
+  /**
+   * This checks whether the datamapSchema is already registered
+   */
+  Boolean isSchemaAlreadyRegistered(String dataMapName);
+
 }
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapCatalog.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapCatalog.java
@@ -51,6 +51,6 @@ public interface DataMapCatalog<T> {
   /**
    * This checks whether the datamapSchema is already registered
    */
-  Boolean isSchemaAlreadyRegistered(String dataMapName);
+  Boolean isMVExists(String mvName);
 
 }
--- a/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
@@ -249,7 +249,7 @@ public final class DataMapStoreManager {
         dataMapCatalog.registerSchema(dataMapSchema);
       }
     } else {
-      if (!dataMapCatalog.isSchemaAlreadyRegistered(dataMapSchema.getDataMapName())) {
+      if (!dataMapCatalog.isMVExists(dataMapSchema.getDataMapName())) {
         dataMapCatalog.registerSchema(dataMapSchema);
       }
     }
@@ -275,8 +275,10 @@ public final class DataMapStoreManager {
    * @param providerName
    * @return
    */
-  public synchronized DataMapCatalog getDataMapCatalog(DataMapProvider dataMapProvider,
-      String providerName, boolean clearCatalogs) throws IOException {
+  public synchronized DataMapCatalog getDataMapCatalog(
+      DataMapProvider dataMapProvider,
+      String providerName,
+      boolean clearCatalogs) throws IOException {
     // This method removes the datamapCatalog for the corresponding provider if the session gets
     // refreshed or updated
     if (clearCatalogs) {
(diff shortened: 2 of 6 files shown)


CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/516/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2216/

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -269,26 +275,23 @@ public synchronized void unRegisterDataMapCatalog(DataMapSchema dataMapSchema) {
    * @param providerName
    * @return
    */
-  public synchronized DataMapCatalog getDataMapCatalog(
-      DataMapProvider dataMapProvider,
-      String providerName) throws IOException {
+  public synchronized DataMapCatalog getDataMapCatalog(DataMapProvider dataMapProvider,
```
jackylk: move `DataMapProvider dataMapProvider,` to next line

mv/core/src/main/scala/org/apache/carbondata/mv/rewrite/SummaryDatasetCatalog.scala
```diff
@@ -107,12 +110,19 @@ private[mv] class SummaryDatasetCatalog(sparkSession: SparkSession)
...
-      // here setting back to current database of current session, because if the actual query
-      // contains db name in query like, select db1.column1 from table and current database is
-      // default and if we drop the db1, still the session has current db as db1.
-      // So setting back to current database.
-      setCurrentDataBase(currentDatabase)
+      val mvPlan = try {
+        MVParser.getMVPlan(dataMapSchema.getCtasQuery, sparkSession)
+      } catch {
+        case ex: Exception =>
+          LOGGER.error("Error executing the updated query during register datamap schema", ex)
```
jackylk:
```suggestion
          LOGGER.error("Error executing the updated query during register MV schema", ex)
```

core/src/main/java/org/apache/carbondata/core/datamap/DataMapCatalog.java
```diff
@@ -48,4 +48,9 @@
    */
   void refresh();
 
+  /**
+   * This checks whether the datamapSchema is already registered
+   */
+  Boolean isSchemaAlreadyRegistered(String dataMapName);
```
jackylk:
```suggestion
  Boolean isMVExists(String mvName);
```

akashrn5: done

mv/core/src/main/scala/org/apache/carbondata/mv/rewrite/SummaryDatasetCatalog.scala
akashrn5: done

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
akashrn5: done

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/556/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2256/

jackylk:
LGTM

core/src/main/java/org/apache/carbondata/core/datamap/DataMapStoreManager.java
```diff
@@ -271,24 +277,23 @@ public synchronized void unRegisterDataMapCatalog(DataMapSchema dataMapSchema) {
    */
   public synchronized DataMapCatalog getDataMapCatalog(
       DataMapProvider dataMapProvider,
-      String providerName) throws IOException {
+      String providerName,
+      boolean clearCatalogs) throws IOException {
```
jackylk: This is not clear, maybe it is better to add a separate API to clear the catalog

akashrn5: initialy we had a separate API for clear, i have removed in this PR because it was synchronized and instead of calling that API everytime if session changes, it better if we handle here where we intialize catalogs and its synchronized also, so it takes care of all.

ajantha-bhat: @jackylk : This is ok, clear is just assigning to null and multiple function calls in concurrent scenario always better to avoid.
However, initializeDataMapCatalogs can be optimized in the future with more testing. we have some duplicate code there to handle the concurrent scenario

jackylk:
LGTM

PR #3644, opened 2020-02-29
[CARBONDATA-3727] Fix NPE of sort merger

Zhangshunyu:
### Why is this PR needed?
 Fix NPE of sort merger in range sort cases.
in ParallelReadMergeSorterWithColumnRangeImpl, finalMerger.startFinalMerge will check whether
filesToMerge.size() == 0, if true it wil return without createRecordHolderQueue and as a result it is null, and finally the code will call finalMerger.hasNext() to check whether the iterator has next, but the recordHolderHeapLocal is null, it will throw NPE.
Add null check if recordHolderHeapLocal is null, it means not created and need merged files == 0, so it also means no next value.
    
 - No

 - No

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/sort/sortdata/SingleThreadFinalSortFilesMerger.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/sort/sortdata/SingleThreadFinalSortFilesMerger.java
@@ -306,7 +306,7 @@ public class SingleThreadFinalSortFilesMerger extends CarbonIterator<Object[]> {
    * @return more element is present
    */
   public boolean hasNext() {
-    return this.recordHolderHeapLocal.size() > 0;
+    return this.recordHolderHeapLocal != null && this.recordHolderHeapLocal.size() > 0;
   }
 
   public void close() {


CarbonDataQA1:
Build Failed  with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/535/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2235/

jackylk:
LGTM

ajantha-bhat:
retest this please

processing/src/main/java/org/apache/carbondata/processing/sort/sortdata/SingleThreadFinalSortFilesMerger.java
```diff
@@ -306,7 +306,7 @@ private IntermediateSortTempRow getSortedRecordFromFile() throws CarbonDataWrite
    * @return more element is present
    */
   public boolean hasNext() {
-    return this.recordHolderHeapLocal.size() > 0;
+    return this.recordHolderHeapLocal != null && this.recordHolderHeapLocal.size() > 0;
```
ajantha-bhat: good. can you please add a test for this scenario ?

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.4/540/

processing/src/main/java/org/apache/carbondata/processing/sort/sortdata/SingleThreadFinalSortFilesMerger.java
Zhangshunyu: @ajantha-bhat after bucket table optimized and working fine can this NPE problem be exposed, so the test cases added with bucket table feature test cases in another pr based on bucket working. By the way, from code anylyse we alse can find this parameter might be null without npe checking.

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2240/

processing/src/main/java/org/apache/carbondata/processing/sort/sortdata/SingleThreadFinalSortFilesMerger.java
ajantha-bhat: Then I suggest to keep this change in bucketing optimization PR itself as it has a testcase to cover this scenario.

Zhangshunyu: @ajantha-bhat ok, will close this pr and add the change into bucket table feature.

PR #3680, opened 2020-03-26
[CARBONDATA-3766] Fixed desc formatted and show segment data size issues

kunal642:
### Why is this PR needed?
For added segments the index size was being displayed in show segments instead of data size. 
 
Take data soze from indexsize variable and show in show segments.
   
 - No

 - No
 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
@@ -527,10 +527,11 @@ public class SegmentFileStore {
         for (LoadMetadataDetails detail : listOfLoadFolderDetailsArray) {
           // if the segments is in the list of marked for delete then update the status.
           if (segmentId.equals(detail.getLoadName())) {
+            detail.setLoadEndTime(System.currentTimeMillis());
             detail.setSegmentFile(segmentFile);
             if (segmentStatus != null) {
               HashMap<String, Long> dataSizeAndIndexSize =
-                  CarbonUtil.getDataSizeAndIndexSize(segmentFileStore);
+                  CarbonUtil.getDataSizeAndIndexSize(segmentFileStore, detail.isCarbonFormat());
               detail.setDataSize(
                   dataSizeAndIndexSize.get(CarbonCommonConstants.CARBON_TOTAL_DATA_SIZE)
                       .toString());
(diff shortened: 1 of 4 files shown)


CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/855/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2562/

kunal642:
@akashrn5 @jackylk @ajantha-bhat Please review

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
```diff
@@ -527,10 +527,11 @@ public static boolean updateTableStatusFile(CarbonTable carbonTable, String segm
         for (LoadMetadataDetails detail : listOfLoadFolderDetailsArray) {
           // if the segments is in the list of marked for delete then update the status.
           if (segmentId.equals(detail.getLoadName())) {
+            detail.setLoadEndTime(System.currentTimeMillis());
```
akashrn5, line 530: why we need this change? is it sepicific to non carbon segment?

akashrn5:
retest this please

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/941/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2651/

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
kunal642: yes

akashrn5: is this common code to both carbon and non carbon segments? if it is  then better to set it only in non carbon segments case.

kunal642: I think we should not add external segments check for common things, it should be okay if we set here because we are not using this end time in any other logic/decision making

akashrn5:
LGTM

PR #3682, opened 2020-03-26
[CARBONDATA-3753] optimize double/float stats collector

ajantha-bhat:
### Why is this PR needed?
For every double/float column's value. we call 
`PrimitivePageStatsCollector.getDecimalCount(double value)`
problem is, here we create new bigdecimal object and plain string object every time.
Which leads in huge memory usage during insert.

Create only Bigdecimal object and use scale from that. 
    
 - No
 
 - No
 
Before the change:
Screenshot from 2020-03-26 16-45-12

After the change:
Screenshot from 2020-03-26 16-51-31

There is about 5% improvement in insert for the TPCH lineitem table with10GB data without any change in store size.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/page/statistics/PrimitivePageStatsCollector.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/page/statistics/PrimitivePageStatsCollector.java
@@ -233,20 +233,18 @@ public class PrimitivePageStatsCollector implements ColumnPageStatsCollector, Si
 
   /**
    * Return number of digit after decimal point
-   * TODO: it operation is costly, optimize for performance
    */
   private int getDecimalCount(double value) {
     int decimalPlaces = 0;
     try {
-      String strValue = BigDecimal.valueOf(Math.abs(value)).toPlainString();
-      int integerPlaces = strValue.indexOf('.');
-      if (-1 != integerPlaces) {
-        decimalPlaces = strValue.length() - integerPlaces - 1;
+      BigDecimal decimalValue = BigDecimal.valueOf(value);
+      decimalPlaces = decimalValue.scale();
+      if (decimalPlaces == 1) {
         // If decimal places are one and it is just zero then treat the decimal count a zero.
-        if (decimalPlaces == 1) {
-          if (strValue.substring(integerPlaces + 1, strValue.length()).equals(ZERO_STRING)) {
-            decimalPlaces = 0;
-          }
+        // note: here toString() uses stringCache of BigDecimal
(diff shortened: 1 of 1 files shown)


CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/859/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2567/

ajantha-bhat:
@jackylk , @ravipesala : please check

brijoobopanna:
retest this please

CarbonDataQA1:
Build Success with Spark 2.4.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/864/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2572/

kunal642:
LGTM

core/src/main/java/org/apache/carbondata/core/datastore/page/statistics/PrimitivePageStatsCollector.java
```diff
@@ -233,20 +233,18 @@ public void update(long value) {
...
-   * TODO: it operation is costly, optimize for performance
    */
   private int getDecimalCount(double value) {
     int decimalPlaces = 0;
     try {
-      String strValue = BigDecimal.valueOf(Math.abs(value)).toPlainString();
-      int integerPlaces = strValue.indexOf('.');
-      if (-1 != integerPlaces) {
-        decimalPlaces = strValue.length() - integerPlaces - 1;
+      BigDecimal decimalValue = BigDecimal.valueOf(value);
```
QiangCai, line 240: better to write code to implement it.
not required to use BigDecimal.

ajantha-bhat: Actually double will not always be like `xx.yyy`, it will be having `exponent` also. 
So, Bigdecimal already converts to string and do that logic. 

May be next version we can reduce further by removing big decimal and copying API inside big decimal to do that.

ajantha-bhat: Also we need something without converting to string

QiangCai:
LGTM

PR #3717, opened 2020-04-16
[CARBONDATA-3780]Old store read compatibility issue for Index handler column

VenuReddy2103:
### Why is this PR needed?
 Table created in older version do not have indexColumn property in columnSchema. Newer versions have indexColumn property in columnSchema with default value as false. Due to this, when Old store is read, index handler column which is hidden to user is also treated as normal schema column and can perform operations like alter table on the index column.
 
During schema read, if the index_handler property is present, check if column name matches to index handler name and set indexColumn property value to true.
    
 - No

 - No

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/converter/ThriftWrapperSchemaConverterImpl.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/converter/ThriftWrapperSchemaConverterImpl.java
@@ -605,6 +605,14 @@ public class ThriftWrapperSchemaConverterImpl implements SchemaConverter {
           .setLocalDictColumnsToWrapperSchema(listOfColumns, externalTableSchema.tableProperties,
               externalTableSchema.tableProperties
                   .get(CarbonCommonConstants.LOCAL_DICTIONARY_ENABLE));
+      String handler = externalTableSchema.tableProperties.get(CarbonCommonConstants.INDEX_HANDLER);
+      if (handler != null) {
+        for (ColumnSchema columnSchema : listOfColumns) {
+          if (handler.equalsIgnoreCase(columnSchema.getColumnName())) {
+            columnSchema.setIndexColumn(true);
+          }
+        }
+      }
     }
     wrapperTableSchema.setListOfColumns(listOfColumns);
     wrapperTableSchema.setSchemaEvolution(


CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2756/

jackylk:
Isn't indexColumn property in columnSchema is optional, why need to add this check?

core/src/main/java/org/apache/carbondata/core/metadata/converter/ThriftWrapperSchemaConverterImpl.java
```diff
@@ -605,6 +605,14 @@ public TableSchema fromExternalToWrapperTableSchema(
           .setLocalDictColumnsToWrapperSchema(listOfColumns, externalTableSchema.tableProperties,
               externalTableSchema.tableProperties
                   .get(CarbonCommonConstants.LOCAL_DICTIONARY_ENABLE));
+      String handler = externalTableSchema.tableProperties.get(CarbonCommonConstants.INDEX_HANDLER);
+      if (handler != null) {
+        for (ColumnSchema columnSchema : listOfColumns) {
+          if (handler.equalsIgnoreCase(columnSchema.getColumnName())) {
```
jackylk, line 611: Why are you checking column name against handler? 
Either of these is not a proper name

VenuReddy2103: index column was created with index_handler property value as its name during create table in `processIndexProperty` method. So matching handler value against the column name.

VenuReddy2103:
> Isn't indexColumn property in columnSchema is optional, why need to add this check, or this check is in wrong place (should have check when this property is being used)

Yes, It is optional. But, Table created in older version do not have indexColumn property in columnSchema.(schema.thrift). We had considered columnSchema.schemaOrdinal being -1 as index column instead of defining a new property. 
        But in  the later version, we have defined indexColumn property in columnSchema with default value as false. Due to this, when Old store is read, index handler column which is hidden to user is also treated as normal schema column(since indexColumn property default value is false) and user can perform operations like alter table on the index column and create inconsistency.
        Thats why, when we read the the schema, we update the indexColumn field in columnSchema to true if it is index column.

core/src/main/java/org/apache/carbondata/core/metadata/converter/ThriftWrapperSchemaConverterImpl.java
ajantha-bhat: I strongly agree with @jackylk, we already have many index like SI, bloom, lucene. 
This index handler and isIndex column is confusing. 
so we should have something like `spatialIndexHandler` and `isSpatialIndexColumn` as we use these interface for geo spatial / spatial only now.

But renaming thrift will again create compatibility issue if somebody started geospatial already.

@VenuReddy2103 : we can have that renaming change in another PR.

jackylk: @ajantha-bhat I think this feature is still alpha feature in 2.0, 1.6 does not include this feature. For long term benefit, I still think these names should be changed to make it clear to user and developers

PR #3733, opened 2020-04-30
[CARBONDATA-3789]Fix cache issue in case of compaction failure in compaction post listeners

akashrn5:
### Why is this PR needed?
 Consider a scenario where the post-event Is called in case of compaction and then the listener is basically selecting the ongoing compacted segment data and loading SI table load after main table compaction. During that time the cache is loaded, if in case this is failed, then the cache is still present but the actual segment isn't, so which will lead to consecutive failures.
 
So, when the failure happens, the cache should be cleared either in the index server or in the driver cache.
    
 - No

 ### Is any new test case added?
 - No

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
@@ -18,7 +18,7 @@
 package org.apache.carbondata.spark.rdd
 
 import java.util
-import java.util.List
+import java.util.{Collections, List}
 import java.util.concurrent.ExecutorService
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2904/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1187/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1190/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/2907/

ajantha-bhat:
@akashrn5 : Issue is happening when a segment is removed from main table but cache is not cleaned right? 

When the SI table compaction disable why not just disable SI, why need to remove whole segment of main table ?

akashrn5:
> @akashrn5 : Issue is happening when a segment is removed from main table but cache is not cleaned right?
> 
> When the SI table compaction disable why not just disable SI, why need to remove whole segment of main table ?

@ajantha-bhat , once the main table compaction is finished, say example, 0.1 for segments 1, 2, then 0.1 segment data of the main table is selected and loaded to SI table's 0.1 segment. So if while loading 0.1 to SI fails, then 0.1 entry won't be present in the main table, and the cache will have this entry for the main table.

So when next time again the same 0.1 is compacted for 0,1 segments as it was failed before, in SI it will fire a select query to 0.1 of the main table, since the cache is not cleared, it will try for old carbondata files and say file not found.

ajantha-bhat:
> So if while loading 0.1 to SI fails, then 0.1 entry won't be present in the main table

This is what I mean, why we need to fail the main table compaction when SI compaction fails ?

akashrn5:
> we can load SI separately from compacted segments later from rebuild/refresh command

we do not have any refresh command for SI FYI. as we haven't made it like the other indexed on carbon table. So this is the old behavior.

ajantha-bhat:
compacting whole main table again will take huge time, as maintable already compaction is success we should handle like other datamaps.

akashrn5:
> compacting whole main table again will take huge time, as maintable already compaction is success we should handle like other datamaps.

main table compaction is success functionally, but not yet committed as success segment in metadata. So handling like other datamaps is altogether new requirement. cannot handle here.

ajantha-bhat:
ok. LGTM

PR #3788, opened 2020-06-05
[CARBONDATA-3844]Fix scan the relevant database instead of scanning all

xiaohui0318:
### Why is this PR needed?
 When a query is executed, all databases of metadata are scanned, and if the number of databases is very large, the execution time increases for a long time each time, reducing query performance
 
scan the relevant database instead of scanning all
    
 - No

 - No
 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/view/MVCatalog.java
+++ b/core/src/main/java/org/apache/carbondata/core/view/MVCatalog.java
@@ -17,6 +17,8 @@
 
 package org.apache.carbondata.core.view;
 
+import java.util.List;
+
 import org.apache.carbondata.core.metadata.schema.table.RelationIdentifier;
 
 /**
@@ -27,9 +29,9 @@ import org.apache.carbondata.core.metadata.schema.table.RelationIdentifier;
 public interface MVCatalog<T> {
 
   /**
-   * List all registered valid schema catalogs
+   * List list databases valid schema catalogs
    */
-  T[] getValidSchemas();
+  T[] getValidSchemas(List<String> databases);
 
   /**
    * Register schema to the catalog.
--- a/core/src/main/java/org/apache/carbondata/core/view/MVManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/view/MVManager.java
@@ -234,6 +234,23 @@ public abstract class MVManager {
     }
     return statusDetails;
   }
+  /**
+   * Get databases list mv status details
+   */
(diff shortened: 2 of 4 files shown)


CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3136/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1412/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1414/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3138/

niuge01:
May be create an mv in database A as select from a table in database B, when query the table, we should check all mv on this table, not only in database B.

xiaohui0318:
> May be create an mv in database A as select from a table in database B, when query the table, we should check all mv on this table, not only in database B.

We have a cluster with more than 1,500 databases, and it takes more than 20 seconds to scan each select statement, no matter how simple the SQL. I wonder it should be placed in a specified database?

VenuReddy2103:
If MV is in different database than that of its source tables, then we will not be able to rewrite the plan. Consider below inner join example.
`spark.sql("create database db1")
spark.sql("create database db2")
spark.sql("create database db3")
spark.sql("use db1")
spark.sql("create table db1_table(a int, b int) stored as carbondata")
spark.sql("insert into db1_table select 1, 2")
spark.sql("use db2")
spark.sql("create table db2_table(i int, j int) stored as carbondata")
spark.sql("insert into db2_table select 1, 4")
spark.sql("use db3")
spark.sql("create materialized view db3_mv as select t1.a,t2.i from db1.db1_table t1,db2.db2_table t2 where t1.a=t2.i")
spark.sql("explain select t1.a, t2.i from db1.db1_table t1, db2.db2_table t2 where t1.a=t2.i").show(100, false)`

if we get `getValidSchemas` only from `db1` and `db2` in `hasSuitableMV()`. We don't find one and do not rewrite the plan.

vikramahuja1001:
@VenuReddy2103 , maybe such a test case can be added in the code

PR #3818, opened 2020-06-30
[Carbondata-3883] Added filtering for the deleted rows for local dictionary fields

Nitin-Kashyap:
### Why is this PR needed?
 In Presto integration when reading records for table with string fields and local dictionary enabled gives improper result.
 
Added a CarbondataColumnVectorWrapper for the dictionary vector for column for which internally create dictionary vector.
Wrapper is used to filter the deleted records.
    
 - No

 - No

### DIFF
--- a/integration/presto/src/main/java/org/apache/carbondata/presto/CarbonColumnVectorWrapper.java
+++ b/integration/presto/src/main/java/org/apache/carbondata/presto/CarbonColumnVectorWrapper.java
@@ -28,6 +28,7 @@ import org.apache.carbondata.core.scan.scanner.LazyPageLoader;
 public class CarbonColumnVectorWrapper implements CarbonColumnVector {
 
   private CarbonColumnVectorImpl columnVector;
+  private CarbonColumnVectorWrapper dictionaryVectorWrapper;
 
   private boolean[] filteredRows;
 
@@ -40,6 +41,11 @@ public class CarbonColumnVectorWrapper implements CarbonColumnVector {
   public CarbonColumnVectorWrapper(CarbonColumnVectorImpl columnVector, boolean[] filteredRows) {
     this.columnVector = columnVector;
     this.filteredRows = filteredRows;
+
+    if (columnVector.getDictionaryVector() != null) {
+      dictionaryVectorWrapper = new CarbonColumnVectorWrapper(
+              (CarbonColumnVectorImpl)columnVector.getDictionaryVector(), filteredRows);
+    }
   }
 
   @Override
@@ -273,7 +279,7 @@ public class CarbonColumnVectorWrapper implements CarbonColumnVector {
 
   @Override
   public CarbonColumnVector getDictionaryVector() {
-    return this.columnVector.getDictionaryVector();
+    return this.dictionaryVectorWrapper;
   }
 
   @Override


CarbonDataQA1:

brijoobopanna:
add to whitelist

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3281/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1544/

ajantha-bhat:
@Nitin-Kashyap : Can you please add a test here in `PrestoAllDataTypeLocalDictTest.scala` for this scenario. May be just add one more query for existing table local dictionary column that can cause this problem.

integration/presto/src/main/java/org/apache/carbondata/presto/CarbonColumnVectorWrapper.java
```diff
@@ -273,7 +279,7 @@ public boolean hasDictionary() {
 
   @Override
   public CarbonColumnVector getDictionaryVector() {
-    return this.columnVector.getDictionaryVector();
+    return this.dictionaryVectorWrapper;
```
ajantha-bhat, line 282: Currently in our code it is used only in `SliceStreamReader#buildBlock`.
I don't see any problem here.

Also in your issue JIRA,  saw that it is in delete scenario, **our version of carbon doesn't support insert delete from presto yet**. Feel free to contribute.

Nitin-Kashyap: Thanks for the review.
JIRA issue I created had insert and delete carried out using spark integration; and read using presto integration.

I'm not sure about the applicability part, I trust the contributes and maintainers wisdom in here :)
However, I observed the case while my local verification and suggested the fix after verification.
Feel free to use or reject.

Nitin-Kashyap: I reviewed the `SliceStreamReader::buildBlock`, I dont see how it can filter out the deleted record when `carbon.push.rowfilters.for.vector=true` property is set.

ajantha-bhat: ok, thanks for the reply. I got confused with presto delete update. 
Please add a test case that can reproduce issue in the suite.

Nitin-Kashyap:
> @Nitin-Kashyap : Can you please add a test here in `PrestoAllDataTypeLocalDictTest.scala` for this scenario. May be just add one more query for existing table local dictionary column that can cause this problem.

Since the test step required me to insert and delete record using Spark-Integration; The same feature is unavailable in Presto-Integration; I'm not sure how to add the same in `PrestoAllDataTypeLocalDictTest.scala`.

Please suggest, I would be happy to add!

ajantha-bhat:
> Since the test step required me to insert and delete record using Spark-Integration; The same feature is unavailable in Presto-Integration; I'm not sure how to add the same in PrestoAllDataTypeLocalDictTest.scala.

Please suggest, I would be happy to add!

we currently manually create store as presto module is independent of spark. 
Need add spark dependency in test module to create store. We can take this as separate activity.

ajantha-bhat:
LGTM. Thanks for the contribution !

PR #3764, opened 2020-05-12
[CARBONDATA-3820] Fix CDC failure when sort columns present in source dataframe

marchpure:
### Why is this PR needed?
While merging into table with sortcolumns in the CDC Flow. The following exception will be throwed:
"column: id specified in sort columns does not exist in schema".
Root cause is that we use TBLProperteis with sortcolumns to create the TUPLEID_statusOnMerge carbonwriter, in which the sortcolumn、sortscope are all useless.

 ### What changes were proposed in this PR?
remove the sortcolumn property when creating TUPLEID_statusOnMerge carbonwriter.

 ### Does this PR introduce any user interface change?
No

 ### Is any new testcase added?
Yes

PR #3856, opened 2020-07-21
[CARBONDATA-3929]Improve CDC performance

akashrn5:
### Why is this PR needed?
This PR is to improve the CDC merge performance. CDC is currently very slow in the case of full outer joins and slow in normal cases. Identified pain points are as below:
1. currently we are writing the intermediate delete data to carbon format, which is the columnar format, and we do a full scan which is slow. Here since intermediate, we do full scan, compression, columnar format, its all-time taking.
2. Full outer join case is very slow.
3. when we insert new data into new segments, we follow the old insert flow with the converter step.
4. since we write the intermediate data carbon format, we use coalesce to limit the partition to number of active executors.
 
Some improvements points are identified as below
1. Write the intermediate data to a faster row format like Avro.
2. use bucketing on join column and do the repartition of the Dataframe before performing the join operation, which avoids the shuffle on one side as shuffle is major time consuming part in join.
3. make the insert flow to the new flow without the converter step.
4. remove coalesce and can use resource to write the intermediate Avro data in a faster way.
    
Performance results
| --- |
| DataSize -> 2GB target table data |
| 230MB source table data |
| InnerJoin case - around 17000+ deleted rows, 70400 odd updated rows |
| Full outer Join case - 2million target data, 0.2million src data, 70400 odd rows updated and some deleted |

|   | Old Time(sec) |   | New Time(sec) |   |
| :---: | :---: | :---: | :---: | :---: |
| **Join Type** | **1st time query** | **2nd time query** | **1st time query** | **2nd time query** |
| **Inner Join** | 20 | 9.6 | 14 | 4.6 |
| **Full Outer Join** | 43 | 17.8 | 26 | 7.7 |
 - No

 - Yes

### DIFF
--- a/integration/spark/pom.xml
+++ b/integration/spark/pom.xml
@@ -154,6 +154,28 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.databricks</groupId>
+      <artifactId>spark-avro_${scala.binary.version}</artifactId>
+      <version>4.0.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-avro_${scala.binary.version}</artifactId>
+      <version>2.4.5</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-sql-kafka-0-10_${scala.binary.version}</artifactId>
       <exclusions>
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
@@ -28,7 +28,7 @@ import org.apache.spark.sql.{AnalysisException, CarbonEnv, CarbonToSparkAdapter,
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Literal, NamedExpression}
 import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, Project}
-import org.apache.spark.sql.execution.command.AtomicRunnableCommand
+import org.apache.spark.sql.execution.command.{AtomicRunnableCommand, UpdateTableModel}
 import org.apache.spark.sql.execution.datasources.LogicalRelation
 import org.apache.spark.sql.types.{StructField, StructType}
 import org.apache.spark.util.CausedBy
@@ -64,7 +64,8 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
     var tableInfo: TableInfo,
     var internalOptions: Map[String, String] = Map.empty,
     var partition: Map[String, Option[String]] = Map.empty,
-    var operationContext: OperationContext = new OperationContext)
+    var operationContext: OperationContext = new OperationContext,
+    var updateModel: Option[UpdateTableModel] = None)
   extends AtomicRunnableCommand {
 
   var table: CarbonTable = _
@@ -200,11 +201,20 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
           options = options.asJava,
           isOverwriteTable = isOverwriteTable,
           isDataFrame = true,
-          updateModel = None,
+          updateModel = updateModel,
           operationContext = operationContext)
 
       // add the start entry for the new load in the table status file
-      if (!table.isHivePartitionTable) {
+      if ((updateModel.isEmpty || updateModel.isDefined && updateModel.get.loadAsNewSegment)
+          && !table.isHivePartitionTable) {
+        if (updateModel.isDefined ) {
+          carbonLoadModel.setFactTimeStamp(updateModel.get.updatedTimeStamp)
+        }
+        CarbonLoaderUtil.readAndUpdateLoadProgressInTableMeta(
+          carbonLoadModel,
+          isOverwriteTable)
+        isUpdateTableStatusRequired = true
+      } else if (!table.isHivePartitionTable) {
         CarbonLoaderUtil.readAndUpdateLoadProgressInTableMeta(
           carbonLoadModel,
           isOverwriteTable)
@@ -238,7 +248,7 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
         partitionStatus,
         None,
         Some(scanResultRdd),
-        None,
+        updateModel,
         operationContext)
       LOGGER.info("Sort Scope : " + carbonLoadModel.getSortScope)
       val (rows, loadResult) = insertData(loadParams)
@@ -439,6 +449,13 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
 
   def insertData(loadParams: CarbonLoadParams): (Seq[Row], LoadMetadataDetails) = {
     var rows = Seq.empty[Row]
+    val loadDataFrame = if (updateModel.isDefined && !updateModel.get.loadAsNewSegment) {
+      // TODO: handle the update flow for new insert into flow without converter step
+      throw new UnsupportedOperationException(
+        "Update flow is not supported without no converter step yet.")
+    } else {
+      Some(dataFrame)
+    }
     val table = loadParams.carbonLoadModel.getCarbonDataLoadSchema.getCarbonTable
     var loadResult : LoadMetadataDetails = null
     if (table.isHivePartitionTable) {
@@ -449,9 +466,9 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
         loadParams.partitionStatus,
         isOverwriteTable,
         loadParams.hadoopConf,
-        None,
+        loadDataFrame,
         loadParams.scanResultRDD,
-        None,
+        updateModel,
         operationContext)
     }
     (rows, loadResult)
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
@@ -26,18 +26,17 @@ import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapreduce.{Job, JobID, TaskAttemptID, TaskID, TaskType}
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
 import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
-import org.apache.spark.sql.{AnalysisException, CarbonDatasourceHadoopRelation, CarbonUtils, Column, DataFrame, Dataset, Row, SparkSession}
-import org.apache.spark.sql.carbondata.execution.datasources.SparkCarbonFileFormat
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{AnalysisException, CarbonUtils, Column, DataFrame, Dataset, Row, SparkSession}
+import org.apache.spark.sql.avro.AvroFileFormatFactory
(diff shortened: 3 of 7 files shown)


CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3451/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1709/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1712/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3454/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3506/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1764/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1789/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1963/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3702/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3793/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2051/

integration/spark/pom.xml
```diff
@@ -153,6 +153,28 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-avro_${scala.binary.version}</artifactId>
```
ajantha-bhat, line 169: why can't spark2.3 and 2.4 both use databricks spark-avro ? I can understand that other way around is not possible (for both to use  spark avro)

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
```diff
@@ -439,6 +449,11 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
 
   def insertData(loadParams: CarbonLoadParams): (Seq[Row], LoadMetadataDetails) = {
     var rows = Seq.empty[Row]
+    val loadDataFrame = if (updateModel.isDefined && !updateModel.get.loadAsNewSegment) {
+      Some(CommonLoadUtils.getDataFrameWithTupleID(Some(dataFrame)))
```
ajantha-bhat, line 453: This InsertIntoCommand flow is not meant for update flow yet. Because update will have an implicit column and rearrange schema and all will fail. so, I suggest if `updateModel.get.loadAsNewSegment` is `false` throw unsupported exception now and handle this requirement later.

Also when `updateModel.get.loadAsNewSegment = true` (which is our current cdc history data case), **this flow can be used** (as it is just a insert, no actual update flow used). only when `updateModel.get.loadAsNewSegment = false` we cannot use this flow. 

so someone might use it because of update model support. so, I suggest to throw an exception in the beginning of this function when  `updateModel.get.loadAsNewSegment = false`

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
```diff
@@ -194,29 +210,32 @@ case class CarbonMergeDataSetCommand(
...
       None
     }
 
-    CarbonInsertIntoWithDf(
-      databaseNameOp = Some(carbonTable.getDatabaseName),
+    val dataFrame = loadDF.select(tableCols.map(col): _*)
+    CarbonInsertIntoCommand(databaseNameOp = Some(carbonTable.getDatabaseName),
       tableName = carbonTable.getTableName,
-      options = Map("fileheader" -> header, "sort_scope" -> "nosort"),
+      options = Map("fileheader" -> header, "sort_scope" -> "no_sort"),
```
ajantha-bhat, line 222: I know it is base behvior, is it better to use target table sort scope ?

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
ajantha-bhat: Also, I feel no need of updateModel only if it is just insert flow. because updateModel.get.loadAsNewSegment = true, is just insert flow. please also get it confirmed with @ravipesala

integration/spark/src/main/spark2.3/org/apache/spark/sql/avro/AvroFileFormatFactory.scala
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.apache.spark.sql.avro
+
+import com.databricks.spark.avro.{AvroReader, AvroWriter}
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.execution.datasources.OutputWriterFactory
+
+object AvroFileFormatFactory {
```
ajantha-bhat, line 24: same as above doubt, may be just use databricks spark acro for both 2.3 and 2.4

integration/spark/src/main/spark2.3/com/databricks/spark/avro/AvroWriter.scala
```diff
@@ -0,0 +1,51 @@
...
+      dataSchema: org.apache.spark.sql.types.StructType,
+      options: scala.Predef.Map[scala.Predef.String, scala.Predef.String] = Map.empty)
+  : OutputWriterFactory = {
+    new DefaultSource().prepareWrite(spark, job,
+      options, dataSchema)
+  }
+}
+
+/**
+ * This reds the avro files from the given path and return the RDD[Row]
```
ajantha-bhat, line 41:
```suggestion
 * This reads the avro files from the given path and return the RDD[Row]
```

integration/spark/pom.xml
akashrn5: since we are integrated with spark, spark-avro is preferable, but as we know spark-2.3 does not have spark-avro, so anyway once in future we remove support for spark-2.3, so databricks spark-avro will be removed for our project. So here i'm avoiding out of spark project dependency as its possible in spark-2.4

integration/spark/src/main/spark2.3/org/apache/spark/sql/avro/AvroFileFormatFactory.scala
akashrn5: replied above

integration/spark/src/main/spark2.3/com/databricks/spark/avro/AvroWriter.scala
akashrn5: done

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
akashrn5: done

akashrn5: > This InsertIntoCommand flow is not meant for update flow yet. Because update will have an implicit column and rearrange schema and all will fail. so, I suggest if `updateModel.get.loadAsNewSegment` is `false` throw unsupported exception now and handle this requirement later.
> 
> Also when `updateModel.get.loadAsNewSegment = true` (which is our current cdc history data case), **this flow can be used** (as it is just a insert, no actual update flow used). only when `updateModel.get.loadAsNewSegment = false` we cannot use this flow.
> 
> so someone might use it because of update model support. so, I suggest to throw an exception in the beginning of this function when `updateModel.get.loadAsNewSegment = false`

done

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
```diff
@@ -106,18 +106,34 @@ case class CarbonMergeDataSetCommand(
     // decide join type based on match conditions
     val joinType = decideJoinType
 
+    val joinColumn = mergeMatches.joinExpr.expr.asInstanceOf[EqualTo].left
+      .asInstanceOf[UnresolvedAttribute].nameParts.tail.head
+    // repartition the the srsDs, if the target as bucketing and the bucketing column and join
+    // column are same
+    val repartitionedSrsDs =
```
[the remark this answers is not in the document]
akashrn5: done

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
```diff
@@ -106,18 +106,34 @@ case class CarbonMergeDataSetCommand(
     // decide join type based on match conditions
     val joinType = decideJoinType
 
+    val joinColumn = mergeMatches.joinExpr.expr.asInstanceOf[EqualTo].left
+      .asInstanceOf[UnresolvedAttribute].nameParts.tail.head
+    // repartition the the srsDs, if the target as bucketing and the bucketing column and join
```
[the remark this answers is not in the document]
akashrn5: done

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
akashrn5: if we focus on the cdc performance, existing no sort is fine. As your point, we cant blindly go with target table's sort scope, because it can happen that the target table has the sort scope has no sort.

ravipesala: I agree with @ajantha-bhat  we should use the target table sort scope. But this PR is only about performance we can cover this change in another PR with proper test

ravipesala:
LGTM

QiangCai:
why don't use row_v1 format in the streaming module?

akashrn5:
> why don't use row_v1 format in the streaming module?
already avro is row format and same write interface it was using. I took the same with less code changes in merge command base code and as its more stable, later if we find better performance or other advantages we can try

akashrn5:
retest this please

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3866/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2125/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/merge/CarbonMergeDataSetCommand.scala
ajantha-bhat: @akashrn5 : instead of fixing no_sort, would have analyzed why it was added. Originally it was using Target table sort scope only. #3764 added this. would have removed it here.
Now that you changed to no_sort, target table sort_scope is not used.

ajantha-bhat: check this.
https://github.com/apache/carbondata/pull/3901

PR #3880, opened 2020-08-06
[CARBONDATA-3879] Filtering Segmets Optimazation

marchpure:
### Why is this PR needed?
During filter segments flow, there are a lot of LIST.CONTAINS, which has heavy time overhead when there are tens of thousands segments.

For example, if there are 50000 segments. it will trigger LIST.CONTAINS  for each segment, the LIST also has about 50000 elements. so the time complexity will be O（50000 * 50000 ）
 
Change List.CONTAINS to MAP.containsKEY
    
 - No

 - No

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
@@ -389,14 +389,16 @@ public class CarbonTableInputFormat<T> extends CarbonInputFormat<T> {
 
   public void updateLoadMetaDataDetailsToSegments(List<Segment> validSegments,
       List<org.apache.carbondata.hadoop.CarbonInputSplit> prunedSplits) {
+    Map<String, Segment> validSegmentsMap = validSegments.stream()
+        .collect(Collectors.toMap(Segment::getSegmentNo, segment -> segment, (e1, e2) -> e1));
     for (CarbonInputSplit split : prunedSplits) {
       Segment segment = split.getSegment();
       if (segment.getLoadMetadataDetails() == null || segment.getReadCommittedScope() == null) {
-        if (validSegments.contains(segment)) {
+        if (validSegmentsMap.containsKey(segment.getSegmentNo())) {
           segment.setLoadMetadataDetails(
-              validSegments.get(validSegments.indexOf(segment)).getLoadMetadataDetails());
+              validSegmentsMap.get(segment.getSegmentNo()).getLoadMetadataDetails());
           segment.setReadCommittedScope(
-              validSegments.get(validSegments.indexOf(segment)).getReadCommittedScope());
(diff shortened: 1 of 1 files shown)


CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/1884/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3623/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -389,14 +389,16 @@ protected FileSplit makeSplit(String segmentId, String filePath, long start, lon
 
   public void updateLoadMetaDataDetailsToSegments(List<Segment> validSegments,
       List<org.apache.carbondata.hadoop.CarbonInputSplit> prunedSplits) {
+    Map<String, Segment> validSegmentsMap = validSegments.stream()
```
ajantha-bhat, line 392: creating a map everytime for query is also overhead when the valid segments is in thousands, I suggest we can use `SET` for `validSegments` when it is formed originally instead of 'LIST'

marchpure: Appreciate for that good suggestion. But here we can only use Map instead of Set. 

Reason: The The pseudo-code for this function is shown as below.

        // **0. segments.hashcode is segmentno, to when we compare 2 segments, only segmentno will be compared**
        if (validSegments.contains(segmentInSplit)) {
		  **1. fetch the segment from validSegments** 
		  Segment segmentInValidSegment <- fetch the segment from validSegments
		  2. 
          segmentInSplit.setLoadMetadataDetails(segmentInValidSegment.getLoadMetadataDetails);
        }

For SET. it's hard to fetch the segment from validSegments.

ajantha-bhat: if you see the `equals` implementation of `Segment.java`, it is based on only segment number comparison. so, I think you can still use `SET`

marchpure: Yeah. But if use SET. it's hard to read a element with specified segmentno. 
In this code. we need to read a segment with specified segmentno from validSegments.

ajantha-bhat: oh, I got what you mean, we still need to get the element from valid segment to read its LoadMetadataDetails. So, SET can help only for contains, but not for getting the valid segment.

ajantha-bhat:
LGTM

PR #3899, opened 2020-08-25
[HOTFIX] Aviod cleaning segments after reading tablestatus failed

QiangCai:
### Why is this PR needed?
 1. After reading tablestatus file failed,  method TableProcessingOperations. deletePartialLoadDataIfExist will delete all related segments.

 2. If the tablestatus file was removed, the system will delete all original segment before loading. 

1. check the result of reading tablestatus file, if it is empty, no need to process.
2. if the table status was removed, the system shouldn't delete any segment.
3. re-factory code to avoid invoking getAbsolutePath method too many times.

    
 - No

 - No

### DIFF
--- a/integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
@@ -19,25 +19,25 @@ package org.apache.carbondata.integration.spark.testsuite.dataload
 
 import java.math.BigDecimal
 
+import org.apache.commons.io.FileUtils
+import org.apache.commons.lang3.RandomStringUtils
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.test.util.QueryTest
 import org.scalatest.BeforeAndAfterEach
 
-import org.apache.carbondata.core.util.path.CarbonTablePath
-import org.apache.carbondata.core.datastore.impl.FileFactory
-import org.apache.carbondata.core.metadata.CarbonMetadata
 import org.apache.carbondata.core.constants.{CarbonCommonConstants, CarbonLoadOptionConstants}
+import org.apache.carbondata.core.datastore.impl.FileFactory
 import org.apache.carbondata.core.index.Segment
+import org.apache.carbondata.core.metadata.CarbonMetadata
 import org.apache.carbondata.core.util.CarbonProperties
+import org.apache.carbondata.core.util.path.CarbonTablePath
 import org.apache.carbondata.spark.util.BadRecordUtil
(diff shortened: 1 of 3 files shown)


CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2122/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3863/

QiangCai:
@kunal642 please check this modification.

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3872/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2131/

Zhangshunyu:
LGTM

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
```diff
@@ -310,7 +310,8 @@ class TestLoadDataGeneral extends QueryTest with BeforeAndAfterEach {
     val tableStatusFile = CarbonTablePath.getTableStatusFilePath(carbonTable.getTablePath)
     FileFactory.getCarbonFile(tableStatusFile).delete()
     sql("insert into stale values('k')")
-    checkAnswer(sql("select * from stale"), Row("k"))
+    // if table lose tablestatus file, the system should keep all data.
+    checkAnswer(sql("select * from stale"), Seq(Row("k"), Row("k")))
```
kunal642, line 314: I think this behaviour would be harrmful when the segment is marked for delete..If the segment is not removed then the results would be incorrect.

QiangCai: yes
1. now, after removing the tablestatus file,  we can consider all segments are successful by default.
2. in the future, we can add a flag file into the segment when we mark the segment to be deleted.
   the flag file contains a message the segment is mark_for_delete.

QiangCai: for this test case, the result is two rows,  the behavior is right.

kunal642: ok

kunal642:
LGTM

PR #3908, opened 2020-08-31
[CARBONDATA-3967] cache partition on select to enable faster pruning

kunal642:
### Why is this PR needed?
 spark's getPartition method can sometime take a lot of time which can degrade the performance for the queries.
 
Cache the partition on each load and query so that the subsequent queries can perform faster. Caching is done under the expiring map interface so the cache expiration time can work for this also.
    
 - No

 - No
 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -1274,6 +1274,13 @@ public final class CarbonCommonConstants {
   public static final String CARBON_MAX_DRIVER_LRU_CACHE_SIZE = "carbon.max.driver.lru.cache.size";
 
   /**
+   * max driver lru cache size upto which partition lru cache will be loaded in memory
+   */
+  @CarbonProperty
+  public static final String CARBON_PARTITION_MAX_DRIVER_LRU_CACHE_SIZE =
+      "carbon.partition.max.driver.lru.cache.size";
+
+  /**
    * max executor lru cache size upto which lru cache will be loaded in memory
    */
   @CarbonProperty
--- a/integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
@@ -1,0 +1,181 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.util
+
+import java.net.URI
+import java.util
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+import org.apache.log4j.Logger
+import org.apache.spark.sql.catalyst.catalog.{CatalogStorageFormat, CatalogTablePartition}
+
+import org.apache.carbondata.common.logging.LogServiceFactory
+import org.apache.carbondata.core.cache.{Cache, Cacheable, CarbonLRUCache}
+import org.apache.carbondata.core.constants.CarbonCommonConstants
+import org.apache.carbondata.core.datastore.impl.FileFactory
+import org.apache.carbondata.core.index.Segment
+import org.apache.carbondata.core.metadata.{AbsoluteTableIdentifier, SegmentFileStore}
+import org.apache.carbondata.core.statusmanager.{LoadMetadataDetails, SegmentStatusManager}
+import org.apache.carbondata.core.util.path.CarbonTablePath
+
+object PartitionCacheManager extends Cache[PartitionCacheKey,
+  java.util.List[CatalogTablePartition]] {
+
+  private val CACHE = new CarbonLRUCache(
+    CarbonCommonConstants.CARBON_PARTITION_MAX_DRIVER_LRU_CACHE_SIZE,
+    CarbonCommonConstants.CARBON_MAX_LRU_CACHE_SIZE_DEFAULT)
+
+  val LOGGER: Logger = LogServiceFactory.getLogService(this.getClass.getName)
+
+  def get(identifier: PartitionCacheKey): java.util.List[CatalogTablePartition] = {
+    LOGGER.info("Reading partition values from store")
+    // read the tableStatus file to get valid and invalid segments
+    val validInvalidSegments = new SegmentStatusManager(AbsoluteTableIdentifier.from(
+      identifier.tablePath, null, null, identifier.tableId))
+      .getValidAndInvalidSegments
+    val existingCache = CACHE.get(identifier.tableId)
+    val cacheablePartitionSpecs = validInvalidSegments.getValidSegments.asScala.map { segment =>
+      val segmentFileName = segment.getSegmentFileName
+      val segmentFilePath = FileFactory.getCarbonFile(
+        CarbonTablePath.getSegmentFilePath(identifier.tablePath, segmentFileName))
+      // read the last modified time
+      val segmentFileModifiedTime = segmentFilePath.getLastModifiedTime
+      if (existingCache != null) {
+        val segmentCache = existingCache.asInstanceOf[CacheablePartitionSpec]
+          .partitionSpecs.get(segment.getSegmentNo)
+        segmentCache match {
+          case Some(c) =>
+            // check if cache needs to be updated
+            if (segmentFileModifiedTime > c._2) {
+              (segment.getSegmentNo, (readPartition(identifier,
+                segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+            } else {
+              (segment.getSegmentNo, c)
+            }
+          case None =>
+            (segment.getSegmentNo, (readPartition(identifier,
+              segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+        }
+      } else {
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
+          segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+      }
+    }.toMap
+    // remove all invalid segment entries from cache
+    val finalCache = cacheablePartitionSpecs --
+                     validInvalidSegments.getInvalidSegments.asScala.map(_.getSegmentNo)
+    val cacheObject = CacheablePartitionSpec(finalCache)
+    if (finalCache.nonEmpty) {
+      // remove the existing cache as new cache values may be added.
+      // CarbonLRUCache does not allow cache updation until time is expired.
+      // TODO: Need to fix!!
+      CACHE.remove(identifier.tableId)
+      CACHE.put(identifier.tableId,
+        cacheObject,
+        cacheObject.getMemorySize,
+        identifier.expirationTime)
+    }
+    finalCache.values.flatMap(_._1).toList.asJava
+  }
+
+  override def getAll(keys: util.List[PartitionCacheKey]):
+  util.List[util.List[CatalogTablePartition]] = {
+    keys.asScala.toList.map(get).asJava
+  }
+
+  override def getIfPresent(key: PartitionCacheKey): java.util.List[CatalogTablePartition] = {
+    CACHE.get(key.tableId).asInstanceOf[CacheablePartitionSpec].partitionSpecs.values.flatMap(_._1)
+      .toList.asJava
+  }
+
+  override def invalidate(partitionCacheKey: PartitionCacheKey): Unit = {
+    CACHE.remove(partitionCacheKey.tableId)
+  }
+
+  private def readPartition(identifier: PartitionCacheKey, segmentFilePath: String) = {
+    val segmentFile = SegmentFileStore.readSegmentFile(segmentFilePath)
+    segmentFile.getLocationMap.values().asScala
+      .flatMap(_.getPartitions.asScala).toSet.map { uniquePartition: String =>
+      val partitionSplit = uniquePartition.split("=")
+      val storageFormat = CatalogStorageFormat(
+        Some(new URI(identifier.tablePath + "/" + uniquePartition)),
+        None, None, None, compressed = false, Map())
+      CatalogTablePartition(Map(partitionSplit(0) -> partitionSplit(1)), storageFormat)
+    }.toSeq
+  }
+
+  override def put(key: PartitionCacheKey,
+      value: java.util.List[CatalogTablePartition]): Unit = {
+
+  }
+
+  override def clearAccessCount(keys: util.List[PartitionCacheKey]): Unit = {
+
+  }
+}
+
+case class PartitionCacheKey(tableId: String, tablePath: String, expirationTime: Long)
+
+/**
+ * Cacheable instance of the CatalogTablePartitions.
+ *
+ * Maintains a mapping of segmentNo -> (Seq[CatalogTablePartition], lastModifiedTime)
+ */
+case class CacheablePartitionSpec(partitionSpecs: Map[String, (Seq[CatalogTablePartition], Long)])
+  extends Cacheable {
+
+  /**
+   * This method will return the access count for a column based on which decision will be taken
+   * whether to keep the object in memory
+   *
+   * @return
+   */
+  override def getAccessCount: Int = {
+    0
+  }
+
+  /**
+   * This method will return the memory size of the cached partition
+   *
+   * @return
+   */
+  override def getMemorySize: Long = {
+    partitionSpecs.values.flatMap {
+      partitionSpec =>
+        partitionSpec._1.map { specs =>
+          if (specs.stats.isDefined) {
+            specs.stats.get.sizeInBytes.toLong
+          } else {
+            SizeEstimator.estimate(specs)
+          }
+        }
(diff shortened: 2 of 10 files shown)


CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2193/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/3933/

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  override def getIfPresent(key: PartitionCacheKey): CacheablePartitionSpec = {
+    CACHE.get(key.tableId).asInstanceOf[CacheablePartitionSpec]
+  }
+
+  override def invalidate(partitionCacheKey: PartitionCacheKey): Unit = {
+    CACHE.remove(partitionCacheKey.tableId)
+  }
+
+  private def readPartitions(identifier: PartitionCacheKey, tableStatusModifiedTime: Long) = {
```
QiangCai, line 74: now it will reload all .segment files again. can we support incremental loading?

kunal642: Do you mean caching of partitions after load?

QiangCai: "readPartitions" will read all .segment files after loading. better to load the new .segment file only.

kunal642: ok got it

QiangCai: please check query flow. it also use incremental loading index.
here maybe can reuse the segment info cache.

kunal642: added per segment modification check...now only the updated/new segments would be loaded

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4002/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2262/

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
kunal642: @QiangCai each load or query would be loading the already success segments, so it will now solve the problem you mentioned

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  override def clearAccessCount(keys: util.List[PartitionCacheKey]): Unit = {
+
+  }
+}
+
+case class PartitionCacheKey(tableId: String, tablePath: String, expirationTime: Long)
+
+case class CacheablePartitionSpec(partitionSpecs: Map[String, (Seq[CatalogTablePartition], Long)])
```
QiangCai, line 137: please add a comment about partitionSpecs

kunal642: added

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2290/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4029/

kunal642:
@QiangCai build passed

QiangCai:
please check the comment

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
```diff
@@ -0,0 +1,181 @@
...
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
+          segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+      }
+    }.toMap
+    // remove all invalid segment entries from cache
+    val finalCache = cacheablePartitionSpecs --
+                     validInvalidSegments.getInvalidSegments.asScala.map(_.getSegmentNo)
+    val cacheObject = CacheablePartitionSpec(finalCache)
+    if (finalCache.nonEmpty) {
```
QiangCai, line 86: if finalCache empty, should remove cache if exists

core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -1273,6 +1273,13 @@ private CarbonCommonConstants() {
   @CarbonProperty
   public static final String CARBON_MAX_DRIVER_LRU_CACHE_SIZE = "carbon.max.driver.lru.cache.size";
 
+  /**
+   * max driver lru cache size upto which partition lru cache will be loaded in memory
+   */
+  @CarbonProperty
```
ajantha-bhat, line 1279: please also need to update the document

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
```diff
@@ -0,0 +1,181 @@
...
+            } else {
+              (segment.getSegmentNo, c)
+            }
+          case None =>
+            (segment.getSegmentNo, (readPartition(identifier,
+              segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+        }
+      } else {
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
```
ajantha-bhat, line 78: readPartition reads the segment file. 
so basically first time we read all the segment files of the table.

I have seen the users with 200K segments in one table stored in cloud. query takes more than 5 hours for them if we do this.

So, need to change the key as partition spec and cache only matched partitions

@kumarvishal09 , @QiangCai @marchpure : what is your opinion on this ?

marchpure: agree with you. 
also avoid to use getAbsolutePath function，which has high IO overhead

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
```diff
@@ -0,0 +1,181 @@
...
+      // read the last modified time
+      val segmentFileModifiedTime = segmentFilePath.getLastModifiedTime
+      val existingCache = CACHE.get(identifier.tableId)
+      if (existingCache != null) {
+        val segmentCache = CACHE.get(identifier.tableId).asInstanceOf[CacheablePartitionSpec]
+          .partitionSpecs.get(segment.getSegmentNo)
+        segmentCache match {
+          case Some(c) =>
+            // check if cache needs to be updated
+            if (segmentFileModifiedTime > c._2) {
```
akashrn5, line 66: As i can see from here, we are basically removing the cache if the segment file is updated. Actually same thing we do in `TableSegmentRefresher`, and use that for updating all cache, can move this piece of logic there also to unify for all? we use `TableSegmentRefresher` during query, but here we are doing in case of load, can you try if we can do that? 

If not this PR, may be you can give a try in another, as it will be unified and good.

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
akashrn5: As you mentioned first time it will take lot of time, but query time we will get benefit. You said matched partitions and since here for every load we are trying to cache, we wont be able to get any matched partitions right?

It is something similar to our Indexes(MV, SI) etc where we load all segment at once in first time and from next time we do incremental, as @QiangCai mentioned. So here from subsequent loads, only that segment should be considered, which i think its being taken care.

@ajantha-bhat are you trying to say to change to handle in query or load itself? Correct me, if i'm wrong.

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/standardpartition/StandardPartitionTableLoadingTestCase.scala
```diff
@@ -585,6 +587,30 @@ class StandardPartitionTableLoadingTestCase extends QueryTest with BeforeAndAfte
     }
   }
 
+  test("test partition caching") {
+    CarbonProperties.getInstance().addProperty("carbon.read.partition.hive.direct", "false")
+    sql("drop table if exists partition_cache")
+    sql("create table partition_cache(a string) partitioned by(b int) stored as carbondata")
+    sql("insert into partition_cache select 'k',1")
+    sql("select * from partition_cache where b = 1").collect()
```
akashrn5, line 595: can we add one assert to check the cache, after the load instead of query in one scenario?

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
akashrn5: and i also checked now, this caching happens in flow query too, so you can consider my reply specific to load.

ajantha-bhat: @akashrn5 : yeah, my problem scenario is first time query or upgrade, taking huge time (near to 5 hours for first time query) is still not ok.  In this scenario, I feel reading from hive metastore itself is very fast compared to number of IO from the cloud storage

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,176 @@
...
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
+          segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+      }
+    }.toMap
+    // remove all invalid segment entries from cache
+    val finalCache = cacheablePartitionSpecs --
+                     validInvalidSegments.getInvalidSegments.asScala.map(_.getSegmentNo)
+    val cacheObject = CacheablePartitionSpec(finalCache)
+    if (finalCache.nonEmpty) {
```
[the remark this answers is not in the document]
kunal642: Can be empty during first load. in that case empty partitions would be stored in the cache

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
kunal642: this check is to avoid caching empty cache map

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
kunal642: @ajantha-bhat This is disabled by default, so if the user feels that hive metastore getPartitions is slow, can enable this to improve performance.

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
kunal642: TableSegmentRefresher is in core module, cannot add there

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,176 @@
...
+        }
+      } else {
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
+          segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+      }
+    }.toMap
+    // remove all invalid segment entries from cache
+    val finalCache = cacheablePartitionSpecs --
+                     validInvalidSegments.getInvalidSegments.asScala.map(_.getSegmentNo)
```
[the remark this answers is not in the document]
kunal642: this is to remove entry of any invalid segment(like deleted segment) from the cacheablePartitionSpecs.

As cacheablePartitionSpecs is a map of (segmentNo -> PartitionSpecs), therefore i am removing all the invalid segments from the final cache.

Does this address your comment?

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,176 @@
...
+    val validInvalidSegments = new SegmentStatusManager(AbsoluteTableIdentifier.from(
+      identifier.tablePath, null, null, identifier.tableId))
+      .getValidAndInvalidSegments
+    val cacheablePartitionSpecs = validInvalidSegments.getValidSegments.asScala.map { segment =>
+      val segmentFileName = segment.getSegmentFileName
+      val segmentFilePath = FileFactory.getCarbonFile(
+        CarbonTablePath.getSegmentFilePath(identifier.tablePath, segmentFileName))
+      // read the last modified time
+      val segmentFileModifiedTime = segmentFilePath.getLastModifiedTime
+      val existingCache = CACHE.get(identifier.tableId)
```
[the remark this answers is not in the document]
kunal642: done

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManger.scala
```diff
@@ -0,0 +1,176 @@
...
+      if (existingCache != null) {
+        val segmentCache = CACHE.get(identifier.tableId).asInstanceOf[CacheablePartitionSpec]
```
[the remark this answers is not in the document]
kunal642: done

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
ajantha-bhat: ok. If the user feels hive metastore getPartitions is slow, still he cannot enable this in cloud scenario if he had the huge segments as it will be slower than hive metastore. 

As the feature is disabled by default. Ok for me.

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
akashrn5: yes, you are right, i didn't observe that module level.

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
kunal642: @ajantha-bhat

kunal642: @ajantha-bhat 

`sparkSession.sessionState.catalog.listPartitionsByFilter(identifier, partitionFilters)` is already being used for to filter the partitions based on the filter, but in concurrent queries as apark has to take a lock on metastore before accessing it, therefore the query performance was being degraded.

And no user will have 200k segment, they need to anyways compact otherwise the query performance would be degraded

ajantha-bhat: we have users with 200k segments. But the data is partitioned by date and time. so all queries will have partition columns filter. so they query-specific segments every time (without any degrade in query time)

kunal642: Yeah, because they are using presto, where compaction is not yet supported.

If they are using hivemetastore surely they will face degrade in concurrent queries

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2311/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4049/

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/standardpartition/StandardPartitionTableLoadingTestCase.scala
kunal642: done

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
```diff
@@ -0,0 +1,181 @@
...
+              segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+        }
+      } else {
+        // read the partitions if not available in cache.
+        (segment.getSegmentNo, (readPartition(identifier,
+          segmentFilePath.getAbsolutePath), segmentFileModifiedTime))
+      }
+    }.toMap
+    // remove all invalid segment entries from cache
+    val finalCache = cacheablePartitionSpecs --
```
QiangCai, line 83: substraction of different type

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
QiangCai: I mean it should has "else" branch to remove cache

integration/spark/src/main/scala/org/apache/spark/util/PartitionCacheManager.scala
```diff
@@ -0,0 +1,181 @@
...
+                     validInvalidSegments.getInvalidSegments.asScala.map(_.getSegmentNo)
+    val cacheObject = CacheablePartitionSpec(finalCache)
+    if (finalCache.nonEmpty) {
+      // remove the existing cache as new cache values may be added.
+      // CarbonLRUCache does not allow cache updation until time is expired.
+      // TODO: Need to fix!!
+      CACHE.remove(identifier.tableId)
```
QiangCai, line 90: for concurrent query,  line 53 will return null and it will lead to read all partitions.

QiangCai:
@kunal642 please fix the comments in another PR.

LGTM

ajantha-bhat:
Agree with @QiangCai. Merging the PR now

PR #3921, opened 2020-09-11
[CARBONDATA-3928] Removed records from exception message.

nihal0107:
### Why is this PR needed?
Currently, when the string length exceeds 32000 and bad record action as fail then
records are thrown in exception message which violates security concern.
 
 Removed the records from the exception message and only printed in logger file and redirected csv file.
    
 - No

 - Yes

### DIFF
--- a/processing/src/main/java/org/apache/carbondata/processing/loading/converter/impl/RowConverterImpl.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/loading/converter/impl/RowConverterImpl.java
@@ -18,7 +18,6 @@
 package org.apache.carbondata.processing.loading.converter.impl;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4046/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2308/

processing/src/main/java/org/apache/carbondata/processing/loading/converter/impl/RowConverterImpl.java
```diff
@@ -119,10 +118,6 @@ public CarbonRow convert(CarbonRow row) throws CarbonDataLoadingException {
             .getTableProperties();
     String spatialProperty = properties.get(CarbonCommonConstants.SPATIAL_INDEX);
     boolean isSpatialColumn = false;
-    Object[] rawData = row.getRawData();
-    if (rawData == null) {
-      rawData = row.getData() == null ? null : row.getData().clone();
```
ajantha-bhat, line 124: Please revert back this change, somecases rawData will not be set, hence they are setting here before converting the row. 
Now your current code will use null instead of row.getData().clone() in this scenario.

nihal0107: I had only added this scenario last time because at that time we wanted rawdata for every bad record action. But now we need this only in case of bad record logger is enable or action is redirect. And in this case, rawdata will be always available.

ajantha-bhat: Ok then.

ajantha-bhat:
LGTM

PR #3924, opened 2020-09-14
[CARBONDATA-3988] Allow SI creation on first dimension column

ajantha-bhat:
### Why is this PR needed?
 In some scenarios like, A table having 100 columns. with sort columns as (A,B,C). Currently we don't allow SI creation on column 'A'.  Due to this If the filter is B=x, A=y.  There will be huge task to scan main table as SI is not supported for column 'A'. Instead if we allow SI for A,B. There will be less task to scan main table and over all query execution will be faster.
**Let user decide he need to create SI for first column or not, instead of we blocking always.**
 
Allow first dimension column also as SI column. 

 - No

 - No

### DIFF
--- a/index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestCreateIndexTable.scala
+++ b/index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestCreateIndexTable.scala
@@ -126,21 +126,6 @@ class TestCreateIndexTable extends QueryTest with BeforeAndAfterAll {
 
   }
 
-  test("test create index table cols order same as start order of parent table") {
-    try {
-      sql("drop index if exists index_on_measure on carbon")
-      sql("create index index_on_measure on table carbon (empno,empname,designation,doj,"+
-          "workgroupcategory) AS 'carbondata'")
-      assert(false)
-    } catch {
-      case ex: Exception =>
-        assert(ex.getMessage.equalsIgnoreCase(
-          "Index table column indexing order is same as Parent table column start order"))
-    } finally{
-      sql("drop index if exists index_on_measure on carbon")
-     }
-  }
-
   test("test create index table on more than one column") {
     try {
       sql("drop index if exists index_more_columns on carbon")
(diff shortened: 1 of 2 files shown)


CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4058/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2320/

ajantha-bhat:
@jack86596 : You can give me more scenario example to add if you have for this.

ajantha-bhat:
@QiangCai : please check and merge this

QiangCai:
I don't understand the advantage of this change.
maybe we need to improve the use logic of both SI and the main index.

ajantha-bhat:
retest this please

jack86596:
The main idea of this PR is to improve the performance of following query: select * from table where col1 = 'xxx' or col2 = 'yyy'; where sort_columns order is col1,...,col2. Currently we can only create SI on col2, but if only 1 column SI is created another is not, above query will still just scan the main table directly, will not use col2 SI, leads to bad performance.
So we have two options to improve this kind of query:
1. create SI on col1 also, this is the solution of this PR
2. if only one SI is created, still use this SI, other columns use main index to prune, this will be done later.

And for no_sort table, first column is not sorted either, main index is worse than SI, under this scenario, we need this PR to create SI on the first column.

ajantha-bhat:
@jack86596 : Thanks for adding the user problem.
@QiangCai : The scenario is mentioned above, if we don't create the SI on first column, the task launched for scanning main table is more (as the data files in main table is more), so if we support SI for first column and other column the task launched is very less. which will improve concurrent performance.

Also I agree that if user creates SI with just first column, may be SI is no use. But when multiple column SI creation, This can be useful.

Please check and merge.

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2619/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4369/

QiangCai:
LGTM

PR #3929, opened 2020-09-16
[CARBONDATA-3991] Fix the set modified time function on S3 and Alluxio file system

Klaus-xjp:
##Why is this PR needed?
If there are some update or create mv in S3 and Alluxio files ,those operations do not take effect. And the other tenant can't find the changes in this operation and cause a data consistency problem.

##What changes were proposed in this PR?
In this PR, we set two function to fix this problem.
1. Create a new file for S3 and Alluxio to set a tag to update.
2. Set a switch case like function to select different scenario in update or create.

##Does this PR introduce any user interface change?
No

##Is any new testcase added?
Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/filesystem/AbstractDFSCarbonFile.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/filesystem/AbstractDFSCarbonFile.java
@@ -420,6 +420,20 @@ public abstract class AbstractDFSCarbonFile implements CarbonFile {
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


CarbonDataQA1:

QiangCai:
if the old file exists, how does it work?

QiangCai:
retest this please

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4090/

PR #3933, opened 2020-09-16
Remove deletePartialLoadData command in loading and compaction process

Pickupolddriver:
### Why is this PR needed?
Currently, in data management scenarios(Data Loading, Segments Compaction.etc), there exist some data deletion actions. And these actions are dangerous because they are written in different places and some corner cases will cause data deletion accidentally. 
 
This PR remove deleting INSERT_IN_PROGRESS and INSERT_OVERWRITE_IN_PROGRESS status segments during loading and compaction flow.
    
 - No

 - No


CarbonDataQA1:

PR #3934, opened 2020-09-16
[WIP] Support Global Unique Id for SegmentNo

marchpure:
### Why is this PR needed?
 
 
 ### What changes were proposed in this PR?

    
 ### Does this PR introduce any user interface change?
 - No
 - Yes. (please explain the change and update document)

 ### Is any new testcase added?
 - No
 - Yes

PR #3917, opened 2020-09-09
[CARBONDATA-3978] Clean Files Refactor and support for trash folder in carbondata

vikramahuja1001:
### Why is this PR needed?
1. To add functionality of a temp trash folder, where all the carbondata and index files all the stale segments(segments whose entry is not in tablestatus file but exists in Fact folder and metadata folder) will go after clean files is called.
2. There is no way to see what all data will be deleted and moved to trash once clean files is called. 

 
 ### What changes were proposed in this PR?
1. Added trash folder(recycle bin) where all the data files of stale segments(segments whose entry is not in tablestatus file but exists in Fact folder and metadata folder) go to after clean files operation. Timestamp based subdirectories are added in the trash folder and those subdirectories will contain the segment folders which are moved to the trash. By default the expiration day of the timestamp subdirectory is 3 days, the subsequent clean files command after the expiration will remove that timestamp subdirectory. The property exposed for this purpose is "carbon.trash.expiration.time".
2. Clean files dry run - The dry run logic will show all the segments that are going to be manipulated in the clean files command. 
2. After cleaning logic removing stale files from metadata folder. Whatever segment files which are present in metadata folder, whose entry is not in tablestatus file are deleted as they are stale segments.
2. Refactored clean files code - Created a new clean files util which contains all the cleanfiles related code.
    
 ### Does this PR introduce any user interface change?
 - Yes --> Added cleanfiles.md

 ### Is any new testcase added?
 - Yes

PR #3935, opened 2020-09-16
[CARBONDATA-3993] Remove auto data deletion in IUD processs

Pickupolddriver:
### Why is this PR needed?
Currently, in data management scenarios(Data Loading, Segments Compaction.etc), there exist some data deletion actions. And these actions are dangerous because they are written in different places and some corner cases will cause data deletion accidentally.
 
This PR remove auto data deletion during IUD( Insert, Update, Delete) process;
    
 - No

 - No

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
@@ -29,9 +29,6 @@ import org.apache.carbondata.core.index.IndexStoreManager;
 import org.apache.carbondata.core.index.Segment;
 import org.apache.carbondata.core.index.TableIndex;
 import org.apache.carbondata.core.indexstore.PartitionSpec;
-import org.apache.carbondata.core.locks.CarbonLockFactory;
-import org.apache.carbondata.core.locks.ICarbonLock;
-import org.apache.carbondata.core.locks.LockUsage;
 import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier;
 import org.apache.carbondata.core.metadata.SegmentFileStore;
 import org.apache.carbondata.core.metadata.schema.table.CarbonTable;
@@ -227,34 +224,9 @@ public final class DeleteLoadFolders {
     if (details != null && details.length != 0) {
       for (LoadMetadataDetails oneLoad : details) {
         if (checkIfLoadCanBeDeleted(oneLoad, isForceDelete)) {
-          ICarbonLock segmentLock = CarbonLockFactory.getCarbonLockObj(absoluteTableIdentifier,
-              CarbonTablePath.addSegmentPrefix(oneLoad.getLoadName()) + LockUsage.LOCK);
-          try {
-            if (oneLoad.getSegmentStatus() == SegmentStatus.INSERT_OVERWRITE_IN_PROGRESS
-                || oneLoad.getSegmentStatus() == SegmentStatus.INSERT_IN_PROGRESS) {
-              if (segmentLock.lockWithRetries(1, 5)) {
-                LOGGER.info("Info: Acquired segment lock on segment:" + oneLoad.getLoadName());
-                LoadMetadataDetails currentDetails =
-                    getCurrentLoadStatusOfSegment(oneLoad.getLoadName(), metadataPath);
-                if (currentDetails != null && checkIfLoadCanBeDeleted(currentDetails,
-                    isForceDelete)) {
-                  oneLoad.setVisibility("false");
-                  isDeleted = true;
-                  LOGGER.info("Info: Deleted the load " + oneLoad.getLoadName());
-                }
-              } else {
-                LOGGER.info("Info: Load in progress for segment" + oneLoad.getLoadName());
-                return isDeleted;
-              }
-            } else {
-              oneLoad.setVisibility("false");
-              isDeleted = true;
-              LOGGER.info("Info: Deleted the load " + oneLoad.getLoadName());
-            }
-          } finally {
-            segmentLock.unlock();
-            LOGGER.info("Info: Segment lock on segment:" + oneLoad.getLoadName() + " is released");
-          }
+          oneLoad.setVisibility("false");
+          isDeleted = true;
+          LOGGER.info("Info: Deleted the load " + oneLoad.getLoadName());
         }
       }
     }
--- a/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -57,7 +57,6 @@ import org.apache.carbondata.core.segmentmeta.SegmentMetaDataInfo
 import org.apache.carbondata.core.statusmanager.{LoadMetadataDetails, SegmentStatus, SegmentStatusManager, SegmentUpdateStatusManager}
 import org.apache.carbondata.core.util.{CarbonProperties, CarbonSessionInfo, CarbonUtil, SessionParams, ThreadLocalSessionInfo}
 import org.apache.carbondata.core.util.path.CarbonTablePath
-import org.apache.carbondata.core.view.{MVSchema, MVStatus}
 import org.apache.carbondata.events.{OperationContext, OperationListenerBus}
 import org.apache.carbondata.indexserver.{DistributedRDDUtils, IndexServer}
 import org.apache.carbondata.processing.loading.FailureCauses
@@ -267,9 +266,8 @@ object CarbonDataRDDFactory {
             throw new Exception("Exception in compaction " + exception.getMessage)
           }
         } finally {
-          executor.shutdownNow()
           try {
-            compactor.deletePartialLoadsInCompaction()
+            executor.shutdownNow()
           } catch {
             // no need to throw this as compaction is over
             case ex: Exception =>
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/DeleteExecution.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/DeleteExecution.scala
@@ -374,8 +374,6 @@ object DeleteExecution {
               blockMappingVO.getSegmentNumberOfBlockMapping)
           }
         } else {
-          // In case of failure , clean all related delete delta files
-          CarbonUpdateUtil.cleanStaleDeltaFiles(carbonTable, timestamp)
           val errorMsg =
             "Delete data operation is failed due to failure in creating delete delta file for " +
             "segment : " + resultOfBlock._2._1.getSegmentName + " block : " +
@@ -410,8 +408,6 @@ object DeleteExecution {
       LOGGER.info(s"Delete data operation is successful for " +
                   s"${ carbonTable.getDatabaseName }.${ carbonTable.getTableName }")
     } else {
-      // In case of failure , clean all related delete delta files
-      CarbonUpdateUtil.cleanStaleDeltaFiles(carbonTable, timestamp)
       val errorMessage = "Delete data operation is failed due to failure " +
                          "in table status update."
       LOGGER.error("Delete data operation is failed due to failure in table status update.")
@@ -440,8 +436,6 @@ object DeleteExecution {
               blockMappingVO.getSegmentNumberOfBlockMapping)
           }
         } else {
-          // In case of failure , clean all related delete delta files
-          CarbonUpdateUtil.cleanStaleDeltaFiles(carbonTable, timestamp)
           val errorMsg =
             "Delete data operation is failed due to failure in creating delete delta file for " +
             "segment : " + resultOfBlock._2._1.getSegmentName + " block : " +
--- a/processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
+++ b/processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
@@ -40,7 +40,6 @@ import org.apache.carbondata.core.datastore.impl.FileFactory;
 import org.apache.carbondata.core.index.Segment;
 import org.apache.carbondata.core.locks.CarbonLockUtil;
 import org.apache.carbondata.core.locks.ICarbonLock;
-import org.apache.carbondata.core.locks.LockUsage;
 import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier;
 import org.apache.carbondata.core.metadata.SegmentFileStore;
 import org.apache.carbondata.core.metadata.schema.table.CarbonTable;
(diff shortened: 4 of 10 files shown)


CarbonDataQA1:

jackylk:
add to whitelist

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2362/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4104/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4112/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2370/

Zhangshunyu:
From this pr we can see some delete actions are removed, as a result there would be some unused files left, any idea about how to handle the left files?

Pickupolddriver:
> From this pr we can see some delete actions are removed, as a result there would be some unused files left, any idea about how to handle the left files?

The stale data will remain there until the user calls clean files action manually. Since the data will be check by the table status file, the stale data will not do harm to data management. Since data deleted by accident is a critical problem, but the stale data exist is acceptable.

akashrn5:
during insert or load, if you dont clean up still fine, but if you remove `deletePartialLoadsInCompaction` directly, it can create problem of load or query failures or wrong data in query or extra data. This is because we dont enter any entry before compaction in table status in order to maintain the segment ID logic of compaction. So its dangerous to remove this blindly. 

I remember @ajantha-bhat faced similar issues recently during compaction.

ajantha-bhat:
Agree with @Zhangshunyu and @akashrn5 
a) When compaction retried as it use the same segment ID, if stale files are not cleaned. It gives duplicate data.
So, before this change, we need #3934 to be merged which can use a unique segment id for compaction retry.
b) please check and move the logic of deletePartialLoadsInCompaction in clean files command, instead of permanently removing it. If the clean files don't have this logic, it may not able to clean stale files
c) Also if the purpose of this PR at deleting accidental data loss. you need to handle `cleanStaleDeltaFiles` in` CarbonUpdateUtil.java` and also identify other places. Just handling in once place will not guarantee that we cannot have data loss.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -267,9 +266,8 @@ object CarbonDataRDDFactory {
             throw new Exception("Exception in compaction " + exception.getMessage)
           }
         } finally {
-          executor.shutdownNow()
           try {
-            compactor.deletePartialLoadsInCompaction()
```
ajantha-bhat, line 272: a) When compaction retries, it uses the same segment ID, if stale files are not cleaned. It gives duplicate data.
So, before this change, we need #3934 to be merged which can use a unique segment id for compaction retry.

b) please check and move the logic of deletePartialLoadsInCompaction in clean files command, instead of permanently removing it. If the clean files don't have this logic, it may not able to clean stale files.

c) Also if the purpose of this PR is to avoid accidental data loss. you need to handle `cleanStaleDeltaFiles` in `CarbonUpdateUtil.java` and also identify other places. Just handling in few place will not guarantee that we cannot have data loss.

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2495/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4239/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
Pickupolddriver: (a). My PR will dependent on #3934 , if it merged first we don't need to worry about the same segment ID in compaction situations. So it the deletePartialLoadsInCompaction could be deleted. 

(b). In this PR's design, the stale files would not be deleted automatically. Users need to call cleanFiles function to clean them besides PR #3917 can handle the clean files function.

(c). More places of calling claenStaleFiles have been removed.

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4352/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2602/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
QiangCai: a) root cause of stale files:  It uses listFiles to collect index files when writing the segment file, so it will add stale index file names into segment file. 
  it is hard to add a unique id, so we change the plan to avoid to listFiles during wirte segment file.

akashrn5: @QiangCai how its handled now, without list files? why cant we do list files with the timestamp filter, which is load timestamp/fact timestamp, we can get from load model or somewhere right?

akashrn5: > We cannot remove the clean stale files in case of IUD and wait for clean files command to clean them, we should immediately clean the stale ones in the respective command itself, as there will be chances of extra data or data inconsistency.

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4363/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2613/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2625/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4376/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
QiangCai: @akashrn5 After we avoid using listFiles during loading,  the stale segment (for example 0.1) will not impact data consistency. If the stale segment has stale index files and data files,  we will not add them to segment file.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/CarbonProjectForDeleteCommand.scala
```diff
@@ -108,9 +108,6 @@ private[sql] case class CarbonProjectForDeleteCommand(
       }
       val executorErrors = ExecutionErrors(FailureCauses.NONE, "")
 
-      // handle the clean up of IUD.
-      CarbonUpdateUtil.cleanUpDeltaFiles(carbonTable, false)
```
QiangCai, line 112: if we don't clean up stale delta files, it will be used after the next update.
maybe we can't remove it now.

Pickupolddriver: Yes, the purpose of PR is to prevent delete data by accident, after checking the code of `cleanUpDeltaFiles` function, I think it's safe to keep it here since it will only delete the files in the aborted scenario and it's crucial for data accuracy. I will revert removing this function.

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2672/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4426/

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonDataRDDFactory.scala
Pickupolddriver: @ajantha-bhat after the merge of #3978 , cleanStaleDeltaFiles will only be called when there are exceptions during the update process. And it will only delete the delta and index file created by the exception update.

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4429/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2675/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/CarbonProjectForDeleteCommand.scala
```diff
@@ -149,14 +149,10 @@ private[sql] case class CarbonProjectForDeleteCommand(
       case e: HorizontalCompactionException =>
         LOGGER.error("Delete operation passed. Exception in Horizontal Compaction." +
                      " Please check logs. " + e.getMessage)
-        CarbonUpdateUtil.cleanStaleDeltaFiles(carbonTable, e.compactionTimeStamp.toString)
         Seq(Row(0L))
 
       case e: Exception =>
         LOGGER.error("Exception in Delete data operation " + e.getMessage, e)
-        // ****** start clean up.
```
akashrn5, line 157: i dont think we can remove this directly, it might create problem, as mentioned in comments

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/DeleteExecution.scala
```diff
@@ -374,8 +374,6 @@ object DeleteExecution {
               blockMappingVO.getSegmentNumberOfBlockMapping)
           }
         } else {
-          // In case of failure , clean all related delete delta files
-          CarbonUpdateUtil.cleanStaleDeltaFiles(carbonTable, timestamp)
```
akashrn5, line 378: same as above

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4458/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2705/

CarbonDataQA1:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2717/

CarbonDataQA1:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4471/

QiangCai:
retest this please

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2878/

CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4634/

core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
```diff
@@ -227,34 +224,9 @@ public static boolean deleteLoadFoldersFromFileSystem(
     if (details != null && details.length != 0) {
       for (LoadMetadataDetails oneLoad : details) {
         if (checkIfLoadCanBeDeleted(oneLoad, isForceDelete)) {
-          ICarbonLock segmentLock = CarbonLockFactory.getCarbonLockObj(absoluteTableIdentifier,
```
QiangCai, line 230: after this PR, method deleteLoadFoldersFromFileSystem will be invoked by clean files only, right?
if yes, why we remove this code?

processing/src/main/java/org/apache/carbondata/processing/util/CarbonLoaderUtil.java
```diff
@@ -323,10 +320,6 @@ public static boolean recordNewLoadMetadata(LoadMetadataDetails newMetaEntry,
             for (LoadMetadataDetails entry : listOfLoadFolderDetails) {
```
QiangCai, line 320: Let's discuss insertOverwrite behavior.
Should we delete old data immediately?

Pickupolddriver: Seems in `insertOverwrite` process, the stale data can be removed immediately.  I think we can keep the original `addToStaleFolders` function here

> Let's discuss insertOverwrite behavior.
> Should we delete old data immediately?

Seems in `insertOverwrite` process, the stale data can be removed immediately.  I think we can keep the original `addToStaleFolders` function here

Pickupolddriver: > Let's discuss insertOverwrite behavior.
> Should we delete old data immediately?

Seems in `insertOverwrite` process, the stale data can be removed immediately.  I think we can keep the original `addToStaleFolders` function here

Pickupolddriver: Seems in `insertOverwrite` process, the stale data can be removed immediately.  I think we can keep the original `addToStaleFolders` function here

core/src/main/java/org/apache/carbondata/core/util/DeleteLoadFolders.java
Pickupolddriver: Actually, this part of the modifications comes from #3971 contributed by @Indhumathi27 .  It removes the condition of insert_in_progress and lock acquirement, I think it may delete the segments which are in progress. So maybe I can revoke this modification.

QiangCai:
@Pickupolddriver I will rework by basing on this PR and raise another PR.  Please close this PR, thanks for your contribution.

Pickupolddriver:
> @Pickupolddriver I will rework by basing on this PR and raise another PR. Please close this PR, thanks for your contribution.

OK, PR closed.

PR #3944, opened 2020-09-21
[CARBONDATA-4001] Fix SI global sort load on partition table

ajantha-bhat:
### Why is this PR needed?
On a partition table,  when SI is created with global sort and when data is loaded. It shows 0 rows as main table query results  is 0 rows.
 
For partition table, local sort SI flow `current.segmentfile` is set in `CarbonSecondaryIndexRDD`
For the global sort, this value was not set. so, the main table query was resulting in 0 rows. Setting this value for global sort flow also.
    
 - No

 - Yes

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
@@ -466,7 +466,7 @@ public abstract class CarbonInputFormat<T> extends FileInputFormat<Void, T> {
    * get data blocks of given segment
    */
   protected List<CarbonInputSplit> getDataBlocksOfSegment(JobContext job, CarbonTable carbonTable,
-      IndexFilter expression, List<Segment> segmentIds,
+      IndexFilter expression, List<Segment> validSegments,
       List<Segment> invalidSegments, List<String> segmentsToBeRefreshed)
       throws IOException {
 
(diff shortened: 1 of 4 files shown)


CarbonDataQA1:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4152/

CarbonDataQA1:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/2411/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
```diff
@@ -466,15 +466,15 @@ public IndexFilter getFilterPredicates(Configuration configuration) {
    * get data blocks of given segment
    */
   protected List<CarbonInputSplit> getDataBlocksOfSegment(JobContext job, CarbonTable carbonTable,
-      IndexFilter expression, List<Segment> segmentIds,
+      IndexFilter expression, List<Segment> validSegments,
```
kunal642, line 469: revert this change.. Not related to this PR

ajantha-bhat: Actually when the main table was giving 0 rows. I was debugging this flow. And this naming is wrong. segmentID should be string, here it is not the segmentID but the complete Segment object. so renamed. Still you want to revert it ?

kunal642: its fine.. ill merge

kunal642:
LGTM

PR #4009, opened 2020-11-18
[CARBONDATA-4029] Fix Old Timestamp issue in alter add segement

Karan980:
### Why is this PR needed?
Earlier timestamp present in name of carbondata files was in nanoseconds. Currently the timestamp is in milliseconds. When old SDK file segment is added to table through alter table add segment query then it is treated as invalid block due to timestamp present in nanoseconds.
 
Removed update validation for SDK written files.

    
 - No

 - Yes

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
@@ -66,6 +66,7 @@ import org.apache.carbondata.core.util.path.CarbonTablePath;
 import org.apache.carbondata.hadoop.CarbonInputSplit;
 
 import com.google.common.collect.Sets;
+import org.apache.commons.lang.ArrayUtils;
 import org.apache.hadoop.fs.BlockLocation;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
@@ -368,7 +369,10 @@ public class CarbonTableInputFormat<T> extends CarbonInputFormat<T> {
       if (isIUDTable) {
         // In case IUD is not performed in this table avoid searching for
         // invalidated blocks.
-        if (CarbonUtil
+        String segmentId = getSegmentIdFromFilePath(inputSplit.getFilePath());
+        // SDK segments have segment id as null. If a segment has segmentId as null then no need
+        // to check updateValidation, because that segment is not created through update operation.
+        if (!segmentId.equalsIgnoreCase("null") && CarbonUtil
             .isInvalidTableBlock(inputSplit.getSegmentId(), inputSplit.getFilePath(),
                 invalidBlockVOForSegmentId, updateStatusManager)) {
           continue;
@@ -571,5 +575,19 @@ public class CarbonTableInputFormat<T> extends CarbonInputFormat<T> {
 
   public void setReadCommittedScope(ReadCommittedScope readCommittedScope) {
     this.readCommittedScope = readCommittedScope;
+  }
+
+  public String getSegmentIdFromFilePath(String filePath) {
+    String tempFilePath = filePath.replace(
+        CarbonCommonConstants.WINDOWS_FILE_SEPARATOR, CarbonCommonConstants.FILE_SEPARATOR);
+    int fileNameStartIndex = tempFilePath.lastIndexOf(CarbonCommonConstants.FILE_SEPARATOR);
+    String fileName = tempFilePath.substring(fileNameStartIndex);
+    if (fileName != null && !fileName.isEmpty()) {
+      String[] pathElements = fileName.split(CarbonCommonConstants.DASH);
+      if (ArrayUtils.isNotEmpty(pathElements)) {
+        return pathElements[pathElements.length - 2];
+      }
+    }
+    return CarbonCommonConstants.INVALID_SEGMENT_ID;
   }
 }
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
@@ -781,6 +781,48 @@ class AddSegmentTestCase extends QueryTest with BeforeAndAfterAll {
     sql(s"drop table $tableName")
   }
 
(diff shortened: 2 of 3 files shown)


hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -572,4 +574,18 @@ public ReadCommittedScope getReadCommitted(JobContext job, AbsoluteTableIdentifi
   public void setReadCommittedScope(ReadCommittedScope readCommittedScope) {
     this.readCommittedScope = readCommittedScope;
   }
+
+  public String getSegmentIdFromFilePath(String filePath) {
```
ajantha-bhat, line 578: Already `CarbonTablePath#getSegmentIdFromPath` method exist. check if you can use it.

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
```diff
@@ -781,6 +781,26 @@ class AddSegmentTestCase extends QueryTest with BeforeAndAfterAll {
...
+    sql(
+      s"""
+         |create table external_primitive (id int, name string, rank smallint, salary double,
+         | active boolean, dob date, doj timestamp, city string, dept string) stored as carbondata
+         |""".stripMargin)
+    val externalSegmentPathWithOldTimestamp = storeLocation + "/" +
+      "external_segment_with_old_timestamp"
+    val externalSegmentPath = storeLocation + "/" + "external_segment"
+    FileFactory.deleteAllFilesOfDir(new File(externalSegmentPath))
+    copy(externalSegmentPathWithOldTimestamp, externalSegmentPath)
```
ajantha-bhat, line 795: We should not keep binary files in the repo (like carbondata and index files). It has to generate on the fly.
Try to write from SDK and use nano time timeStamp by calling `CarbonWriterBuilder#uniqueIdentifier`

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
```diff
@@ -367,8 +368,9 @@ protected FileSplit makeSplit(String segmentId, String filePath, long start, lon
       String[] deleteDeltaFilePath = null;
       if (isIUDTable) {
         // In case IUD is not performed in this table avoid searching for
-        // invalidated blocks.
-        if (CarbonUtil
+        // invalidated blocks. No need to check validation for splits written by SDK.
+        String segmentId = getSegmentIdFromFilePath(inputSplit.getFilePath());
```
ajantha-bhat, line 372: add more comments explaining SDK case segment id will always be null and using that as a decision point to skip this validation for SDK and why you need to skip it.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
Karan980: _CarbonTablePath#getSegmentIdFromPath_ returns segmentId present in path. For eg: Fact/Part0/segment_**0**. But in this case we need segment id present in carbondata file name. So above method cannot be used.

ajantha-bhat: The caller of this method, inputSplit itself has getSegmentId() method. I think you can use it instead of this. Please avoid writing new code and try to reuse the code.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
Karan980: Done

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
Karan980: getSegmentId() also return segmentId from LoadMetaDataDetails of segment which is not null. Only null for segmentId is present in name of carbondata file of SDK segment and there is no already existing method to get that.

integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
Karan980: Done

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4859/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3106/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonTableInputFormat.java
ajantha-bhat: ok

ajantha-bhat:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4867/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3114/

Karan980:
retest this please

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3122/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4875/

ajantha-bhat:
LGTM

ajantha-bhat:
@Karan980 : I tried to merge,
But when I checkout using `git fetch origin pull/4009/head:4009`, this 4009 is pointing some old PR. I guess some problem with this PR. Can you verify once. If same problem just raise a new PR.

PR #4013, opened 2020-11-19
[CARBONDATA-4062] Make clean files as data trash manager

QiangCai:
### Why is this PR needed?
 To prevent accidental deletion of data, carbon will introduce data trash management. It will provide buffer time for accidental deletion of data to roll back the delete operation.

Data trash management is a part of carbon data lifecycle management. Clean files as a data trash manager should contain the following two parts.
part 1: manage metadata-indexed data trash.
  This data is at the original place of the table and indexed by metadata. carbon manages this data by metadata index and should avoid using listFile() interface.
part 2: manage ".Trash" folder.
   Now ".Trash" folder is without metadata index, and the operation on it bases on timestamp and listFile() interface. In the future, carbon will index ".Trash" folder to improve data trash management.
 
1. remove data clean function from all features, but keep exception-handling part
  Notes: the following features still clean data
  a) drop table/database/partition/index/mv
  b) insert/load overwrite table/partition
2. only clean files function works as a data trash manager now
3. support concurrent operation with other feature(loading, compaction, update/delete, and so on)

 - No

 - No


akashrn5:
@QiangCai i think we need to wait before considering these changes, because already multiple people are working on same area

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3208/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4963/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4965/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3210/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4968/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3213/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4984/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3229/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4994/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3239/

QiangCai:
@akashrn5 @ajantha-bhat @vikramahuja1001 please review this PR

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5025/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3269/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5034/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3277/

core/src/main/java/org/apache/carbondata/core/util/TrashUtil.java
```diff
@@ -208,6 +205,14 @@ public static boolean isTrashRetentionTimeoutExceeded(long fileTimestamp) {
     return difference > retentionMilliSeconds;
   }
 
+  /**
+   * whether trash data outside of .Trash folder is time out
+   */
+  public static boolean isTrashDataTimeout(long fileTimestamp) {
+    return isTrashRetentionTimeoutExceeded(fileTimestamp) &&
```
ajantha-bhat, line 212: In the default case, when isTrashRetentionTimeoutExceeded  , isMaxQueryTimeoutExceeded will always be true. 
why not remove `queryMaxTimeout` and in query flow clean the metadata indexed trash based on `trashRetentionTimeout` ??

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
```diff
@@ -1044,7 +1045,7 @@ public static void cleanSegments(CarbonTable table,
           Long fileTimestamp = CarbonUpdateUtil.getTimeStampAsLong(indexFile
               .substring(indexFile.lastIndexOf(CarbonCommonConstants.HYPHEN) + 1,
                   indexFile.length() - CarbonTablePath.INDEX_FILE_EXT.length()));
-          if (CarbonUpdateUtil.isMaxQueryTimeoutExceeded(fileTimestamp) || forceDelete) {
+          if (TrashUtil.isTrashDataTimeout(fileTimestamp) || forceDelete) {
```
ajantha-bhat, line 1048: Also update this behavior change in document, now `max.query.execution.time`  expire will not delete the data but trash time has to expire also.

core/src/main/java/org/apache/carbondata/core/util/CleanFilesUtil.java
```diff
@@ -146,8 +146,8 @@ public static void cleanStaleSegmentsForPartitionTable(CarbonTable carbonTable)
    * in the metadata folder and is not present in the table status file is considered as a
    * stale segment. Only comparing from tablestatus file, not checking tablestatus.history file
    */
-  private static void getStaleSegmentFiles(CarbonTable carbonTable, List<String> staleSegmentFiles,
-      List<String> redundantSegmentFile) {
+  private static void collectStaleSegmentFiles(CarbonTable carbonTable,
```
ajantha-bhat, line 149: I suggest `getStaleSegmentFiles` is fine, but make a list inside the function and send as a return value.  Passing as an argument makes sense only when the list has some pre-entries or used in recursion.

QiangCai: this method need return two lists, so it passes two lists in parameters

core/src/main/java/org/apache/carbondata/core/util/TrashUtil.java
```diff
@@ -192,12 +192,9 @@ public static void emptyTrash(String tablePath) {
   }
 
   /**
-   * This will tell whether the trash retention time has expired or not
-   *
-   * @param fileTimestamp
-   * @return
+   * whether trash data inside of .Trash folder is time out
```
ajantha-bhat, line 195: now we are using this for data outside trash also, so change the description.

`returns true, if the current timestamp has expired based on carbon.trash.retention.days `

core/src/main/java/org/apache/carbondata/core/util/TrashUtil.java
QiangCai: isTrashRetentionTimeoutExceeded:  whether trash data inside of .Trash folder is time out
isTrashDataTimeout: whether trash data outside of .Trash folder is time out

TrashRetentionTimeout can be 0, but MaxQueryTimeout not.

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
QiangCai: yes, the requirement is just like it.

core/src/main/java/org/apache/carbondata/core/util/TrashUtil.java
QiangCai: no

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
```diff
@@ -0,0 +1,112 @@
...
+    val absoluteTableIdentifier = carbonTable.getAbsoluteTableIdentifier
+    try {
+      val errorMsg = "Clean files request is failed for " +
+        s"${ carbonTable.getQualifiedName }" +
+        ". Not able to acquire the clean files lock due to another clean files " +
+        "operation is running in the background."
+      carbonCleanFilesLock = CarbonLockUtil.getLockObject(absoluteTableIdentifier,
+        LockUsage.CLEAN_FILES_LOCK, errorMsg)
+      // step 1: clean trash folder
+      cleanTrashFolder(carbonTable, force)
```
ajantha-bhat, line 63: `checkAndCleanTrashFolder`  
As we dont clean always. Function name give a feeling that we are cleaning trash folder here

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
```diff
@@ -0,0 +1,112 @@
...
+      // step 2: clean stale segments which are not exists in metadata
+      cleanStaleSegments(carbonTable)
```
ajantha-bhat, line 65: `moveStaleSegmentsToTrash` is the better name for this function ?

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/CarbonTruncateCommand.scala
```diff
@@ -18,39 +18,31 @@
...
       throw new MalformedCarbonCommandException(
         "Unsupported truncate table with specified partition")
     }
-    CarbonCleanFilesCommand(
-      databaseNameOp = Option(dbName),
-      tableName = Option(tableName),
-      None,
-      truncateTable = true
-    ).run(sparkSession)
+
```
ajantha-bhat, line 43: I didn't see this command in any document. If possible please update it. I think this will just empty the table by removing all the data (success, fail, stale)

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
```diff
@@ -482,176 +482,6 @@ public boolean accept(CarbonFile file) {
 
   }
 
-  /**
-   * Handling of the clean up of old carbondata files, index files , delete delta,
-   * update status files.
-   * @param table clean up will be handled on this table.
-   * @param forceDelete if true then max query execution timeout will not be considered.
-   */
-  public static void cleanUpDeltaFiles(CarbonTable table, boolean forceDelete) throws IOException {
```
ajantha-bhat, line 491: I didn't see delta file cleanup from the trashManager, will there be any impact because of this ? we have to test with stale delta files once before removing I think.

ajantha-bhat:
@QiangCai : I finished review, have you tested with some stale delta files ? Now I think it won't be cleaned up forever (when segment is not stale but only delta files are stale) as code is removed.

core/src/main/java/org/apache/carbondata/core/util/TrashUtil.java
ajantha-bhat: ok

core/src/main/java/org/apache/carbondata/core/util/CleanFilesUtil.java
ajantha-bhat: ok, doesn't make much difference

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
akashrn5: agree with @ajantha-bhat , please test and also this method used to take care of the aborted scenario, now how its being taken care?

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
```diff
@@ -482,176 +482,6 @@ public boolean accept(CarbonFile file) {
...
-      }
-    }
-  }
-
-  /**
-   * this is the clean up added specifically for SI table, because after we merge the data files
-   * inside the secondary index table, we need to delete the stale carbondata files.
-   * refer org.apache.spark.sql.secondaryindex.rdd.CarbonSIRebuildRDD
-   */
-  private static void cleanUpDataFilesAfterSmallFilesMergeForSI(CarbonTable table,
```
akashrn5, line 615: how is this scenario handled after removing?

integration/spark/src/main/scala/org/apache/carbondata/events/CleanFilesEvents.scala
```diff
@@ -34,5 +34,6 @@ case class CleanFilesPreEvent(carbonTable: CarbonTable, sparkSession: SparkSessi
  * @param carbonTable
  * @param sparkSession
  */
-case class CleanFilesPostEvent(carbonTable: CarbonTable, sparkSession: SparkSession)
-  extends Event with CleanFilesEventInfo
+case class CleanFilesPostEvent(carbonTable: CarbonTable,
```
akashrn5, line 37: move carbontable also to next line

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
```diff
@@ -90,7 +90,6 @@ class CarbonTableCompactor(carbonLoadModel: CarbonLoadModel,
 
     while (loadsToMerge.size() > 1 || needSortSingleSegment(loadsToMerge)) {
       val lastSegment = sortedSegments.get(sortedSegments.size() - 1)
-      deletePartialLoadsInCompaction()
```
akashrn5, line 93: if you are removing this, how are you handling the case of segment number same in next compaction retry as we dont have any entry for table status file during compaction.

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
```diff
@@ -0,0 +1,112 @@
...
+  private val LOGGER = LogServiceFactory.getLogService(this.getClass.getCanonicalName)
+
+  /**
+   * clean garbage data
+   *  1. clean .Trash folder
+   *  2. clean stale segments without metadata
+   *  3. clean expired segments (MARKED_FOR_DELETE, Compacted, In Progress)
+   *
+   * @param carbonTable     : CarbonTable Object
+   * @param partitionSpecs : Hive Partitions  details
```
akashrn5, line 47: please remove @params, name suggests what it is clearly

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/index/DropIndexCommand.scala
```diff
@@ -210,12 +210,6 @@ private[sql] case class DropIndexCommand(
           logError("Table metadata unlocking is unsuccessful, index table may be in stale state")
         }
       }
-      // in case if the the physical folders still exists for the index table
-      // but the carbon and hive info for the index table is removed,
-      // DROP INDEX IF EXISTS should clean up those physical directories
-      if (ifExistsSet && carbonTable.isEmpty) {
```
akashrn5, line 216: i think we should keep this change, from the comment its clear that there will be such scenarios and we have observed many times

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/mutation/CarbonTruncateCommand.scala
QiangCai: let clean files to remove data.

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
QiangCai: only keep cleanStaleDeltaFiles to handle exceptions.

clean files should work with the concurrent update, so remove it.

integration/spark/src/main/scala/org/apache/carbondata/events/CleanFilesEvents.scala
QiangCai: ok

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
QiangCai: In this case, clean files have no chance to clean data.
keep the stale data in the folder, it will not impact compaction.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/index/DropIndexCommand.scala
QiangCai: create table t1...
create table t2...
drop index t2 on t1

this drop index sql will remove table folder of t2 by mistake.

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
QiangCai: ok

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
QiangCai: clean files on the main table will trigger clean files of the index tables

integration/spark/src/main/scala/org/apache/carbondata/events/CleanFilesEvents.scala
QiangCai: ok

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
QiangCai: ok

integration/spark/src/main/scala/org/apache/carbondata/trash/DataTrashManager.scala
QiangCai: ok

core/src/main/java/org/apache/carbondata/core/mutate/CarbonUpdateUtil.java
akashrn5: cleanStaleDeltaFiles will be called only in case of exception here, please see below points
1. in Projectfordeletecommand, its called two times both in exception  case and finally block, thats wrong, it should be only in finally block
2. its called multiple times in `DeleteExecution ` class which needs to be checked and avoid it.
3. since its handled only in case of exception case, how its taken care in case of application crash or shutdown scenario?
4. cleanStaleDeltaFiles  had cases of handling the update and delete aborted case, how its handled now, as the complete method is removed?

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
akashrn5: but the stale data will always be present inside segment folder right? by chance assume if the segment file is corrupted or deleted, still carbon should pass the query, which it does by listing, in that case we will get the wrong data or query will always fail. I think we need to have some way to clean them, what you think @QiangCai @ajantha-bhat

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/index/DropIndexCommand.scala
akashrn5: are you saying that, even though the t2 is not an index, and i call drop index on t2, it will consider as a index and drop t2 table folder? If so, we should handle there and keep this code here. This was handled specially for SI i think in negative scenarios

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5050/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3292/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/index/DropIndexCommand.scala
ajantha-bhat: I agree with @akashrn5 : If code is buggy, we need to fix it. If we remove it. It might bring back the original issues that caused this code change.

integration/spark/src/main/scala/org/apache/carbondata/spark/rdd/CarbonTableCompactor.scala
ajantha-bhat: I think this will go as a stale segment in trash folder when clean files is called right ?

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5065/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3306/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5077/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3319/

ajantha-bhat:
@QiangCai : please rebase as #4035 got merged.

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5056/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3297/

PR #4033, opened 2020-12-01
[CARBONDATA-4066] data mismatch observed with SI and without SI when SI global sort and SI segment merge is true

maheshrajus:
### Why is this PR needed?
 data mismatch observed with SI and without SI when SI global sort and SI segment merge is true. After merge si data files position reference is also sorted and due to this pointing to wrong position reference causing data mismatch with SI and without SI
 
no need to calculate the position references after data files merge should use existed position reference column from SI table.
    
 - No

 - Yes


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/4990/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3235/

ajantha-bhat:
LGTM

PR #4047, opened 2020-12-07
[CARBONDATA-4078] Add external segment and query with index server fails

ShreelekhyaG:
### Why is this PR needed?
Query after adding an external segment to carbon table tries to `getSplits `from Index server and throws an exception as it cannot read the external(orc/parquet) file format.
When the fallback mode is disabled, it throws an exception and fails.
 
To avoid the exception, filtered only valid carbon segments to cache in the index server.
    
 - No

 - No, tested in cluster.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/indexserver/IndexJobs.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/indexserver/IndexJobs.scala
@@ -47,6 +47,11 @@ class DistributedIndexJob extends AbstractIndexJob {
 
   override def execute(indexFormat: IndexInputFormat,
       configuration: Configuration): util.List[ExtendedBlocklet] = {
+    // get only carbon segments.
+    if (indexFormat.getValidSegments != null) {
+      indexFormat.setValidSegments(indexFormat.getValidSegments.asScala
+        .filter(segment => segment.isCarbonSegment).toList.asJava)
+    }
     if (LOGGER.isDebugEnabled) {
       val messageSize = SizeEstimator.estimate(indexFormat)
       LOGGER.debug(s"Size of message sent to Index Server: $messageSize")
--- a/core/src/main/java/org/apache/carbondata/core/index/IndexInputFormat.java
+++ b/core/src/main/java/org/apache/carbondata/core/index/IndexInputFormat.java
@@ -418,6 +418,10 @@ public class IndexInputFormat extends FileInputFormat<Void, ExtendedBlocklet>
     return validSegments;
   }
 
+  public void setValidSegments(List<Segment> validSegments) {
+    this.validSegments = validSegments;
+  }
+
   public void createIndexChooser() throws IOException {
     if (null != filterResolverIntf) {
       this.indexChooser = new IndexChooser(table);
--- a/integration/spark/src/main/scala/org/apache/spark/sql/listeners/PrePrimingListener.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/listeners/PrePrimingListener.scala
@@ -33,9 +33,11 @@ object PrePrimingEventListener extends OperationEventListener {
       operationContext: OperationContext): Unit = {
     val prePrimingEvent = event.asInstanceOf[IndexServerLoadEvent]
     val carbonTable = prePrimingEvent.carbonTable
+    // get only carbon segments.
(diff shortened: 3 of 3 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5101/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3340/

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
```diff
@@ -290,6 +291,9 @@ private static FileInputFormat createIndexJob(CarbonTable carbonTable,
       Boolean isFallbackJob, List<String> segmentsToBeRefreshed, boolean isCountJob,
       Configuration configuration) {
     List<String> invalidSegmentNo = new ArrayList<>();
```
vikramahuja1001, line 293: This is not the only place where we call IndexJobs from, please check it in all the places and add there. Also check it in the case of prepriming too, you can ignore it if it's a non carbon segment.

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5145/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3383/

integration/spark/src/main/scala/org/apache/carbondata/indexserver/IndexJobs.scala
```diff
@@ -47,6 +47,9 @@ class DistributedIndexJob extends AbstractIndexJob {
 
   override def execute(indexFormat: IndexInputFormat,
       configuration: Configuration): util.List[ExtendedBlocklet] = {
+    // get only carbon segments.
+    indexFormat.setValidSegments(indexFormat.getValidSegments.asScala
```
Indhumathi27, line 51: In Fall back mode, this scenario is handled?

Indhumathi27:
@ShreelekhyaG Check and handle in BroadCastSIFilterPushJoin

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5160/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3398/

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
ShreelekhyaG: Done. Added in IndexJobs and in pre-priming.

integration/spark/src/main/scala/org/apache/carbondata/indexserver/IndexJobs.scala
ShreelekhyaG: Yes, in the fallback mode it uses `getCarbonSegments` method.

Indhumathi27: In EmbeddedIndexJob fall back mode, i think, you have to handle

ShreelekhyaG: in EmbeddedIndexJob fallback mode, we have `toDistributable ` method, here `loadmetadatadetails `for segments are present and carbon segments are filtered.

ShreelekhyaG:
> @ShreelekhyaG Check and handle in BroadCastSIFilterPushJoin

Tested queries in cluster with SI, no issues.

Indhumathi27:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5177/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3415/

Indhumathi27:
LGTM

PR #4053, opened 2020-12-15
[WIP] Using listfiles instead of listDirs, as listDirs do not work when fileStatus is null

vikramahuja1001:
### Why is this PR needed?
 In the case when AbstractDFSCarbonFile is created with path and HadoopConf instead of fileStatus and if the file does not exist, since fileStatus is empty ListDirs returns empty result and getAbsolutePath throws file does not exist exception
 
Using listFiles() operation and checking if it is a directory.
    
 - No

 - No


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5167/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3405/

PR #4057, opened 2020-12-17
[CARBONDATA-4088] Drop metacache didn't clear some cache information …

jack86596:
…which leads to memory leak

 ### Why is this PR needed?
When there are two spark applications, one drop a table, some cache information of this table stay in another application and cannot be removed with any method like "Drop metacache" command. This leads to memory leak. With the passage of time, memory leak will also accumulate which finally leads to driver OOM. Following are the leak points: 1) tableModifiedTimeStore in CarbonFileMetastore; 2) segmentLockMap in BlockletDataMapIndexStore; 3) absoluteTableIdentifierByteMap in SegmentPropertiesAndSchemaHolder; 4) tableInfoMap in CarbonMetadata.
 
Using expiring map to cache the table information in CarbonMetadata and modified time in CarbonFileMetaStore so that stale information will be cleared automatically after the expiration time. Operations in BlockletDataMapIndexStore no need to be locked, remove all the logic related to segmentLockMap.
    
 - New configuration carbon.metacache.expiration.seconds is added.

 - No

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/util/CarbonProperties.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/CarbonProperties.java
@@ -2086,6 +2086,27 @@ public final class CarbonProperties {
     return Boolean.parseBoolean(configuredValue);
   }
 
+  public long getMetaCacheExpirationTime() {
+    String configuredValue = CarbonProperties.getInstance()
+        .getProperty(CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS);
+    if (configuredValue == null || configuredValue.equalsIgnoreCase("0")) {
+      return CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS_DEFAULT;
+    }
+    try {
+      long expirationTime = Long.parseLong(configuredValue);
+      LOGGER.info("Value for "
+          + CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS + " is "
+          + expirationTime + ".");
+      return expirationTime;
+    } catch (NumberFormatException e) {
+      LOGGER.warn(configuredValue + " is not a valid input for "
+          + CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS + ", taking "
+          + CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS_DEFAULT
+          + " as default value.");
+      return CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS_DEFAULT;
+    }
+  }
+
   public boolean isSIRepairEnabled(String dbName, String tableName) {
     // Check if user has enabled/disabled the use of property for the current db and table using
     // the set command
--- a/docs/configuration-parameters.md
+++ b/docs/configuration-parameters.md
@@ -149,6 +149,7 @@ This section provides the details of all the configurations required for the Car
 | carbon.max.pagination.lru.cache.size.in.mb | -1 | Maximum memory **(in MB)** upto which the SDK pagination reader can cache the blocklet rows. Suggest to configure as multiple of blocklet size. Default value of -1 means there is no memory limit for caching. Only integer values greater than 0 are accepted. |
 | carbon.partition.max.driver.lru.cache.size | -1 | Maximum memory **(in MB)** upto which driver can cache partition metadata. Beyond this, least recently used data will be removed from cache before loading new set of values.
 | carbon.mapOrderPushDown.<db_name>_<table_name>.column| empty | If order by column is in sort column, specify that sort column here to avoid ordering at map task . |
+| carbon.metacache.expiration.seconds | Long.MAX_VALUE | Expiration time **(in seconds)** for tableInfo cache in CarbonMetadata and tableModifiedTime in CarbonFileMetastore, after the time configured since last access to the cache entry, tableInfo and tableModifiedTime will be removed from each cache. Recent access will refresh the timer. Default value of Long.MAX_VALUE means the cache will not be expired by time. **NOTE:** At the time when cache is being expired, queries on the table may fail with NullPointerException. |
 
 ## Data Mutation Configuration
 | Parameter | Default Value | Description |
--- a/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/apache/carbondata/core/constants/CarbonCommonConstants.java
@@ -2557,6 +2557,20 @@ public final class CarbonCommonConstants {
           " characters. Please consider long string data type.";
 
   /**
+   * Expiration time for tableInfo cache in CarbonMetadata, after the time configured
+   * since last access to the cache entry, tableInfo will be removed from cache. Recent
+   * access will refresh the timer. At the time when cache is being expired, queries on
+   * the table may fail with NullPointerException.
+   */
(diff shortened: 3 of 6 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5196/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3436/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3290/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5048/

jack86596:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5216/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3456/

kunal642:
LGTM

docs/configuration-parameters.md
```diff
@@ -149,6 +149,7 @@ This section provides the details of all the configurations required for the Car
 | carbon.max.pagination.lru.cache.size.in.mb | -1 | Maximum memory **(in MB)** upto which the SDK pagination reader can cache the blocklet rows. Suggest to configure as multiple of blocklet size. Default value of -1 means there is no memory limit for caching. Only integer values greater than 0 are accepted. |
 | carbon.partition.max.driver.lru.cache.size | -1 | Maximum memory **(in MB)** upto which driver can cache partition metadata. Beyond this, least recently used data will be removed from cache before loading new set of values.
 | carbon.mapOrderPushDown.<db_name>_<table_name>.column| empty | If order by column is in sort column, specify that sort column here to avoid ordering at map task . |
+| carbon.metacache.expiration.seconds | Long.MAX_VALUE | Expiration time **(in seconds)** for tableInfo cache in CarbonMetadata, after the time configured since last access to the cache entry, tableInfo will be removed from cache. Recent access will refresh the timer. Default value of Long.MAX_VALUE means the cache will not be expired by time. **NOTE:** At the time when cache is being expired, queries on the table may fail with NullPointerException. |
```
Indhumathi27, line 152: it is not only for CarbonMetadata, this property is also used for tableModifiedTimeStore. Please update

core/src/main/java/org/apache/carbondata/core/util/CarbonProperties.java
```diff
@@ -2086,6 +2086,27 @@ public boolean isSetLenientEnabled() {
...
+      return CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS_DEFAULT;
+    }
+    try {
+      long expirationTime = Long.parseLong(configuredValue);
+      LOGGER.info("Value for "
+          + CarbonCommonConstants.CARBON_METACACHE_EXPIRATION_TIME_IN_SECONDS + " is "
+          + expirationTime + ".");
+      return expirationTime;
+    } catch (NumberFormatException e) {
+      LOGGER.info(configuredValue + " is not a valid input for "
```
Indhumathi27, line 2102: could be a warning log

jack86596: changed to warning log.

docs/configuration-parameters.md
jack86596: added.

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5233/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3472/

docs/configuration-parameters.md
```diff
@@ -149,6 +149,7 @@ This section provides the details of all the configurations required for the Car
...
 | carbon.partition.max.driver.lru.cache.size | -1 | Maximum memory **(in MB)** upto which driver can cache partition metadata. Beyond this, least recently used data will be removed from cache before loading new set of values.
 | carbon.mapOrderPushDown.<db_name>_<table_name>.column| empty | If order by column is in sort column, specify that sort column here to avoid ordering at map task . |
+| carbon.metacache.expiration.seconds | Long.MAX_VALUE | Expiration time **(in seconds)** for tableInfo cache in CarbonMetadata and tableModifiedTime in CarbonFileMetastore, after the time configured since last access to the cache entry, tableInfo and tableModifiedTime will be removed from each cache. Recent access will refresh the timer. Default value of Long.MAX_VALUE means the cache will not be expired by time. **NOTE:** At the time when cache is being expired, queries on the table may fail with NullPointerException. |
```
kunal642, line 152: During your test did you face NullPointerException as mentioned in this doc?

jack86596: Yes, if we set carbon.metacache.expiration.seconds to very small like 1s, UT will fail with NullPointerException.

kunal642: then its a problem, because queries would fail randomly.

@akashrn5 @ajantha-bhat what do you suggest?

ajantha-bhat: In one application itself if we do drop table and query concurrently what is the behavior ? 
a) Giving NPE is not good idea. If the other application has OOM, user can call drop metacache ?
b) Also for multiple application scenario, user should use index server for caching, which will avoid these problems as only one cache will be present for all the application.

akashrn5: agree with @kunal642 , if randomly it fails then it will give false info to user and everytime checking whether its because of this time expiration or some other genuine NullPointer errors

jack86596: I think this feature is just for those customers who care about the memory leak. So if the customer doesn't want memory leak, he can choose a proper expired time which at the time of cache expired, the table will never be queried. For example, customer will have a new table every day, and this table will be queried only at that day. So he can choose 1 week as the property value so that this table cache will only leak for one week and after one week will be cleared.
@ajantha-bhat the memory leak is not about the index cache, it is table metadata cache.

akashrn5: > c) or may be we should enhance drop table to clean these entries if exists when called from application 2 even though table doesn't exist but stale metadata exist in memory.

there is one method called `removeStaleTimeStampEntries` which removes invalid timestamp entries in memory, but this will be called in drop table command, and in other application until and unless user calls any drop table (basically any valid drop table), else it wont be cleared

jack86596: @kunal642 https://issues.apache.org/jira/browse/CARBONDATA-4098 raised one jira to tack the null pointer issue.

kunal642: ok, ill merge this as the default behavior is not to remove cache.

kunal642: please fix the NPE in a separate PR

kunal642:
LGTM

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5242/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3481/

PR #4058, opened 2020-12-18
[CARBONDATA-4084]Fixed data corruption issue after fallback of Local dictionary

kumarvishal09:
### Why is this PR needed?
 Data is getting corrupted when fallback happens for local dictionary encoded page.
 
In LVByteBufferColumnPage Length getting added inside put method. In case of fallback when reverse lookup is done and encoded page is replaced with actual value[Dictionary values]. As actual Dictionary value is already present in LV format again one more length is getting added[LLV], so during query it's showing extra character for those columns
    
 - No

 - No


CarbonDataQA2:
test it - ignore it

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5208/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3448/

ajantha-bhat:
LGTM. 


As the actual value is already LV, putBytes() no need to append L again, but that change is more, as many classes need to modified.
Current change of removing L from actual value and appending L again is fine as it is (corner) fallback scenario.

PR #4059, opened 2020-12-18
[CARBONDATA-4092] Fix concurrent issues in delete segment API's and MV flow

vikramahuja1001:
### Why is this PR needed?
 They are multiple issues with the Delete segment API:
1. Not using the latest loadmetadatadetails while writing to table status file, thus can remove table status entry of any concurrently loaded Insert In progress/success segment.
2. The code reads the table status file 2 times 
 
When in concurrent queries, they both access checkAndReloadSchema for MV on all databases, 2 different queries try to create a file on same location, HDFS takes the lock for one and fails for another, thus failing the query
1. Only reading the table status file once.
2. Using the latest tablestatus to mark the segment Marked for delete, thus no concurrent issues will come
3.  Made touchMDT and checkAndReloadSchema methods syncronized, so that only instance can access it at one time.
    
 - No

 - No

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatusManager.java
+++ b/core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatusManager.java
@@ -427,6 +427,8 @@ public class SegmentStatusManager {
   public static List<String> updateDeletionStatus(AbsoluteTableIdentifier identifier,
       List<String> loadIds, String tableFolderPath) throws Exception {
     CarbonTableIdentifier carbonTableIdentifier = identifier.getCarbonTableIdentifier();
+    ICarbonLock carbonCleanFilesLock =
+        CarbonLockFactory.getCarbonLockObj(identifier, LockUsage.CLEAN_FILES_LOCK);
     ICarbonLock carbonDeleteSegmentLock =
         CarbonLockFactory.getCarbonLockObj(identifier, LockUsage.DELETE_SEGMENT_LOCK);
     ICarbonLock carbonTableStatusLock =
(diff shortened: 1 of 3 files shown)


CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3449/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5209/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3451/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5211/

vikramahuja1001:
@ajantha-bhat @akashrn5 , please review

core/src/main/java/org/apache/carbondata/core/view/MVProvider.java
```diff
@@ -540,7 +540,7 @@ private void checkAndReloadSchemas(MVManager viewManager, boolean touchFile)
       }
     }
 
-    private void touchMDTFile() throws IOException {
+    private synchronized void touchMDTFile() throws IOException {
```
ajantha-bhat, line 543: In saveSchema() method of same file, 
checkAndReloadSchemas(viewManager, true);
        touchMDTFile();

line 432: after passing touch = true, again we are calling touchMDTFile, is itrequired?

ajantha-bhat: @Indhumathi27 , @vikramahuja1001

Indhumathi27: @ajantha-bhat yes, it is required. On second MV creation, have to update last modified time of schema INdex file

core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatusManager.java
```diff
@@ -445,38 +445,28 @@ private static Integer compareDateValues(Long loadValue, Long userValue) {
...
-              // To handle concurrency scenarios, always take latest metadata before writing
-              // into status file.
-              LoadMetadataDetails[] latestLoadMetadataDetails = readLoadMetadata(tableFolderPath);
-              updateLatestTableStatusDetails(listOfLoadFolderDetailsArray,
-                  latestLoadMetadataDetails);
+        if (carbonTableStatusLock.lockWithRetries()) {
+          LOG.info("Table status lock has been successfully acquired.");
+          listOfLoadFolderDetailsArray = readLoadMetadata(tableFolderPath);
+          if (listOfLoadFolderDetailsArray.length != 0) {
+            updateDeletionStatus(identifier, loadIds, listOfLoadFolderDetailsArray, invalidLoadIds);
```
ajantha-bhat, line 452: If the table status has 200K segments, finding and marking the matching segments to delete will take lot of time. Holding table status lock for quite a long time will impact concurrent query running. That is why previously they didn't had this logic in the lock. 

If some user reports issue because of this PR change, we have to optimize this logic may be.

core/src/main/java/org/apache/carbondata/core/view/MVProvider.java
ajantha-bhat: checkAndReloadSchemas, better to return a flag whether it has touched MDT or not, if false then only we can call touchMDT

vikramahuja1001: done

core/src/main/java/org/apache/carbondata/core/statusmanager/SegmentStatusManager.java
vikramahuja1001: changed logic

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3460/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5221/

ajantha-bhat:
LGTM

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5224/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3463/

PR #4063, opened 2020-12-22
[CARBONDATA-4095] Fix Select Query with SI filter fails, when columnDrift is Set

Indhumathi27:
### Why is this PR needed?
After converting expression to IN Expression for maintable with SI, expression is not processed if  ColumnDrift is enabled. Query fails with NPE during resolveFilter. Exception is added in JIRA
 
Process the filter expression after adding implicit expression
    
 - No

 - Yes

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
@@ -52,7 +52,6 @@ import org.apache.carbondata.core.metadata.schema.table.TableInfo;
 import org.apache.carbondata.core.metadata.schema.table.column.ColumnSchema;
 import org.apache.carbondata.core.profiler.ExplainCollector;
 import org.apache.carbondata.core.readcommitter.ReadCommittedScope;
-import org.apache.carbondata.core.scan.expression.Expression;
 import org.apache.carbondata.core.scan.filter.FilterUtil;
 import org.apache.carbondata.core.scan.filter.resolver.FilterResolverIntf;
 import org.apache.carbondata.core.scan.model.QueryModel;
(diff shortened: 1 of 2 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5230/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3469/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonInputFormat.java
```diff
@@ -683,6 +683,8 @@ public QueryModel createQueryModel(InputSplit inputSplit, TaskAttemptContext tas
     }
     if (indexFilter != null) {
       checkAndAddImplicitExpression(indexFilter.getExpression(), inputSplit);
+      // process filter expression after adding implicit expression
+      indexFilter.processFilterExpression();
```
akashrn5, line 687: in `Querymodelbuilder.build()` we call `processFilterExpression` with check of columndrift, so here if we call process, then in case of no column drift , two times will be called, so please refactor to call this only in case of columndrift true

Indhumathi27: done

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3475/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5236/

akashrn5:
LGTM

PR #4075, opened 2021-01-11
[CARBONDATA-4105] Select * query fails after a SDK written segment is added by alter table add segment query.

Karan980:
### Why is this PR needed?
When a SDK written segment on which read is already performed once, is added through alter table add segment query to a carbon table, then select * query fails after adding it.

In SDK segments the segmentId is the timestamp of the segment. When the SDK segment is read before adding, its indexes are stored in cache. Cache is a map of indexFilePath to Indexes. Now when the same segment is added to carbon table its segment ID is no longer the timestamp but the indexFilePath remains same as it is added externally. Now when we run select * query we get the indexes from the cache, but it is unable to map it to segment, because segment id changes.
 
Also added segment Id to the key of cache map to make it more unique.

-No

 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/indexstore/TableBlockIndexUniqueIdentifier.java
+++ b/core/src/main/java/org/apache/carbondata/core/indexstore/TableBlockIndexUniqueIdentifier.java
@@ -45,7 +45,8 @@ public class TableBlockIndexUniqueIdentifier implements Serializable {
     this.indexFileName = indexFileName;
     this.mergeIndexFileName = mergeIndexFileName;
     this.segmentId = segmentId;
-    this.uniqueName = indexFilePath + CarbonCommonConstants.FILE_SEPARATOR + indexFileName;
+    this.uniqueName = segmentId + CarbonCommonConstants.UNDERSCORE +
+        indexFilePath + CarbonCommonConstants.FILE_SEPARATOR + indexFileName;
   }
 
   public TableBlockIndexUniqueIdentifier(String segmentId) {
--- a/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
+++ b/integration/spark/src/test/scala/org/apache/carbondata/spark/testsuite/addsegment/AddSegmentTestCase.scala
@@ -16,7 +16,7 @@
  */
 package org.apache.carbondata.spark.testsuite.addsegment
 
-import java.io.File
+import java.io.{File, FilenameFilter}
 
 import scala.io.Source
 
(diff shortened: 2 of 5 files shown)


CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5296/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3536/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5298/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3538/

Karan980:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5302/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3542/

QiangCai:
do you have the reproduce steps?

Karan980:
> do you have the reproduce steps?

Yes, run this test case which i have added in this PR without any other changes from the PR. (Test add segment by carbon written by SDK on which read is already performed)

QiangCai:
please normalize the title, insert a blank after the close middle bracket ']'

Karan980:
> please normalize the title, insert a blank after the close middle bracket ']'

Done

core/src/main/java/org/apache/carbondata/core/indexstore/TableBlockIndexUniqueIdentifier.java
```diff
@@ -45,7 +45,8 @@ public TableBlockIndexUniqueIdentifier(String indexFilePath, String indexFileNam
     this.indexFileName = indexFileName;
     this.mergeIndexFileName = mergeIndexFileName;
     this.segmentId = segmentId;
-    this.uniqueName = indexFilePath + CarbonCommonConstants.FILE_SEPARATOR + indexFileName;
+    this.uniqueName = segmentId + CarbonCommonConstants.UNDERSCORE +
+        indexFilePath + CarbonCommonConstants.FILE_SEPARATOR + indexFileName;
```
ajantha-bhat, line 49: indexFileName already has the segment id, 

SDK and spark are never meant to use in same JVM, Both are new process right ? why SDK query is going into cabron table LRU cache ?

ajantha-bhat:
@Karan980 : Here in a spark clsuter jvm itself you are running sdk read and write. Hence this problem.
Ideally when you run SDK in cluster it will span up it's own JVM. so spark cluster will not be affected. 
I don't think this issue has to be fixed as it is not a user scenario.

Karan980:
ok, closing this issue as cache will not be shared if both are different JVM processes.

PR #4099, opened 2021-03-02
[CARBONDATA-4141] Index Server is not caching indexes for external tables with sdk segments.

Karan980:
### Why is this PR needed?
Indexes cached in Executor cache are not dropped when drop table is called for external table with SDK segments. Because, external tables with sdk segments will not have metadata like table status file. So in drop table command we send zero segments to indexServer clearIndexes job, which clears nothing from executor side. So when we drop this type of table, executor side indexes are not dropped. Now when we again create external table with same location and do select * or select count(*), it will not cache the indexes for this table, because indexes with same loaction are already present. Now show metacache on this newly created table will use new tableId , but indexes present have the old tableId, whose table is already dropped. So show metacache will return nothing, because of tableId mismatch.
 
Prepared the validSegments from indexFiles present at external table location and send it to IndexServer clearIndexes job through IndexInputFormat.
    
 - No

 - No

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
+++ b/core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
@@ -20,6 +20,7 @@ package org.apache.carbondata.core.index;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -36,7 +37,10 @@ import org.apache.carbondata.core.indexstore.PartitionSpec;
 import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier;
 import org.apache.carbondata.core.metadata.schema.table.CarbonTable;
 import org.apache.carbondata.core.metadata.schema.table.RelationIdentifier;
+import org.apache.carbondata.core.readcommitter.LatestFilesReadCommittedScope;
+import org.apache.carbondata.core.readcommitter.ReadCommittedScope;
 import org.apache.carbondata.core.scan.filter.resolver.FilterResolverIntf;
+import org.apache.carbondata.core.statusmanager.LoadMetadataDetails;
 import org.apache.carbondata.core.statusmanager.SegmentStatusManager;
 import org.apache.carbondata.core.util.BlockletIndexUtil;
 import org.apache.carbondata.core.util.CarbonProperties;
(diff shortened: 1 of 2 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5106/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3347/

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
```diff
@@ -112,15 +116,31 @@ public static IndexJob getIndexJob(Configuration configuration) throws IOExcepti
    */
   private static void executeClearIndexJob(IndexJob indexJob,
       CarbonTable carbonTable, String indexToClear) throws IOException {
-    SegmentStatusManager.ValidAndInvalidSegmentsInfo validAndInvalidSegmentsInfo =
-            getValidAndInvalidSegments(carbonTable, FileFactory.getConfiguration());
-    List<String> invalidSegment = new ArrayList<>();
-    for (Segment segment : validAndInvalidSegmentsInfo.getInvalidSegments()) {
-      invalidSegment.add(segment.getSegmentNo());
+    IndexInputFormat indexInputFormat;
+    if (!carbonTable.isTransactionalTable()) {
```
kunal642, line 120: Please reuse the LatestFilesReadCommittedScope class to take a snapshot of the non-transactional table.

Karan980: Done

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5113/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3354/

vikramahuja1001:
please correct the splits path of getSplits in case of External Tables in DistributedPruneRDD as well

Karan980:
> please correct the splits path of getSplits in case of External Tables in DistributedPruneRDD as well

Done

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
```diff
@@ -112,15 +115,32 @@ public static IndexJob getIndexJob(Configuration configuration) throws IOExcepti
...
+        listOfValidSegments.add(seg);
+      }
+      indexInputFormat =
+          new IndexInputFormat(carbonTable, listOfValidSegments, new ArrayList<>(0), true,
+              indexToClear);
+    } else {
+      SegmentStatusManager.ValidAndInvalidSegmentsInfo validAndInvalidSegmentsInfo =
+          getValidAndInvalidSegments(carbonTable, FileFactory.getConfiguration());
+      List<String> invalidSegment = new ArrayList<>();
+      for (Segment segment : validAndInvalidSegmentsInfo.getInvalidSegments()) {
```
vikramahuja1001, line 137: Use Java stream here instead of a for loop

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
```diff
@@ -112,15 +115,32 @@ public static IndexJob getIndexJob(Configuration configuration) throws IOExcepti
...
-    List<String> invalidSegment = new ArrayList<>();
-    for (Segment segment : validAndInvalidSegmentsInfo.getInvalidSegments()) {
-      invalidSegment.add(segment.getSegmentNo());
+    IndexInputFormat indexInputFormat;
+    if (!carbonTable.isTransactionalTable()) {
+      ReadCommittedScope readCommittedScope =
+          new LatestFilesReadCommittedScope(carbonTable.getTablePath(),
+              FileFactory.getConfiguration());
+      LoadMetadataDetails[] loadMetadataDetails = readCommittedScope.getSegmentList();
+      List<Segment> listOfValidSegments = new ArrayList<>();
```
vikramahuja1001, line 124: give arrayList limit equals to loadMetadataDetails size, because it will never increase that size

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
```diff
@@ -112,15 +115,32 @@ public static IndexJob getIndexJob(Configuration configuration) throws IOExcepti
...
+      for (LoadMetadataDetails segment : loadMetadataDetails) {
```
vikramahuja1001, line 125: try using java stream api here instead of a for loop wherever possible

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
Karan980: Done

core/src/main/java/org/apache/carbondata/core/index/IndexUtil.java
Karan980: Done

kunal642:
LGTM

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12545/job/ApacheCarbon_PR_Builder_2.4.5/3363/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12545/job/ApacheCarbonPRBuilder2.3/5123/

PR #4105, opened 2021-03-12
[CARBONDATA-4148] Reindex failed when SI has stale carbonindexmerge file

jack86596:
### Why is this PR needed?
Reindex failed when SI has stale carbonindexmerge file, throw exception FileNotFoundException. This is because  SegmentFileStore.getIndexFiles stores the mapping of indexfile to indexmergefile, when stale carbon indexmergefile exists, indexmergefile will not be null. During merging index file, new indexmergefile will be created with same name as before in the same location. At the end of CarbonIndexFileMergeWriter.writeMergeIndexFileBasedOnSegmentFile, carbon index file will be deleted. Since indexmergefile is stored in the indexFiles list, newly created indexmergefile will be delete also, which leads to FileNotFoundException.
 
1. SegmentFileStore.getIndexFiles stores the mapping of indexfile to indexmergefile which is redundant.
2. SegmentFileStore.getIndexOrMergeFiles returns both index file and index merge file, so function name is incorrect, rename to getIndexAndMergeFiles.
3. CarbonLoaderUtil.getActiveExecutor actually get active node, so function name is incorrect, rename to getActiveNode, together replace all "executor" with "node" in function assignBlocksByDataLocality.
    
 - No

 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
+++ b/core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
@@ -831,8 +831,8 @@ public class SegmentFileStore {
    * Gets all index files from this segment
    * @return
    */
-  public Map<String, String> getIndexFiles() {
-    Map<String, String> indexFiles = new HashMap<>();
+  public Set<String> getIndexFiles() {
+    Set<String> indexFiles = new HashSet<>();
     if (segmentFile != null) {
       for (Map.Entry<String, FolderDetails> entry : getLocationMap().entrySet()) {
         String location = entry.getKey();
@@ -841,8 +841,7 @@ public class SegmentFileStore {
         }
         if (entry.getValue().status.equals(SegmentStatus.SUCCESS.getMessage())) {
           for (String indexFile : entry.getValue().getFiles()) {
-            indexFiles.put(location + CarbonCommonConstants.FILE_SEPARATOR + indexFile,
-                entry.getValue().mergeFileName);
+            indexFiles.add(location + CarbonCommonConstants.FILE_SEPARATOR + indexFile);
           }
         }
       }
@@ -854,7 +853,7 @@ public class SegmentFileStore {
    * Gets all index files from this segment
    * @return
    */
-  public Map<String, String> getIndexOrMergeFiles() throws IOException {
+  public Map<String, String> getIndexAndMergeFiles() throws IOException {
     Map<String, String> indexFiles = new HashMap<>();
     if (segmentFile != null) {
       for (Map.Entry<String, FolderDetails> entry : getLocationMap().entrySet()) {
@@ -898,17 +897,9 @@ public class SegmentFileStore {
    * @return
    */
   public List<CarbonFile> getIndexCarbonFiles() {
-    Map<String, String> indexFiles = getIndexFiles();
-    Set<String> files = new HashSet<>();
-    for (Map.Entry<String, String> entry: indexFiles.entrySet()) {
-      Path path = new Path(entry.getKey());
-      files.add(entry.getKey());
-      if (entry.getValue() != null) {
-        files.add(new Path(path.getParent(), entry.getValue()).toString());
-      }
-    }
+    Set<String> indexFiles = getIndexFiles();
     List<CarbonFile> carbonFiles = new ArrayList<>();
-    for (String indexFile : files) {
+    for (String indexFile : indexFiles) {
       CarbonFile carbonFile = FileFactory.getCarbonFile(indexFile);
       if (carbonFile.exists()) {
         carbonFiles.add(carbonFile);
@@ -1348,7 +1339,7 @@ public class SegmentFileStore {
     } else {
       SegmentFileStore segmentFileStore =
           new SegmentFileStore(tablePath, segment.getSegmentFileName());
-      indexFiles = segmentFileStore.getIndexOrMergeFiles().keySet();
+      indexFiles = segmentFileStore.getIndexAndMergeFiles().keySet();
     }
     return indexFiles;
   }
--- a/index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
+++ b/index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
@@ -119,6 +119,19 @@ class TestIndexRepair extends QueryTest with BeforeAndAfterAll {
     sql("drop table if exists maintable")
   }
 
+  test("reindex command with stale files") {
+    sql("drop table if exists maintable")
+    sql("CREATE TABLE maintable(a INT, b STRING, c STRING) stored as carbondata")
+    sql("CREATE INDEX indextable1 on table maintable(c) as 'carbondata'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("DELETE FROM TABLE INDEXTABLE1 WHERE SEGMENT.ID IN(0,1,2)")
+    sql("REINDEX INDEX TABLE indextable1 ON MAINTABLE WHERE SEGMENT.ID IN (0,1)")
+    assert(sql("select * from maintable where c = 'string2'").count() == 2)
+    sql("drop table if exists maintable")
+  }
+
   test("insert command after deleting segments from SI table") {
     sql("drop table if exists maintable")
     sql("CREATE TABLE maintable(a INT, b STRING, c STRING) stored as carbondata")
--- a/core/src/main/java/org/apache/carbondata/core/readcommitter/TableStatusReadCommittedScope.java
+++ b/core/src/main/java/org/apache/carbondata/core/readcommitter/TableStatusReadCommittedScope.java
@@ -86,7 +86,7 @@ public class TableStatusReadCommittedScope implements ReadCommittedScope {
     } else {
       SegmentFileStore fileStore =
           new SegmentFileStore(identifier.getTablePath(), segment.getSegmentFileName());
-      indexFiles = fileStore.getIndexOrMergeFiles();
+      indexFiles = fileStore.getIndexAndMergeFiles();
       if (fileStore.getSegmentFile() != null) {
         segment.setSegmentMetaDataInfo(fileStore.getSegmentFile().getSegmentMetaDataInfo());
       }
--- a/core/src/main/java/org/apache/carbondata/core/util/SessionParams.java
+++ b/core/src/main/java/org/apache/carbondata/core/util/SessionParams.java
@@ -24,7 +24,6 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.carbondata.common.constants.LoggerAction;
 import org.apache.carbondata.common.logging.LogServiceFactory;
-import org.apache.carbondata.core.cache.CacheProvider;
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
 import org.apache.carbondata.core.constants.CarbonLoadOptionConstants;
 import org.apache.carbondata.core.exception.InvalidConfigurationException;
@@ -49,7 +48,7 @@ import org.apache.log4j.Logger;
 public class SessionParams implements Serializable, Cloneable {
 
   private static final Logger LOGGER =
-      LogServiceFactory.getLogService(CacheProvider.class.getName());
+      LogServiceFactory.getLogService(SessionParams.class.getName());
   private static final long serialVersionUID = -7801994600594915264L;
 
   private Map<String, String> sProps;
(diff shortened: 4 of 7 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5562/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3796/

vikramahuja1001:
@jack86596 , won't running clean files before doing the reindex operation solve this issue?

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
```diff
@@ -119,6 +119,19 @@ class TestIndexRepair extends QueryTest with BeforeAndAfterAll {
...
 
+  test("reindex command with stale files") {
+    sql("drop table if exists maintable")
+    sql("CREATE TABLE maintable(a INT, b STRING, c STRING) stored as carbondata")
+    sql("CREATE INDEX indextable1 on table maintable(c) as 'carbondata'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("INSERT INTO maintable SELECT 1,'string1', 'string2'")
+    sql("DELETE FROM TABLE INDEXTABLE1 WHERE SEGMENT.ID IN(0,1,2)")
+    sql("REINDEX INDEX TABLE indextable1 ON MAINTABLE WHERE SEGMENT.ID IN (0,1)")
```
vikramahuja1001, line 130: technically speaking reindex should not be allowed at this point, because the segment still exists(even if it is MFD), the purpose of reindex command is to reload missing segments which is not the case here. Also, will this same behaviour also happen for other segment status, say success or partially success? @akashrn5 can give some input maybe

jack86596:
> @jack86596 , won't running clean files before doing the reindex operation solve this issue?

why before reindex operation, need to run clean files? and dependency relation between reindex and clean files?

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
jack86596: So please provide a better solution to solve this issue: main table segment (success), SI table segment (marked for delete). The solution should not be "run clean files for SI table first", because it is not better than current one.

vikramahuja1001:
> > @jack86596 , won't running clean files before doing the reindex operation solve this issue?
> 
> why before reindex operation, need to run clean files? any dependency relation between reindex and clean files?

No, there is no need to do clean files before reindex there is no dependency, but we only do reindex when the segments are missing, it is not the case here though. What i meant was that running clean files on si table would solve your problem

jack86596:
> > > @jack86596 , won't running clean files before doing the reindex operation solve this issue?
> > 
> > 
> > why before reindex operation, need to run clean files? any dependency relation between reindex and clean files?
> 
> No, there is no need to do clean files before reindex there is no dependency, but we only do reindex when the segments are missing, it is not the case here though. What i meant was that running clean files on si table would solve your problem

why I need to run one extra command to solve the problem, why not just solve the problem directly, in this way, no need to run extra command? If in order to solve the problem, you need to run clean files, then this is dependency.

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
Indhumathi27: i think, at first, delete segment directly from indexTable should not be allowed. For second case, where main table contains SUCCESS segments and SI has MFD, during reindex, before loading data again, we can delete old files in that segment

jack86596: 1. "i think, at first, delete segment directly from indexTable should not be allowed." This is unit test, we are simulating the issue(main table semgent success and SI table segment MFD), for real user, they don't need to manually delete SI segment in order to get the issue, they will face this issue directly in their daily running without any manuallly operation.
2. "For second case, where main table contains SUCCESS segments and SI has MFD, during reindex, before loading data again, we can delete old stale files in that segment" Yes, you provide a solution and i agree this will work, but again, the solution you provide is not as good as current one because current one you don't need to delete the stale files, reindex can also work.

Indhumathi27: 1. User can face this scenario, what i meant is, allowing delete directly on index Table is also a new issue, when user does it by mistake,  will make main table and SI not in sync. Then, user unneesarily has to perform Reindex. So, you can handle this scenario in this PR.
2. If you are gonna return only index Files, always merge file name will be null. So, you can use return List of index Files alone instead of Map from getIndexFiles method.

jack86596: 1. User will not delete SI segment directly, the issue (main table segment success and SI segment MFD) is because carbon has so many bugs which will lead to this situation even user only just do the loading and compaction. And actually in production, we (maintainers) need to "delete directly on index table" just because we need to recovery carbon table from broken state to normal state, then later trigger reindex to repair the broken SI segment. So until "all the bugs which will lead to main table and SI table not in syc" are solved, we cannot block this.
2. This "return List of index Files alone" if you say so, I can change it.

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
```diff
@@ -841,8 +841,7 @@ public SegmentFile getSegmentFile() {
         }
         if (entry.getValue().status.equals(SegmentStatus.SUCCESS.getMessage())) {
           for (String indexFile : entry.getValue().getFiles()) {
-            indexFiles.put(location + CarbonCommonConstants.FILE_SEPARATOR + indexFile,
-                entry.getValue().mergeFileName);
+            indexFiles.put(location + CarbonCommonConstants.FILE_SEPARATOR + indexFile, null);
```
VenuReddy2103, line 844: Its caller `SegmentFileStore.getIndexCarbonFiles()`has code to handle for non null value to add merge index file also to index file list. It becomes dead code now. You would want to remove that too.

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
jack86596: 1. For example, if both main table and SI table segment status are success, but one of the carbondata file or index file in SI segment is missing or broken, how you fix this if you cannot manually delete this SI segment?

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
jack86596: This make sense, will do it.

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
```diff
@@ -854,7 +853,7 @@ public SegmentFile getSegmentFile() {
    * Gets all index files from this segment
    * @return
    */
-  public Map<String, String> getIndexOrMergeFiles() throws IOException {
+  public Map<String, String> getIndexAndMergeFiles() throws IOException {
```
VenuReddy2103, line 856: I still think getIndexOrMergeFiles name makes sense. Returned map can have either mergeindex or index. Curious to know when it does have both merge index and index in map ?

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
Indhumathi27: ok

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
jack86596: backward compatibility, carbon 1.3 both index and merge file exist.

jack86596: Or means either, not both, but sometime it will return both index file and merge file, so Or is not correct.

jack86596: Current implementation:
if (null != mergeFileName) {
  add merge file
}
if (null != index file && not empty) {
  add index file
}

index/secondary-index/src/test/scala/org/apache/carbondata/spark/testsuite/secondaryindex/TestIndexRepair.scala
VenuReddy2103: Agree with @Indhumathi27. IMHO, we shouldn't allow delete segments on index table itself. And during repair index, if have segment with partial data, we should delete the segment completely(segment folder, segment file, tablestatus entry for the segment) before proceeding with segment repair.

jack86596: 1. "we shouldn't allow delete segments on index table itself." please refer to the second last comment right before your comment.
2. "And during repair index, if have segment with partial data, we should delete the segment completely(segment folder, segment file, probably tablestatus entry for the segment as well) before proceeding with segment repair." You suggestion of course is right but too more complicate comparing to existing implementation, so please first do the complete analysis and design and then we can discuss and plan next step.

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
jack86596: Done.

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5571/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3805/

jack86596:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5572/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3806/

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
```diff
@@ -898,15 +897,8 @@ public SegmentFile getSegmentFile() {
...
-    Map<String, String> indexFiles = getIndexFiles();
-    Set<String> files = new HashSet<>();
-    for (Map.Entry<String, String> entry: indexFiles.entrySet()) {
-      Path path = new Path(entry.getKey());
-      files.add(entry.getKey());
-      if (entry.getValue() != null) {
-        files.add(new Path(path.getParent(), entry.getValue()).toString());
-      }
-    }
+    List<String> indexFiles = getIndexFiles();
```
Indhumathi27, line 900: Can you return Set<String> from getIndexFiles itself

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5576/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3810/

core/src/main/java/org/apache/carbondata/core/metadata/SegmentFileStore.java
jack86596: done

jack86596:
retest this please

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3813/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5579/

Indhumathi27:
LGTM

Indhumathi27:
retest this please

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3831/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5597/

jack86596:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12444/job/ApacheCarbonPRBuilder2.3/5601/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12444/job/ApacheCarbon_PR_Builder_2.4.5/3835/

PR #4124, opened 2021-04-22
[CARBONDATA-4173][CARBONDATA-4174] Fix inverted index query issue and handle exception for desc column

ShreelekhyaG:
### Why is this PR needed?
1. After creating an Inverted index on the dimension column, some of the filter queries give incorrect results.
2. handle exception for higher level non-existing children column in desc column.
 
1. While sorting byte arrays with inverted index, we use `compareTo `method of `ByteArrayColumnWithRowId`.  Here, it was sorting based on the last byte only. Made changes to sort properly based on the entire byte length when dictionary is used.
2. handled exception and added in testcase.
    
 - No

 - Yes

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayColumnWithRowId.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayColumnWithRowId.java
@@ -26,9 +26,12 @@ public class ByteArrayColumnWithRowId implements Comparable<ByteArrayColumnWithR
 
   private short rowId;
 
-  ByteArrayColumnWithRowId(byte[] column, short rowId) {
+  private boolean isNoDictionary;
+
+  ByteArrayColumnWithRowId(byte[] column, short rowId, boolean isNoDictionary) {
     this.column = column;
     this.rowId = rowId;
+    this.isNoDictionary = isNoDictionary;
   }
 
   public byte[] getColumn() {
@@ -41,8 +44,13 @@ public class ByteArrayColumnWithRowId implements Comparable<ByteArrayColumnWithR
 
   @Override
   public int compareTo(ByteArrayColumnWithRowId o) {
-    return UnsafeComparer.INSTANCE
-        .compareTo(column, 2, column.length - 2, o.column, 2, o.column.length - 2);
+    if (isNoDictionary) {
+      return UnsafeComparer.INSTANCE
+          .compareTo(column, 2, column.length - 2, o.column, 2, o.column.length - 2);
+    } else {
+      return UnsafeComparer.INSTANCE
+          .compareTo(column, 0, column.length, o.column, 0, o.column.length);
+    }
   }
 
   @Override
--- a/core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayBlockIndexerStorage.java
+++ b/core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayBlockIndexerStorage.java
@@ -47,14 +47,8 @@ public class ByteArrayBlockIndexerStorage extends BlockIndexerStorage<byte[][]> 
   private ByteArrayColumnWithRowId[] createColumnWithRowId(byte[][] dataPage,
       boolean isNoDictionary) {
     ByteArrayColumnWithRowId[] columnWithIndexes = new ByteArrayColumnWithRowId[dataPage.length];
-    if (isNoDictionary) {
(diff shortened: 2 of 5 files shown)


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5235/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3487/

core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayColumnWithRowId.java
```diff
@@ -42,7 +42,7 @@ public short getRowId() {
   @Override
   public int compareTo(ByteArrayColumnWithRowId o) {
     return UnsafeComparer.INSTANCE
-        .compareTo(column, 2, column.length - 2, o.column, 2, o.column.length - 2);
+        .compareTo(column, 0, column.length, o.column, 0, o.column.length);
```
ajantha-bhat, line 45: I can see that they were ignoring first two bytes before as it was LV and L is short length field. 
So, now you have changed it to compare whole array. that means it is not coming as LV now ? 
which PR has induced it ?

ajantha-bhat: This ByteArrayColumnWithRowId is common for other flows also. please check the impact. test cases may not fail as it is internal logic.

ajantha-bhat: As discussed, I think no dictionary string column comparison will go wrong after this change. no dictionary string to be alphabetically sorted we need to ignore L in LV. so please test and handle compare logic based on the flag.

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5246/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3499/

core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayColumnWithRowId.java
ShreelekhyaG: yes, this flow will be used for string types only. I have added a check for the dictionary and handled the compare logic.

core/src/main/java/org/apache/carbondata/core/datastore/columnar/ByteArrayColumnWithRowId.java
```diff
@@ -51,11 +59,12 @@ public boolean equals(Object obj) {
       return false;
     }
     ByteArrayColumnWithRowId o = (ByteArrayColumnWithRowId)obj;
-    return Arrays.equals(column, o.column) && rowId == o.rowId;
+    return Arrays.equals(column, o.column) && rowId == o.rowId
+        && isNoDictionary == o.isNoDictionary;
```
ajantha-bhat, line 63: Do we really need to consider `isNoDictionary` here ? when it will be not equal? I think as it is a comparison of data within a column, this condition might always be true. 

If not required, remove from equals and hashcode

ShreelekhyaG: Done. Removed.

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5251/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3504/

ShreelekhyaG:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5256/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3509/

ajantha-bhat:
LGTM

PR #4151, opened 2021-06-15
[CARBONDATA-4211] Fix - from xx Insert into select fails if an SQL statement contains multiple inserts

ShreelekhyaG:
### Why is this PR needed?
 When multiple inserts with single query is used, it fails from SparkPlan with: `java.lang.ClassCastException: GenericInternalRow cannot be cast to UnsafeRow`. 
For every successful insert/load we return Segment ID as a row. For multiple inserts also, we are returning a row containing Segment ID but while processing in spark `ClassCastException` is thrown.
 
When multiple insert query is given, it has `Union` node in the plan. Based on its presence, made changes to use flag `isMultipleInserts` to call class `UnionCommandExec ` and implemented custom `sideEffectResult` which converts `GenericInternalRow `to `UnsafeRow `and return.
    
 - No

 - Yes

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
@@ -22,15 +22,16 @@ import java.util.Locale
 import scala.collection.mutable
 
 import org.apache.log4j.Logger
+import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{CarbonCountStar, CarbonDatasourceHadoopRelation, CarbonToSparkAdapter, CountStarPlan, InsertIntoCarbonTable, SparkSession}
-import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
 import org.apache.spark.sql.catalyst.analysis.UnresolvedAlias
-import org.apache.spark.sql.catalyst.expressions.{Alias, Ascending, AttributeReference, Cast, Descending, Expression, IntegerLiteral, Literal, NamedExpression, ScalaUDF, SortOrder}
+import org.apache.spark.sql.catalyst.expressions.{Alias, Ascending, Attribute, AttributeReference, Cast, Descending, Expression, IntegerLiteral, Literal, NamedExpression, ScalaUDF, SortOrder, UnsafeProjection}
 import org.apache.spark.sql.catalyst.planning.{ExtractEquiJoinKeys, PhysicalOperation}
-import org.apache.spark.sql.catalyst.plans.{Inner, JoinType, LeftSemi}
+import org.apache.spark.sql.catalyst.plans.{Inner, LeftSemi}
 import org.apache.spark.sql.catalyst.plans.logical.{Filter, Join, Limit, LogicalPlan, Project, ReturnAnswer, Sort}
-import org.apache.spark.sql.execution.{CarbonTakeOrderedAndProjectExec, FilterExec, PlanLater, ProjectExec, SparkPlan, SparkStrategy}
-import org.apache.spark.sql.execution.command.{DataWritingCommandExec, ExecutedCommandExec, LoadDataCommand}
+import org.apache.spark.sql.execution.{CarbonTakeOrderedAndProjectExec, FilterExec, LeafExecNode, PlanLater, ProjectExec, SparkPlan, SparkStrategy}
+import org.apache.spark.sql.execution.command.{DataWritingCommandExec, ExecutedCommandExec, LoadDataCommand, RunnableCommand}
 import org.apache.spark.sql.execution.datasources.{InsertIntoHadoopFsRelationCommand, LogicalRelation}
 import org.apache.spark.sql.execution.joins.{BroadCastPolygonFilterPushJoin, BuildLeft, BuildRight}
 import org.apache.spark.sql.execution.strategy.CarbonPlanHelper.isCarbonTable
@@ -57,7 +58,16 @@ object DMLStrategy extends SparkStrategy {
       case loadData: LoadDataCommand if isCarbonTable(loadData.table) =>
         ExecutedCommandExec(DMLHelper.loadData(loadData)) :: Nil
       case insert: InsertIntoCarbonTable =>
-        ExecutedCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
+        if (insert.containsMultipleInserts) {
+          // Successful insert in carbon will return segment ID in a row.
+          // In-case of this specific multiple inserts scenario the Union node executes in the
+          // physical plan phase of the command, so the rows should be of unsafe row object.
+          // So we should override the sideEffectResult to prepare the content of command's
+          // corresponding rdd from physical plan of insert into command.
+          UnionCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
+        } else {
+          ExecutedCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
+        }
       case insert: InsertIntoHadoopFsRelationCommand
         if insert.catalogTable.isDefined && isCarbonTable(insert.catalogTable.get.identifier) =>
         DataWritingCommandExec(DMLHelper.insertInto(insert), planLater(insert.query)) :: Nil
@@ -362,6 +372,29 @@ object DMLStrategy extends SparkStrategy {
       }
       newPlan
     }
+  }
+}
+
+/**
+ * This class will be used when Union node is present in plan with multiple inserts.
+ * It is a physical operator that executes the run method of a RunnableCommand and
+ * saves the result to prevent multiple executions.
+ */
+case class UnionCommandExec(cmd: RunnableCommand) extends LeafExecNode {
+
+  protected[sql] lazy val sideEffectResult: Seq[InternalRow] = {
+    val converter = CatalystTypeConverters.createToCatalystConverter(schema)
+    val internalRow = cmd.run(sqlContext.sparkSession).map(converter(_).asInstanceOf[InternalRow])
+    val unsafeProjection = UnsafeProjection.create(output.map(_.dataType).toArray)
+    // To make GenericInternalRow to UnsafeRow
+    val row = unsafeProjection(internalRow.head)
+    Seq(row)
+  }
+
+  override def output: Seq[Attribute] = cmd.output
+
+  protected override def doExecute(): RDD[InternalRow] = {
+    sqlContext.sparkContext.parallelize(sideEffectResult, 1)
   }
 }
 
--- a/integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonAnalysisRules.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonAnalysisRules.scala
@@ -250,13 +250,15 @@ case class CarbonPreInsertionCasts(sparkSession: SparkSession) extends Rule[Logi
 
       case p@InsertIntoTable(relation: LogicalRelation, _, child, _, _)
         if relation.relation.isInstanceOf[CarbonDatasourceHadoopRelation] =>
-        castChildOutput(p, relation, child)
+        // when plan contains Union, it can have multiple insert statements as its children
+        castChildOutput(p, relation, child, plan.isInstanceOf[Union])
     }
   }
 
   def castChildOutput(p: InsertIntoTable,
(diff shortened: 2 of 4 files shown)


CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5546/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3803/

ShreelekhyaG:
retest this please

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5552/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3809/

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
```diff
@@ -294,6 +294,13 @@ case class CarbonInsertIntoCommand(databaseNameOp: Option[String],
         }
         throw ex
     }
+    if (internalOptions.contains("isMultipleInserts")) {
+      if (internalOptions.get("isMultipleInserts").get.toBoolean) {
+        // when multiple insert statements are given with single plan (Union),
+        // no need to return segment id as row.
+        return Seq.empty
```
akashrn5, line 301: here we are returning the segment ID from load command so that the application code integrated with carbondata can take decisions based on that. So if we don't return in this case, we will not give valid info to underlying application and may lead to wrong decision by applications if they were taking any. So we cant fix it like this

akashrn5:
Since our command classes extends spark command classes, `ExecutedCommandExec` will be the physical operator that executes the run method of a `RunnableCommand`. So we in this case we return row object, and this will be converted to `GenericInternalRow` in `sideEffectResult` of `ExecutedCommandExec`, but during Union operatio execution it expects to convert to Unsafe row and it fails with cast exception.

So as suggested by spark, we can override `sideEffectResult` and handle this case with the output of our command classes, like insert into command in this case, so that we can send proper results from physical node.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/management/CarbonInsertIntoCommand.scala
ShreelekhyaG: Made changes to return `UnsafeRow`. Created new case class `UnionCommandExec  `which overrides `LeafExecNode` and handles the conversion.

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5569/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3826/

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
```diff
@@ -394,6 +394,25 @@ class TestLoadDataGeneral extends QueryTest with BeforeAndAfterEach {
...
+        "cr_item_sk int)ROW FORMAT DELIMITED FIELDS TERMINATED BY '|' LINES TERMINATED BY '\\n'")
+    sql("insert into catalog_returns_5 values(1,2,3)")
+    sql("create table catalog_returns_6(cr_returned_time_sk int,cr_item_sk int) partitioned by" +
+        " (cr_returned_date_sk int) stored as carbondata")
+    sql(
+      "from catalog_returns_5 insert overwrite table catalog_returns_6 partition " +
+      "(cr_returned_date_sk) select cr_returned_time_sk, cr_item_sk, cr_returned_date_sk where " +
+      "cr_returned_date_sk is not null distribute by cr_returned_date_sk insert overwrite table " +
+      "catalog_returns_6 partition (cr_returned_date_sk) select cr_returned_time_sk, cr_item_sk, " +
+      "cr_returned_date_sk where cr_returned_date_sk is null distribute by cr_returned_date_sk")
```
akashrn5, line 410: here please collect the segments IDs returned from the insert command and validate them also.

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
```diff
@@ -394,6 +394,25 @@ class TestLoadDataGeneral extends QueryTest with BeforeAndAfterEach {
...
+    checkAnswer(sql("select *from catalog_returns_6"), Seq(Row(2, 3, 1)))
```
akashrn5, line 411:
```suggestion
    checkAnswer(sql("select *cfrom catalog_returns_6"), Seq(Row(2, 3, 1)))
```

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
```diff
@@ -365,3 +374,33 @@ object DMLStrategy extends SparkStrategy {
...
+
+  override def executeCollect(): Array[InternalRow] = sideEffectResult.toArray
+
+  override def executeToIterator: Iterator[InternalRow] = sideEffectResult.toIterator
+
+  override def executeTake(limit: Int): Array[InternalRow] = sideEffectResult.take(limit).toArray
+
+  protected override def doExecute(): RDD[InternalRow] = {
+    sqlContext.sparkContext.parallelize(sideEffectResult, 1)
+  }
```
akashrn5, line 404: here no need to override all, just can override abstract one like` doExecute()` and `output`, others we are not using so no need to override since no specific logic

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
```diff
@@ -57,7 +59,14 @@ object DMLStrategy extends SparkStrategy {
       case loadData: LoadDataCommand if isCarbonTable(loadData.table) =>
         ExecutedCommandExec(DMLHelper.loadData(loadData)) :: Nil
       case insert: InsertIntoCarbonTable =>
-        ExecutedCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
+        if (insert.isMultipleInserts) {
+          // when multiple insert statements are given with single plan (Union),
+          // Spark expects row to be of UnsafeRow. Here use UnionCommandExec to
+          // implement custom sideEffectResult and return row as UnsafeRow
+          UnionCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
```
akashrn5, line 66: here the comment gives little different meaning, here you can say about how insert command in carbon returns the rows containing corresponding segmentIDs and incase of this specific multiple scenario the Union node executes in the physical plan phase of the command, so the rows should be of unsafe row object. So we should override the `sideEffectResult` to prepare the content of command's corresponding rdd from physical plan of insert into command. You can clean it and update the command something like this

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
```diff
@@ -365,3 +374,33 @@ object DMLStrategy extends SparkStrategy {
   }
 }
 
+case class UnionCommandExec(cmd: RunnableCommand) extends LeafExecNode {
```
akashrn5, line 377: give a class level comment

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
ShreelekhyaG: Done

integration/spark/src/test/scala/org/apache/carbondata/integration/spark/testsuite/dataload/TestLoadDataGeneral.scala
ShreelekhyaG: Done

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
ShreelekhyaG: ok, removed

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
ShreelekhyaG: Done

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
ShreelekhyaG: Done

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
```diff
@@ -57,7 +58,16 @@ object DMLStrategy extends SparkStrategy {
       case loadData: LoadDataCommand if isCarbonTable(loadData.table) =>
         ExecutedCommandExec(DMLHelper.loadData(loadData)) :: Nil
       case insert: InsertIntoCarbonTable =>
-        ExecutedCommandExec(CarbonPlanHelper.insertInto(insert)) :: Nil
+        if (insert.isMultipleInserts) {
+          // Successful insert in carbon will return segment ID as row.
```
akashrn5, line 62:
```suggestion
          // Successful insert in carbon will return segment ID in a row.
```

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonAnalysisRules.scala
```diff
@@ -250,13 +251,19 @@ case class CarbonPreInsertionCasts(sparkSession: SparkSession) extends Rule[Logi
 
       case p@InsertIntoTable(relation: LogicalRelation, _, child, _, _)
         if relation.relation.isInstanceOf[CarbonDatasourceHadoopRelation] =>
-        castChildOutput(p, relation, child)
+        var isMultipleInserts = false
```
akashrn5, line 254:
```suggestion
        var containsMultipleInserts = false
```
Please in all places

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3831/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5574/

integration/spark/src/main/scala/org/apache/spark/sql/execution/strategy/DMLStrategy.scala
ShreelekhyaG: done

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonAnalysisRules.scala
ShreelekhyaG: done

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3832/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5575/

ShreelekhyaG:
retest this please

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/3838/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5581/

akashrn5:
LGTM

PR #4204, opened 2021-08-16
[CARBONDATA-4270]Delete segment expect remain_number

W1thOut:
### Why is this PR needed?
 In some scenarios, need to delete old segments in batches and keep only latest few segments.
 
Add a DML.
    
 - No

 - Yes


CarbonDataQA2:

PR #4209, opened 2021-08-23
[CARBONDATA-4278] Avoid refetching all indexes to get segment properties

maheshrajus:
### Why is this PR needed?
When block index[BlockIndex] is available then no need to prepare indexes[List[BlockIndex] from available segments and partition locations which might cause delay in query performance.

Call directly get segment properties if block index[BlockIndex] available.
```
      if (segmentIndices.get(0) instanceof BlockIndex) {
        segmentProperties =
            segmentPropertiesFetcher.getSegmentPropertiesFromIndex(segmentIndices.get(0));
      } else {
        segmentProperties =
            segmentPropertiesFetcher.getSegmentProperties(segment, partitionLocations);
      }
```
getSegmentPropertiesFromIndex  is calling directly block index segment properties.
    
 - No

 - No. Already index related test cases are present which can cover the added code.

### DIFF
--- a/core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
+++ b/core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
@@ -48,6 +48,7 @@ import org.apache.carbondata.core.indexstore.BlockletDetailsFetcher;
 import org.apache.carbondata.core.indexstore.ExtendedBlocklet;
 import org.apache.carbondata.core.indexstore.PartitionSpec;
 import org.apache.carbondata.core.indexstore.SegmentPropertiesFetcher;
+import org.apache.carbondata.core.indexstore.blockletindex.BlockIndex;
 import org.apache.carbondata.core.metadata.AbsoluteTableIdentifier;
 import org.apache.carbondata.core.metadata.schema.table.CarbonTable;
 import org.apache.carbondata.core.metadata.schema.table.IndexSchema;
@@ -207,15 +208,22 @@ public final class TableIndex extends OperationEventListener {
       Map<Segment, List<Index>> indexes) throws IOException {
     Set<String> missingSISegments = filter.getMissingSISegments();
     for (Segment segment : segments) {
+      List<Index> segmentIndices = indexes.get(segment);
       if (segment == null ||
-          indexes.get(segment) == null || indexes.get(segment).isEmpty()) {
+          segmentIndices == null || segmentIndices.isEmpty()) {
         continue;
       }
       boolean isExternalOrMissingSISegment = segment.isExternalSegment() ||
           (missingSISegments != null && missingSISegments.contains(segment.getSegmentNo()));
       List<Blocklet> pruneBlocklets = new ArrayList<>();
-      SegmentProperties segmentProperties =
-          segmentPropertiesFetcher.getSegmentProperties(segment, partitionLocations);
+      SegmentProperties segmentProperties;
+      if (segmentIndices.get(0) instanceof BlockIndex) {
+        segmentProperties =
+            segmentPropertiesFetcher.getSegmentPropertiesFromIndex(segmentIndices.get(0));
+      } else {
+        segmentProperties =
+            segmentPropertiesFetcher.getSegmentProperties(segment, partitionLocations);
+      }
       if (filter.isResolvedOnSegment(segmentProperties)) {
         FilterExecutor filterExecutor;
         if (!isExternalOrMissingSISegment) {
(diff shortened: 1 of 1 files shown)


CarbonDataQA2:
Build Failed  with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/280/

CarbonDataQA2:
Build Failed  with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/281/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5879/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4136/

core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
```diff
@@ -207,15 +208,22 @@ public CarbonTable getTable() {
...
+          indexes.get(segment) == null || segmentIndices.isEmpty()) {
         continue;
       }
       boolean isExternalOrMissingSISegment = segment.isExternalSegment() ||
           (missingSISegments != null && missingSISegments.contains(segment.getSegmentNo()));
       List<Blocklet> pruneBlocklets = new ArrayList<>();
-      SegmentProperties segmentProperties =
-          segmentPropertiesFetcher.getSegmentProperties(segment, partitionLocations);
+      SegmentProperties segmentProperties;
+      if (segmentIndices.get(0) instanceof BlockIndex) {
```
ajantha-bhat, line 220: Need to loop here and check if any of the segmentIndices is BlockIndex instead of always checking the first index?

ajantha-bhat: Also CC: @Indhumathi27 , you can also review once. 
May be we can remove   `segmentPropertiesFetcher.getSegmentProperties(segment, partitionLocations); `
and always call `segmentPropertiesFetcher.getSegmentPropertiesFromIndex` as we have indexes already at this place.

maheshrajus: @ajantha-bhat  @Indhumathi27 1) we are getting segment properties from one blockindex which is same for all. 
2) In this flow we may receive SI and bloom indexes also which will prepare by SI , bloom index factory [List(SecondaryIndex) or List(BloomIndex)]. In this case we should use "segmentPropertiesFetcher.getSegmentProperties" as they are not instance of BlockIndex. inside getSegmentProperties method we are getting and preparing list of BlockIndex and getting segment propteries from block index.

I will add scenario description in code and push.

core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
```diff
@@ -207,15 +208,22 @@ public CarbonTable getTable() {
       Map<Segment, List<Index>> indexes) throws IOException {
     Set<String> missingSISegments = filter.getMissingSISegments();
     for (Segment segment : segments) {
+      List<Index> segmentIndices = indexes.get(segment);
       if (segment == null ||
-          indexes.get(segment) == null || indexes.get(segment).isEmpty()) {
+          indexes.get(segment) == null || segmentIndices.isEmpty()) {
```
ajantha-bhat, line 213: In line 213, `indexes.get(segment)` is called again. Avoid it by using the existing result from `segmentIndices`.

core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
ajantha-bhat: I mean line 211, when `indexes.get(segment)` is done, all indexes will be present in this list. Like blockindex, SI, bloom, lucence. 
so, we just have to find blockindex (which is always be present even for SI table) and get the the segment properties ? 
@Indhumathi27 : Is my understanding correct or we need both cases here ?

maheshrajus: i checked in SI flow, it has only SI index info and does not have blockindex info. We can add separate checks

core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
maheshrajus: done

core/src/main/java/org/apache/carbondata/core/index/TableIndex.java
Indhumathi27: When cgIndexExprWrapper.prune is called (for SI/Bloom), in that case, it will have either BloomcoarseGainIndex or SIindex. So, we have to keep both cases

ajantha-bhat:
LGTM

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5907/

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/309/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4163/

PR #4211, opened 2021-08-27
[CARBONDATA-4273] Fix Cannot create external table with partitions

Indhumathi27:
### Why is this PR needed?
 Create partition table with location fails with unsupported message.
 
This scenario works in cluster mode. This check can be moved in local mode also and partition table can be created with table with location
    
 - No

 - Yes

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/parser/CarbonSparkSqlParserUtil.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/parser/CarbonSparkSqlParserUtil.scala
@@ -215,10 +215,6 @@ object CarbonSparkSqlParserUtil {
       table.identifier.table.toLowerCase()
     )
     val tableInfo = if (isExternal) {
-      if (partitionColumnNames.nonEmpty) {
-        throw new MalformedCarbonCommandException(
-          "Partition is not supported for external table")
-      }
       // read table info from schema file in the provided table path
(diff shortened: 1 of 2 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/parser/CarbonSparkSqlParserUtil.scala
```diff
@@ -215,10 +215,6 @@ object CarbonSparkSqlParserUtil {
       table.identifier.table.toLowerCase()
     )
     val tableInfo = if (isExternal) {
-      if (partitionColumnNames.nonEmpty) {
```
Lior-AI, line 218: Why can we remove it?
do we support this now?

Indhumathi27: This scenario, already works fine in cluster. I have tested in Spark-Hadoop cluster. Local mode only uses this flow. So, better to keep same for all environment

akashrn5:
@Indhumathi27 can you please add the PR description?

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/307/

akashrn5:
LGTM

PR #4219, opened 2021-09-17
[CARBONDATA-4288][CARBONDATA-4289] Fix various issues with Index Server caching mechanism.

vikramahuja1001:
### Why is this PR needed?
 There are 2 issues in the Index Server flow:
1. In case when there is a main table with a SI table with prepriming disabled and index serve enabled, new load to main table and SI table put the cache for the main table in the index server. Cache is also getting again when a select query is fired. This issue happens because during load to SI table, getSplits is called on the main table segment which is in Insert In Progress state. Index server considers this segment as a legacy segment because it's index size = 0 and does not put it's entry in the tableToExecutor mapping. In the getsplits method isRefreshneeded is false the first time getSplits is called. During the select query, in getSplits method isRefreshNeeded is true and the previous loaded entry is removed from the driver but since there is no entry for that table in tableToExecutor mapping, the previous cache value becomes dead cache and always stays in the index server. The newly loaded cache is loaded to a new executor and 2 copies of cache for the same segment is being mantained.
2. Concurrent select queries to the index server shows wrong cache values in the Index server.
 
The following changes are proposed to the index server code:
1. Removing cache object from the index server in case the segment is INSERT IN PROGRESS and in the case of legacy segment adding the value in tabeToExecutor mappping so that the cache is also removed from the executor side.
2. Concurrent queries were able adding duplicate cache values to other executors. Changed logic of assign executors method so that concurrent queries are not able to add cache for same segment in other executors
    
 - No

 - Yes. Test case added to check assignExecutors method using multiple threads.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
@@ -108,6 +108,11 @@ object DistributedRDDUtils {
           val wrapper: IndexInputSplit = legacySegment
             .asInstanceOf[IndexInputSplitWrapper].getDistributable
           val executor = validExecutorIds(index % validExecutorIds.length)
+          // Below code is used to support concurrent queries on same segment going to
+          // the same executor for caching. Putting a new HashMap in tableToExecutorMapping
+          // and whichever query is able to put entry in existingSegmentMapping decides the
+          // executor which will be used later on. For one thread oldMapping will be null and
+          // for the second it will be the executor returned which will be used.
           tableToExecutorMapping.putIfAbsent(tableUniqueName,
             new ConcurrentHashMap[String, String]())
           val existingSegmentMapping = tableToExecutorMapping.get(tableUniqueName)
@@ -180,7 +185,8 @@ object DistributedRDDUtils {
   def isSegmentInProgress(request: IndexInputFormat, segment: String): Boolean = {
     request.getReadCommittedScope.getSegmentList.find(_.getLoadName
       .equalsIgnoreCase(segment)) match {
-      case Some(value) => value.getSegmentStatus.equals(SegmentStatus.INSERT_IN_PROGRESS)
+      case Some(value) => value.getSegmentStatus.equals(SegmentStatus.INSERT_IN_PROGRESS) || value
+        .getSegmentStatus.equals(SegmentStatus.INSERT_OVERWRITE_IN_PROGRESS)
       case None => false
     }
   }
--- a/integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
+++ b/integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
@@ -357,13 +357,11 @@ object DistributedRDDUtils {
     val existingExecutorMapping = executorToCacheSizeMapping.get(host)
     if (existingExecutorMapping != null) {
       val existingSize = existingExecutorMapping.get(executor)
+      var totalSize = segment.getIndexSize
       if (existingSize != null) {
-        existingExecutorMapping.put(executor, existingSize + segment.getIndexSize
-          .toInt)
-      } else {
-        existingExecutorMapping.put(executor, segment.getIndexSize
-          .toInt)
+        totalSize += existingSize
       }
+      existingExecutorMapping.put(executor, totalSize.toInt)
     } else {
       val newExecutorMapping = new ConcurrentHashMap[String, Long]()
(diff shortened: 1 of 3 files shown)


CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4190/

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/337/

integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
```diff
@@ -164,6 +177,14 @@ object DistributedRDDUtils {
     }
   }
 
+  def isSegmentInProgress(request: IndexInputFormat, segment: String): Boolean = {
+    request.getReadCommittedScope.getSegmentList.find(_.getLoadName
+      .equalsIgnoreCase(segment)) match {
+      case Some(value) => value.getSegmentStatus.equals(SegmentStatus.INSERT_IN_PROGRESS)
```
Indhumathi27, line 183: may be need to handle INSERT_OVERWRITE_IN_PROGRESS segment also here

integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
```diff
@@ -108,7 +108,20 @@ object DistributedRDDUtils {
           val wrapper: IndexInputSplit = legacySegment
             .asInstanceOf[IndexInputSplitWrapper].getDistributable
           val executor = validExecutorIds(index % validExecutorIds.length)
-          wrapper.setLocations(Array("executor_" + executor))
+          tableToExecutorMapping.putIfAbsent(tableUniqueName,
```
akashrn5, line 111: please add a detailed comment explaining this logic, so that next time if any issues, developer will be aware of the old issue fix

vikramahuja1001: done

integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
vikramahuja1001: done

integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
```diff
@@ -313,31 +340,34 @@ object DistributedRDDUtils {
...
-        tableToExecutorMapping.putIfAbsent(tableUniqueName, newSegmentMapping)
+    }
+  }
+
+  private def updateCacheSize(host: String, executor: String, segment: Segment) = {
+    val existingExecutorMapping = executorToCacheSizeMapping.get(host)
+    if (existingExecutorMapping != null) {
+      val existingSize = existingExecutorMapping.get(executor)
+      if (existingSize != null) {
+        existingExecutorMapping.put(executor, existingSize + segment.getIndexSize
```
Indhumathi27, line 361: can get size based on existing size and can move  common code "existingExecutorMapping.put(executor, size)" down

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5939/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4193/

integration/spark/src/main/scala/org/apache/carbondata/indexserver/DistributedRDDUtils.scala
vikramahuja1001: done

Indhumathi27:
LGTM

akashrn5:
LGTM

kunal642:
LGTM

CarbonDataQA2:
Build Failed  with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/341/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/5940/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4194/

PR #4234, opened 2021-10-12
[CARBONDATA-4304] Initialize CarbonEnv will try to create folder on local filesystem if storelocation is set without scheme

jack86596:
### Why is this PR needed?
If storelocation doesn't have scheme, file system is treated as local by default, not the file system defined by fs.defaultFS.
 
1. call fs.makeQualified and get the final storepath with correct scheme.
2. use storepath with scheme to create the folder, keep storepath without scheme in CarbonProperties, because storepath in CarbonProperties will be used to compare with spark.sql.warehouse.dir which doesn't have scheme also.
    
 - No

 - Yes


CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/6051/

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/441/

kunal642:
@jack86596 does this problem exists only with "carbon.storelocation" property?

PR #4237, opened 2021-11-18
line 117

WANGSHIHUAI:
#117, The longitude is six decimal places and the dimension is five digits. Why is it the same length after conversion?

 ### Why is this PR needed?
 here make me confused.
 
update the demo, that make people easily undestand.
    
 - No

 - No


CarbonDataQA2:

chenliang613:
LTGM+1

PR #4242, opened 2021-12-08
[CARBONDATA-4318]Improve load overwrite performance for partition tables

akashrn5:
### Why is this PR needed?
 With the increase in the number of overwrite loads for the partition table, the time takes for each load keeps on increasing over time. This is because,

1. whenever a load overwrite for partition table is fired, it basically means that we need to overwrite or drop the partitions if anything overlaps with current partitions getting loaded. Since carbondata stores the partition information in the segments file, to identify and drop partitions, it's reading all the previous segment files to identify and drop the overwriting partitions, which leads to a decrease in performance.
2. After partition load is completed, a `cleanSegments` method is called which again reads segment file and table status file to identify MArked for delete segments to clean. But Since the force clean is false and timeout also is more than a day by default, it's not necessary to call this method. Clean files should handle this part.

 
1. we already have the information about current partitions, so with that first identify if there are any partitions to overwrite, if yes then only we read segment files to call `dropParitition`, else we don't read the segment files unnecessarily. It also contains other refactoring to avoid reading table status file also.
2. no need to call clean segments after every load. Clean files will take care to delete the expired ones.
    
 - No

 - No

### DIFF
--- a/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonOutputCommitter.java
+++ b/hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonOutputCommitter.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.carbondata.common.logging.LogServiceFactory;
 import org.apache.carbondata.core.constants.CarbonCommonConstants;
@@ -38,7 +39,6 @@ import org.apache.carbondata.core.metadata.schema.table.CarbonTable;
 import org.apache.carbondata.core.mutate.CarbonUpdateUtil;
 import org.apache.carbondata.core.statusmanager.LoadMetadataDetails;
 import org.apache.carbondata.core.statusmanager.SegmentStatus;
-import org.apache.carbondata.core.statusmanager.SegmentStatusManager;
 import org.apache.carbondata.core.util.CarbonProperties;
 import org.apache.carbondata.core.util.CarbonSessionInfo;
 import org.apache.carbondata.core.util.ObjectSerializationUtil;
@@ -119,9 +119,12 @@ public class CarbonOutputCommitter extends FileOutputCommitter {
 
     boolean overwriteSet = CarbonTableOutputFormat.isOverwriteSet(context.getConfiguration());
     CarbonLoadModel loadModel = CarbonTableOutputFormat.getLoadModel(context.getConfiguration());
+    List<PartitionSpec> currentPartitionsOfTable = (List<PartitionSpec>) ObjectSerializationUtil
+        .convertStringToObject(context.getConfiguration().get("carbon.currentpartition"));
     if (loadModel.getCarbonDataLoadSchema().getCarbonTable().isHivePartitionTable()) {
       try {
-        commitJobForPartition(context, overwriteSet, loadModel, partitionPath);
+        commitJobForPartition(context, overwriteSet, loadModel, partitionPath,
+            currentPartitionsOfTable);
       } catch (Exception e) {
         CarbonLoaderUtil.updateTableStatusForFailure(loadModel);
         LOGGER.error("commit job failed", e);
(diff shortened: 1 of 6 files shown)


CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4410/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/6153/

CarbonDataQA2:
Build Failed  with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/544/

CarbonDataQA2:
Build Failed  with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/6154/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4411/

CarbonDataQA2:
Build Failed  with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/545/

CarbonDataQA2:
Build Failed  with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4412/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/6155/

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/546/

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonOutputCommitter.java
```diff
@@ -316,31 +326,38 @@ private void commitJobForPartition(JobContext context, boolean overwriteSet,
...
+      List<String> overlappingPartitions = currentPartitionsOfTable.stream()
+          .map(partitionSpec -> partitionSpec.getLocation().toString())
+          .filter(partitionList::contains).collect(Collectors.toList());
+      if (!overlappingPartitions.isEmpty()) {
+        List<LoadMetadataDetails> validLoadMetadataDetails =
+            loadModel.getLoadMetadataDetails().stream().filter(
+                loadMetadataDetail -> !loadMetadataDetail.getLoadName()
+                    .equalsIgnoreCase(newMetaEntry.getLoadName())).collect(Collectors.toList());
+        String uniqueId = String.valueOf(System.currentTimeMillis());
+        List<String> toBeUpdatedSegments = new ArrayList<>();
```
vikramahuja1001, line 343: maybe for this list we can give the max length as number of segments while instantiating.

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonOutputCommitter.java
```diff
@@ -316,31 +326,38 @@ private void commitJobForPartition(JobContext context, boolean overwriteSet,
...
+        List<String> toBeDeletedSegments = new ArrayList<>();
```
vikramahuja1001, line 344: same as above

hadoop/src/main/java/org/apache/carbondata/hadoop/api/CarbonOutputCommitter.java
akashrn5: anyways, this was old code just added in a condition, but yeah, initialized with an initial capacity of `validLoadMetadataDetails`

CarbonDataQA2:
Build Success with Spark 3.1, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_3.1/582/

CarbonDataQA2:
Build Success with Spark 2.3.4, Please check CI http://121.244.95.60:12602/job/ApacheCarbonPRBuilder2.3/6191/

CarbonDataQA2:
Build Success with Spark 2.4.5, Please check CI http://121.244.95.60:12602/job/ApacheCarbon_PR_Builder_2.4.5/4447/

kunal642:
LGTM

PR #4330, opened 2023-08-19
[ISSUE-4329] optimize some code smells in presto module

xubo245:
static method shouldn't be called by object, it should be call by class

 ### Why is this PR needed?
 There are some code smells in presto module
For example
1.Didn't add @OverRide for override method
2.magic number
3.variable Should Use LowerCamelCase
 
    optimize some code smells in presto module
 - No

 - No


chenliang613:
LGTM

PR #4337, opened 2023-11-06
Fix TestInsertAndOtherCommandConcurrent testcasee issue and enble testcase

QiangCai:
### Why is this PR needed?
1. TestInsertAndOtherCommandConcurrent  testcase not working
2. After testing, there are many garbage files under project folder
 
 1. Fix TestInsertAndOtherCommandConcurrent testcasee issue
 3. Using a proper target path in testcase to avoid generating new files during test
 4. Enable spark-2.4 by default
    
 - No 

 - Yes, enable 4 testcases

### DIFF
--- a/examples/spark/pom.xml
+++ b/examples/spark/pom.xml
@@ -200,15 +200,15 @@
     </profile>
     <profile>
       <id>spark-2.3</id>
-      <activation>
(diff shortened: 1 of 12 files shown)


chenliang613:
looks good to me.

PR #4348, opened 2024-03-15
Use https link instead of http

git-hulk:
### Why is this PR needed?
It'd be better to use HTTPS for the website link instead of HTTP.  
 
Change HTTP to HTTPS in pom.xml and README

    
 - No

 - No


chenliang613:
looks good to me

PR #4356, opened 2024-06-02
[CARBONDATA-4350] Update document for thrift version

jackylk:
### Why is this PR needed?
 The build document should be updated to reflect the thrift version update
 
/build/README is updated
    
 - No

 - No


chenliang613:
LGTM

PR #4367, opened 2025-06-30
[HOTFIX] Add length validation before calling Snappy uncompress

jackylk:
### Why is this PR needed?
 
There is a security issue reported by the Apache Security Team. Calling Snappy.uncompress without length validation is subject to attack.
 
Add length validation before calling Snappy, modified all places in carbondata-core SnappyCompressor

    
 - No

 - No


chenliang613:
LGTM

PR #4388, opened 2026-03-17
[MINOR] Update links for Java and Thrift in README

pjfanning:
### Why is this PR needed?
 
We should avoid non TLS addresses for downloads. There might be other pages in this or other CarbonData repos but I ran into this one.
 
 - No
 
 - No


chenliang613:
+1， thanks for contribution.

